PR #10, opened 2026-03-27
feat(scorecard): strengthen AgentNative dimension with quality checks

tmchow:
## Summary

The AgentNative scorecard dimension (0-10) previously only checked for the *presence* of agent flags and exit codes. This PR redistributes the scoring to also validate the *quality* of agent-readiness features the templates already generate, catching CLIs that have the right flags but broken examples or unbounded defaults.

## What changed

The five core flag checks (+1 each for `--json`, `--select`, `--dry-run`, `--stdin`, `--yes`) stay as-is. The remaining 5 points were redistributed from the old non-interactive, exit-code, and stdin-excellence checks into three new quality checks:

| Check | Points | What it validates |
|-------|--------|-------------------|
| Help example validity | +2 | Samples up to 10 leaf command files, awards +1 if >=80% have non-empty `Example:` fields, +1 if >=80% of examples reference only flags that exist as declarations in the file or root.go |
| Error message actionability | +1 | Awards point if `helpers.go` contains >=3 distinct actionable suggestion patterns (`try`, `use`, `run`, `check`, `see`, `ensure`, `verify`) |
| Bounded output defaults | +1 | Awards point if any CLI file contains a default limit mechanism — scans all `internal/cli/*.go` files (not just root/helpers) to catch command-level limits in search, analytics, etc. Requires non-zero defaults so `"limit", 0` (unlimited) doesn't false-positive |
| Exit codes (reduced) | +1 | Reduced from +2 to +1 since `ErrorHandling` already scores exit codes separately |

Key design decisions:
- **Leaf-command filtering**: Example scoring filters to files containing `RunE`/`Run:`, excluding parent command files (from `command_parent.go.tmpl`) that only define `Use`/`Short` and would otherwise drag the 80% threshold below passing
- **Broad bounded-output scan**: Reads all `.go` files in `internal/cli/` including infra files (`search.go`, `analytics.go`) that `sampleCommandFiles` normally excludes, since those are exactly the files that define their own `--limit` flags
- **Non-zero default enforcement**: The `--limit` fallback regex requires `[1-9][0-9]*` after `"limit",` so `export.go.tmpl`'s `"limit", 0, "... unlimited"` pattern correctly scores as unbounded

## Test plan

Added `scorecard_agent_native_test.go` with 30 tests covering each check in isolation and combined:
- `TestScoreAgentNative`: zero/flags-only/max integration tests
- `TestExtractExampleField`: backtick, double-quote, missing, empty
- `TestAgentNativeHelpExampleValidity`: valid flags, invalid flags, empty examples, root.go flag inheritance, parent-command exclusion
- `TestAgentNativeErrorActionability`: 3+ distinct, fewer than 3, deduplication
- `TestAgentNativeBoundedOutput`: helpers, client, root, command-level, infra files, zero-default rejection, no bounds
- `TestAgentNativeExitCodeReduced`: 3+ codes, fewer than 3

---

[Compound Engineering v2.54.1](https://github.com/EveryInc/compound-engineering-plugin)
🤖 Generated with Claude Opus 4.6 (1M context, extended thinking) via [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/pipeline/scorecard.go
+++ b/internal/pipeline/scorecard.go
@@ -373,6 +373,7 @@ func scoreDoctor(dir string) int {
 func scoreAgentNative(dir string) int {
 	rootContent := readFileContent(filepath.Join(dir, "internal", "cli", "root.go"))
 	helpersContent := readFileContent(filepath.Join(dir, "internal", "cli", "helpers.go"))
+	clientContent := readFileContent(filepath.Join(dir, "internal", "client", "client.go"))
 	score := 0
 	// Presence: core agent flags (1pt each, max 5)
 	if strings.Contains(rootContent, `"json"`) {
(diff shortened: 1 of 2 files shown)


mvanhorn:
I like the theory, but at least have you and your agent use run data

PR #29, opened 2026-03-28
fix(scorecard): handle unscored auth semantics

tmchow:
## Summary

Fixes the scorecard's spec-derived auth semantics so missing or non-applicable auth evidence no longer quietly penalizes generated CLIs, while still treating real auth requirements as scoreable when the OpenAPI spec actually requires them.

## What Changed

- Fail fast in `RunScorecard` when a provided `--spec` cannot be read or parsed instead of renormalizing around a broken semantic-validation input.
- Track unscored spec-derived dimensions explicitly and render them as `N/A` in human output while omitting them from the Tier 2 denominator and gap report.
- Preserve backward compatibility for JSON consumers by keeping numeric score fields and using `unscored_dimensions` as the machine-readable signal for `N/A`.
- Rework auth scoring to follow effective OpenAPI security semantics:
  - unused declared schemes stay unscored
  - OAuth2 and OpenID Connect remain scoreable requirements
  - OR alternatives choose the best valid option
  - anonymous alternatives and operation-level overrides can make auth unscored
- Remove the legacy helper wrappers that still collapsed bad spec input to `0`, leaving a single score-evaluation path.
- Update the scorecard skill guidance and the scorecard architecture learning to document the `N/A` / omitted-from-denominator rule.

## Why This Approach

The original midpoint fallback (`5/10`) looked neutral but still lowered the final score because it stayed inside the denominator. The final implementation separates three cases that had been conflated:

- invalid evidence source: error
- missing or non-applicable evidence: `N/A` / unscored
- real auth requirement with missing implementation: scored low

That keeps the score honest without inflating APIs that genuinely require auth.

## Verification

- `go test ./...`
- Added Tier 2 regressions for invalid specs, unused schemes, OAuth scoring, anonymous alternatives, alternative auth schemes, and root-security overrides

---

[Compound Engineering v2.56.1](https://github.com/EveryInc/compound-engineering-plugin)
🤖 Generated with GPT-5 (unknown context, standard reasoning) via [Codex](https://openai.com/codex/)

### DIFF
--- a/docs/solutions/best-practices/steinberger-scorecard-scoring-architecture-2026-03-27.md
+++ b/docs/solutions/best-practices/steinberger-scorecard-scoring-architecture-2026-03-27.md
@@ -1,6 +1,7 @@
 ---
 title: "Steinberger scorecard scoring system: architecture, conventions, and modification rules"
 date: 2026-03-27
+last_updated: 2026-03-28
 category: best-practices
 module: internal/pipeline
 problem_type: best_practice
@@ -21,6 +22,7 @@ tags:
   - tier-weighting
   - best-practice
   - calibration
+  - unscored-dimensions
 ---
 
 # Steinberger scorecard: architecture, conventions, and modification rules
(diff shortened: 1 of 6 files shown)


mvanhorn:
Nice work on this. The three-state split (error / N/A / scored) is the right call - the old midpoint fallback was quietly dragging scores down for APIs that just don't need auth. Auth semantic handling looks correct across the OpenAPI edge cases too. Solid test coverage.

Issue #275, opened 2026-04-25
Catalog smoke-test sweep: 3 generator bugs + 1 parser bug + 4 stale spec_urls


pejmanjohn:
## TL;DR

A smoke-test sweep across the 19 catalog entries (regenerate → `go build` → `go vet` → `--help`/`version`/`doctor`) surfaced **5 distinct findings**, validated against repro and de-duplicated against false positives.

| ID | Severity | Class | Affected | Validated |
|----|----------|-------|----------|-----------|
| F-1 | HIGH | generator (CLI flag) | every CLI generated | ✅ clean repro |
| F-2 | HIGH | generator (template) | `github`, `twilio` (broken builds) | ✅ confirmed |
| F-3 | HIGH | generator (template) | `github` (broken build) | ✅ confirmed |
| F-4 | MEDIUM | parser | `digitalocean` (can't generate) | ✅ confirmed |
| F-5 | LOW | catalog data | `asana`, `sendgrid`, `square`, `postman-explore` | ✅ confirmed (HTTP 404) |

Discovered while bringing [printing-press-library#109](https://github.com/mvanhorn/printing-press-library/pull/109) (meta-ads) up to date with current main. Findings are independent of meta-ads and apply to the catalog at large. All findings have **clean isolated reproducers** below — they were validated with at least one re-run after the smoke pass to rule out transient or environmental causes.

This issue is intentionally consolidated: every finding has a self-contained reproducer, a probable root cause with template/file references, and a suggested fix direction. Split into separate issues at your discretion if that's easier to track.

---

## Methodology (so this is rerunnable)

Built `printing-press` from `main` HEAD (commit `ab7254b`). For each `catalog/*.yaml`:

1. Download `spec_url` to `/tmp/smoke-specs/<name>.spec` (max 60s).
2. Run `printing-press generate --spec <spec> --output /tmp/smoke-<name>` and record exit code + stderr.
3. If generation succeeded, locate the actual output dir (see F-1) and run:
   - `go build ./...`
   - `go vet ./...`
   - `<cli> --help`
   - `<cli> version`
   - `<cli> doctor`
4. Capture all logs to `/tmp/smoke-results/<name>.<step>.log` and a CSV summary.

Result tally:

```
PASS (8):           discord, launchdarkly, petstore, pipedrive, plaid, stripe, stytch + 1
PASS w/ rename (4): front, hubspot, sentry, telegram (see F-1)
[...]

PR #282, opened 2026-04-25
fix(catalog): update stale spec URLs + add download content-validity check

hnshah:
## What

Fixes two stale catalog spec URLs and adds a content-validity guard that catches broken downloads early.

### URL fixes
- **asana**: `Asana/openapi` uses `master` branch, not `main` (HTTP 404 on the old URL)
- **square**: repo renamed from `square-openapi` to `connect-api-specification`, file moved from `openapi.json` to `api.json`

### Content-validity check (F-5c from #275)

When `fetchOrCacheSpec` downloads a response that is:
- Less than 256 bytes AND
- Starts with `<` (HTML error page) or matches `^\d{3}:\s` (status-text like `404: Not Found`)

...it fails with a clear error instead of forwarding the garbage to the parser. Example:

```
Error: spec_url https://... returned a small response that does not look like an OpenAPI spec (14 bytes): "404: Not Found"
```

Previously the parser would emit confusing errors like `validation: name is required` on the error page body.

### What's still broken

These catalog entries still have stale/broken spec_urls and need maintainer decisions:
- **sendgrid**: consolidated spec removed upstream; split into ~30 `tsg_*.json` files that can't be used as a single spec
- **postman-explore, producthunt**: spec_url points to the private `cli-printing-press` repo; end users without auth can't fetch them
- **digitalocean**: spec parses but `Tag.description` is an object, not a string (F-4 from #275, separate parser fix needed)

## Testing

```
go test ./internal/cli/ -run TestFetchOrCacheSpec -v
--- PASS: TestFetchOrCacheSpec_ContentValidityCheck (0.00s)
```

Full `./internal/cli/` test suite passes.

Refs #275

### DIFF
--- a/internal/cli/root.go
+++ b/internal/cli/root.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


tmchow:
@hnshah i merged in a whack of fixes to the machine based on issues that @pejmanjohn found. Can you rebase your branch off main?

hnshah:
Rebased on latest main ✅ 

- Square URL fix dropped (catalog removed on main)
- Asana URL fix dropped (already correct on main)  
- Content-validity check + tests retained — all tests pass

PR #408, opened 2026-04-30
fix(skills): hoist polish AskUserQuestion load to STEP 0

tmchow:
## Summary

PR #399 added a prose hint in the Setup section telling the polish skill's forked agent to load \`AskUserQuestion\` via \`ToolSearch\`. The fix tested green once on weather-goat and then failed silently on scrape-creators — the forked agent read past the Setup paragraph, hit the divergence-check prompt site, scanned its tool list, didn't see \`AskUserQuestion\`, and fell back to plain-text "reply 1 or 2" prompts the user couldn't answer (their reply lands in the parent session, never reaches the fork, the workflow stalls).

The lesson: prose-instruction reliability for forked agents is fundamentally weak. The agent can read SKILL.md and ignore an instruction it doesn't connect to the moment a tool-availability check fires.

## Changes

- **New "STEP 0" callout** immediately after the title heading, ahead of any other prose. Names the exact action (\`ToolSearch select:AskUserQuestion\`), explains why the tool looks missing initially (deferred), and spells out the failure mode (plain-text reply lands in parent, fork stalls).
- **Trimmed the existing Setup-section paragraph** to a back-reference pointing at Step 0. The load step still appears in the spot agents naturally reach reading top-down, but doesn't repeat the rationale.
- **Inlined a load-and-don't-fall-back reminder at the two prompt sites that have actually fired**: divergence check and publish offer. Even an agent that skipped Step 0 sees the instruction at the moment it's about to make the wrong choice.

## Diff size

+5/-3 across one file.

## If this also fails

The architectural fix is to accept \`--divergence={ask,sync,proceed}\` and similar so the parent orchestrates user decisions and the fork is autonomous. We discussed this earlier in the original session; deferred because the prose fix tested green once. If the louder prose still fails on a real run, the flag-based pattern is the next step.

## Test plan

- Re-run \`/printing-press-polish scrape-creators\` after this merges and reload. Confirm divergence check fires \`AskUserQuestion\` interactively rather than the plain-text fallback.

---

Claude Code-D97757?logo=claude&logoColor=white)

### DIFF
--- a/skills/printing-press-polish/SKILL.md
+++ b/skills/printing-press-polish/SKILL.md
@@ -21,6 +21,8 @@ allowed-tools:
 
 # /printing-press-polish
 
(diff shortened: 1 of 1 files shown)


tmchow:
Superseded by #410. The architectural fix (drop context: fork) makes the prose-only STEP 0 patch unnecessary.

PR #430, opened 2026-04-30
feat(cli): WU-2 sync correctness pass (pagination, ID extraction, exit policy, write serialization)

### LINKED ISSUE #421: Retro from PokéAPI generation: 12 findings, 7 work units (P1: collision detection, sync correctness)
# Printing Press retro: PokéAPI

Generated a fresh print of [pokeapi-pp-cli on v3.0.1](https://github.com/mvanhorn/printing-press-library/pull/158) (replacing a v2.3.6 contribution). 13 transcendence features, 98 MCP tools, 84/100 Grade A, 5/5 shipcheck legs PASS, 253/253 full dogfood.

The run surfaced **12 findings across 7 work units** that would help every future CLI. Most importantly: a generator-level collision-detection gap (PokéAPI's spec resource `version` shadowed the framework's `version` subcommand), several sync-correctness bugs, and one high-leverage missing scaffolding (no template for novel commands — every CLI re-writes the same ~80-LOC cobra-command boilerplate).

## Artifacts

- 📦 [Manuscripts (research brief, absorb manifest, shipcheck, acceptance, retro doc)](https://files.catbox.moe/oofkz8.zip) (~75 KB)
- 📦 [CLI source (Go code, no binaries)](https://files.catbox.moe/fxozdp.zip) (~408 KB)

### F1 — Spec resource collisions with framework reserved commands
PokéAPI's `/api/v2/version/` resource caused the generator to emit a `version` cobra command that shadowed the framework's built-in version printer. Required a manual rename of `Use: "version"` → `Use: "game-version"` post-generation.

**Durable fix:** In `internal/openapi/parser.go` resource-name extraction, maintain a `frameworkReservedCommands` set (`version, auth, doctor, completion, help, agent-context, sync, search, sql, profile, feedback, which, meta, analytics, api, export, import, workflow`). Auto-rename to `<api-slug>-<resource>` when a collision is detected, with a build-time warning.

### F2 — Sync `--max-pages 0` is not actually unlimited
With `--max-pages 0` and even `--max-pages 200`, sync still capped at 100 items per resource. Silent data truncation across every sync-backed CLI.

### F3 — `SQLITE_BUSY` errors during sync at any concurrency > 1
WAL mode is enabled but the upsert path takes a write lock per row; concurrent workers race. Forced `--concurrency 1` for the run.

**Durable fix:** Single-writer goroutine — workers fetch in parallel, upserts dispatch onto a buffered channel, one goroutine drains the channel and serializes writes.

### F4 — Sync silently drops rows for resources with non-`name` primary keys
`pokemon-species`, `pokemon-form`, `evolution-chain` reported "100 fetched" but stored 0 rows. The ID-extraction logic appears to look only at the `name` field; resources with `id`-only primary keys silently get dropped during upsert. **No `sync_warning` was emitted for the silent-drop case** — only `super-contest-effect` and `characteristic` got the explicit `all_items_failed_id_extraction` warning.

**Durable fix:** Look for `id` AND `name` AND any field labeled as the primary key. Always emit `sync_warning` when fetched > 0 but stored == 0.

### F5 — Sync exit code is non-zero on any partial failure
Sync exits non-zero whenever any resource fails — even `super-contest-effect` failures (which are themselves a separate bug). Makes chaining sync into shipcheck require paranoid `|| true` wrappers.

### F6 — `scorecard --spec` rejects YAML
Required converting the spec to JSON before shipcheck. Most official API specs (PokéAPI, Stripe, Twilio, GitHub) are published as YAML.

### F7 — `search --json` returns empty stdout for no matches
Breaks any agent piping through `json.loads` / `jq`. Should always emit `{"results":[]}` even on no matches.

### F11 — Generator does not emit stub files for novel commands approved in Phase 1.5
Building 13 transcendence features required ~17 hand-written cobra command files of ~80-100 LOC each. Pattern is consistent enough to template.

**Durable fix:** When Phase 1.5 approves novel features, emit a stub per feature in `internal/cli/novel/<feature>.go` with cobra wiring filled in from research.json fields, leaving only the RunE body for the agent to implement.

### F9 — Live-check tokenizer doesn't split on commas
Query `pikachu,charizard,blastoise` is treated as one token. Output containing those names individually (in a JSON array with quote separators) doesn't match. Caused 3 spurious live-check failures in this run.

**Durable fix:** `outputMentionsQuery` should split on whitespace AND commas in `internal/pipeline/live_check.go`.

## P3 findings (bundled briefly)

- **F8** — `dogfood` reimplementation_check requires `internal/store` import; flags raw `database/sql` access. Easy workaround but a false positive.
- **F10** — Live-check's `extractQueryToken` treats English-word command names (`find`, `top`) as queries. Should consult the cobra tree, not just the last positional.
- **F12** — No `--rename old=new` flag on `generate`. Mostly subsumed by F1's auto-detect.

## Work units

7 self-contained work units. Each has a goal, target component, acceptance criteria (positive + negative), and complexity sizing. See the retro document in [the manuscripts zip](https://files.catbox.moe/oofkz8.zip) for the full breakdown.

| # | Title | Findings | Component | Complexity |
|---|-------|----------|-----------|------------|
| WU-1 | Framework-collision detection in resource-name extraction | F1, F12 | Generator (parser + templates) | medium |
| WU-2 | Sync correctness — pagination, DB serialization, ID extraction, exit codes | F2, F3, F4, F5 | Binary + parser | large (4 sub-fixes) |
| WU-3 | Spec format flexibility (YAML support across shipcheck legs) | F6 | Binary + shared loader | small |
| WU-4 | `--json` contract for empty result sets | F7 | Generator templates | small |
| WU-5 | Novel-command stub generation from research.json | F11 | Generator + Phase 3 skill | medium-large |
| WU-6 | Live-check tokenizer correctness | F9, F10 | Scorer (live_check.go) | medium |
| WU-7 | Dogfood reimplementation_check accepts raw `database/sql` | F8 | Scorer (reimplementation_check.go) | small |

## What the Printing Press got right

- The runtime cobra-tree MCP mirror is excellent — adding 17 novel commands automatically expanded the MCP surface to 115 tools without any manifest editing. `mcp:read-only` annotations are intuitive.
- Phase 1.5 (Ecosystem Absorb Gate) is the right shape. Stopping for explicit user approval saved one round-trip when the user redirected scope mid-gate (drop battle-sim, add damage-calc + stat-ranker).
- The fix-now contract surfaced real bugs (SQL recipe with non-existent columns) that would have shipped silently.
- `auth_type: none` autodetection worked perfectly. Doctor reported "Auth: not required" without intervention.
- The publish path was clean. Merged-collision detection correctly identified the existing v2.3.6 contribution, and the registry update auto-sorted alphabetically.

## Anti-patterns to avoid

- Don't skip the `version` rename for "convenience" once F1 lands. Even if it works for one user, it shadows a contract and harms others.
- Don't change scorer thresholds to satisfy a CLI's score. F9 and F10 are scorer bugs to fix, not knobs to tune.
- Don't add per-API special cases to the sync upsert path. F4's fix should be format-agnostic.

---

Generated by /printing-press-retro on 2026-04-30 from run `pokeapi/20260429-230641`.

tmchow:
## Summary

Lands the four sync-correctness fixes from PokéAPI retro WU-2 (#421). Sync code is template-side (`internal/generator/templates/sync.go.tmpl` + `store.go.tmpl`), so every printed CLI inherits the fix at next regen.

- **F2** (`U2`) — raise `--max-pages` default 10 → 1000, structured `sync_warning` on cap-hit (both flat and dependent-resource paths), sticky-cursor detector breaks loop when `nextCursor` doesn't advance
- **F3** (`U5`) — `sync.Mutex` on every store write path (`UpsertBatch`, `SaveSyncState`, `SaveSyncCursor`, `ClearSyncCursors`, `Upsert`, typed `Upsert<Pascal>`, `migrate`) eliminating SQLITE_BUSY at any concurrency
- **F4a** (`U1` + `U3`) — profiler-driven primary key via new `x-resource-id` extension with four-tier fallback chain (`x-resource-id → id → name → first required scalar`); reduced runtime fallback list (kalshi-specific names dropped); per-item `sync_anomaly` (`reason: "primary_key_unresolved"`) when extraction fails
- **F5** (`U4`) — new `x-critical: true` extension for spec authors; new `--strict` flag for backward compat; in-band one-shot `sync_warning` (`reason: "exit_policy_default_changed"`) when the new default suppresses what would have been a non-zero exit, so CIs that watch sync output can discover the contract change
- Bonus: F4b symptom probe (`reason: "stored_count_zero_after_extraction"`) catches the held-out F4b case if it recurs without trying to fix the root cause

## Sequenced commits

| Commit | Unit | What it does |
|---|---|---|
| `a9091337` | **U1** | Profiler/parser read `x-resource-id` and `x-critical`; `SyncableResource` populates `IDField` and `Critical` with four-tier fallback chain |
| `3a152450` | **U5** | `sync.Mutex` on store write paths; eliminates SQLITE_BUSY at default concurrency. Audit grep confirms every `s.db.Exec` / `s.db.Begin` is inside a wrapped function |
| `7a872093` | **U2** | `--max-pages` default 10 → 1000; structured cap-hit warnings on both flat and dependent paths; sticky-cursor detector |
| `b1b2cf47` | **U4** | `--strict` flag, `criticalResources` map literal, four-branch exit-code policy, in-band default-flip signal |
| `e9699183` | **U3** | Profiler-driven `IDField` override map; reduced fallback list; per-item `primary_key_unresolved` warnings; F4b symptom probe; `currencies` fixture exercises the runtime fallback |
| `76b9660f` | golden | Lock the new contracts into `testdata/golden/expected/generate-golden-api/` |

## Behavior changes (call out in CHANGELOG)

- **`--max-pages` default raised from 10 to 1000.** Most real-world syncs now complete without truncation. Cap-hits emit a structured `sync_warning` so empirical re-tuning is feasible.
- **Sync exit code on partial failure changes from non-zero to 0.** `--strict` flag preserves legacy behavior. New default emits a one-shot `sync_warning` (`reason: "exit_policy_default_changed"`) so CIs watching sync output discover the change without reading the CHANGELOG.
- **Kalshi-specific fallback names (`ticker`, `event_ticker`, `series_ticker`) dropped from the runtime PK list.** The kalshi CLI maintainer (the user opening this PR) owns it and will regenerate with `x-resource-id` annotations as part of landing this WU. No other public-library CLIs depend on these names.

## Spec author additions

Two new OpenAPI extensions:

- `x-resource-id: <field>` on a path-item — declare the primary-key field for a resource. Recommended on every spec. Without it, sync falls back through `id → name → first required scalar in response schema → runtime list`.
- `x-critical: true` on a path-item — declare a resource as essential. Sync exits non-zero if a critical resource fails, regardless of `--strict`.

## Verification

| Check | Result |
|---|---|
| `go build ./...` | PASS |
| `go vet ./...` | clean |
| `golangci-lint run ./...` | 0 issues |
| `go test -race ./...` | PASS (2423 tests across 28 packages) |
| `scripts/golden.sh verify` | PASS (9/9 cases) |

## Document review

The plan was reviewed by 4 personas (coherence, feasibility, product-lens, adversarial). Cross-persona agreement on the two P0 findings (mutex scope incomplete, silent exit-code flip) led to enumeration of every write method to wrap (U5 audit grep) and the in-band default-flip signal (U4). All 15 actionable findings were auto-resolved per the user's choice; FYI items left as-is. Plan file at [`docs/plans/2026-04-30-001-feat-sync-correctness-wu2-plan.md`](docs/plans/2026-04-30-001-feat-sync-correctness-wu2-plan.md).

Closes #421.


tmchow:
### Code review

Found 3 issues:

1. `ResolveByName` ambiguous-match hint regressed from `"..."` to `".."`. The simplify commit's regex meant to strip `WU-2` markers from comments also collapsed canonical Go ellipses in user-facing strings. Pre-PR (`main`) emits `"..."`; this PR emits `".."`. Visible in every printed CLI's `<resource> resolve` ambiguous-match error path.

https://github.com/mvanhorn/cli-printing-press/blob/44a7b71efcc8722256d262706a5040d30c37e8d6/internal/generator/templates/store.go.tmpl#L843-L849

2. `DependentResource` (parent-child sync) was not extended with `IDField` and `Critical`, so the new `x-resource-id` and `x-critical` annotations are silently ignored on dependent resources. Templates iterate only `.SyncableResources` when building `resourceIDFieldOverrides` / `criticalResources`, so an annotation on a `/parents/{id}/children` path-item never reaches the runtime override or critical-resource maps. Same retro framing (F4a) was kalshi's silent-drop on event/series children — exactly the kind of dependent resources users will want to annotate.

https://github.com/mvanhorn/cli-printing-press/blob/44a7b71efcc8722256d262706a5040d30c37e8d6/internal/profiler/profiler.go#L73-L81

3. The cobra `Long:` help text documents an exit-code contract the new RunE no longer honors: it still says `The command exits non-zero only when every selected resource was access-denied or any resource hit a hard error`, but PR #430 changes the default to exit 0 on partial failure (the in-band `exit_policy_default_changed` warning fires but the process exits successfully). The Long: text was not updated to mention `--strict` or the `x-critical` opt-in. AGENTS.md says "When a machine change alters what a command guarantees, update [the user-visible contract] in the same change."

https://github.com/mvanhorn/cli-printing-press/blob/44a7b71efcc8722256d262706a5040d30c37e8d6/internal/generator/templates/sync.go.tmpl#L48-L60

PR #434, opened 2026-04-30
feat(cli): cost-based throttling primitives for GraphQL CLIs

cathrynlavery:
## Summary

Adds generator-side primitives so any cost-throttled GraphQL CLI can opt into proactive, budget-aware rate limiting. The generic harness — bucket math, retry, `--throttle-mode` flag — is shape-agnostic; per-API parsers are selected via `Throttling.Shape` (only `"shopify"` ships in v1, but adding `github-graphql` or `datadog` becomes a new gated block plus a constant, not a core refactor).

The primitives are forward-projecting (consume the bucket the API reports back, sleep on deficit) rather than reactive — they compose with `cliutil.AdaptiveLimiter`'s 429-driven RPS pacing, they don't replace it.

Default off: existing GraphQL CLIs (Linear) and REST CLIs regenerate byte-identical when `Throttling.Enabled` is unset.

## Updates since the original PR (commits `df7bc1d5` + `2b338a7a`)

- **CI lint fix.** `internal/spec/spec.go:108` had `json:"throttling,omitempty"` on a struct value; modernize lint correctly flagged that `omitempty` has no effect on Go struct values. Dropped `,omitempty` from the json tag so it matches the convention used by `Cache`/`Share`/`MCP` siblings.
- **Named-adapter shape.** Original PR hardcoded the Shopify parser shape and noted "pluggable per-spec is a follow-up." Promoted that into v1: `ThrottlingConfig` now carries a `Shape` field; only `"shopify"` is valid today, but the Shopify parser block in `graphql_client.go.tmpl` is gated on `Shape == "shopify"` so adding GitHub or Datadog support is a new gated block, not a refactor of the existing one. Spec validation rejects `Enabled=true` without a recognized Shape (catches typos at spec-load time, not silently no-op at runtime).
- **Pinning + symmetry polish.** `validateThrottling` is a free function alongside `validateMCP`/`validateExtraCommands`. `TestThrottleShapeShopifyValue` pins the constant's wire value to `"shopify"` so a silent rename would surface as a test failure (Go templates can't import the Go constant, so the template gate uses the literal string).

## What's in this PR

- **Spec opt-in** — `Throttling.Enabled` + `Throttling.Shape` on `APISpec`, queried via `HasCostThrottling()`. Default off; existing GraphQL CLIs (Linear) and REST CLIs regenerate byte-identical when unset. `validateThrottling` rejects `Enabled=true` without a recognized Shape.
- **`internal/client/throttle.go`** (new template) — `ThrottleStatus`, `ThrottleState` (mutex-guarded), `WaitForBudget`, `Update`, `HandleThrottleError` (max 5 retries, exponential backoff with jitter), `ThrottleMode` constants. Unseeded init returns immediately so the API can teach us our budget.
- **`graphql_client.go.tmpl` updates** — when throttling is enabled, `Query` routes through `queryWithThrottle`: `awaitBudget` → `Post` → per-Shape parser (Shopify-shape today, gated on `Shape == "shopify"`) → retry on `THROTTLED` extension code via `HandleThrottleError`. HTTP 429s still flow through the existing `client.go` AdaptiveLimiter retry — no double-handling.
- **`--throttle-mode {strict,lenient,off}`** — wired to root command. Strict (default) blocks via `WaitForBudget`. Lenient warns to stderr and proceeds. Off bypasses entirely. Validated in `PersistentPreRunE`.

## Opt-in mechanism

```yaml
throttling:
  enabled: true
  shape: shopify
```

`ThrottlingConfig{Enabled bool, Shape ThrottleShape}` on `APISpec`. Spec validation runs at parse time so misconfigurations (`enabled` without `shape`, unknown shape) fail-fast instead of silently producing a no-op parser at runtime. The struct future-proofs to per-API knobs (max retries, custom defaults) without ad-hoc spec-level fields.

## Default mode

`--throttle-mode strict`. First-time runs see correct behavior — a slow user (low budget) blocks for tens of seconds before any query, but the alternative is a confusing 429 cascade. Scripted callers can opt to `lenient` (warn but proceed) or `off`. Documented in `--help` text.

## Composition with AdaptiveLimiter

The two coexist. `cliutil.AdaptiveLimiter` paces RPS reactively from observed 429s; `ThrottleState` projects forward from the bucket the API reports. Stack:

1. AdaptiveLimiter waits per-RPS-budget before any request (existing behavior).
2. WaitForBudget waits per-cost-budget before a GraphQL query (new).
3. On 429: AdaptiveLimiter halves RPS + retries via Retry-After (existing).
4. On `THROTTLED` extension: graphql client retries via `HandleThrottleError` (new).

These compose because they observe different signals (RPS vs cost). The `--throttle-mode off` escape hatch exists if they ever interact badly in practice.

## Codex review

Ran `codex review --base origin/main`. Surfaced two P2 issues, both fixed in `4777d7c`:

1. `WaitForBudget` did not reserve budget — concurrent goroutines all read the same projected balance and raced through. Fixed: `reserveBudget` now debits the requested cost from the local bucket under the same mutex that computes the wait. `computeWait` stays pure (used by lenient mode + tests).
2. `GraphQLResponse.Extensions` was an unused field that drifted byte-compat for existing GraphQL CLIs regardless of throttling opt-in. Dropped — the cost parser reads `extensions.cost` from the raw body via a separate typed struct.

## Byte-compat receipts

`TestGenerateNoCostThrottlingByteCompat` regenerates a Linear-shape GraphQL CLI with throttling off and asserts: no `throttle.go`, no `Throttle`/`ThrottleMode` references in `client.go` or `graphql.go`, no `--throttle-mode` flag in `root.go`. `TestGenerateProjectsCompile` covers the REST byte-compat case (loops/stytch/clerk). Golden harness (`scripts/golden.sh verify`) passes byte-identical for all 9 cases.

## Tests

End-to-end behavioral coverage in `cost_throttling_test.go`. Generates a fixture spec with `Throttling.Enabled` + `Shape: ThrottleShapeShopify`, compiles the generated tree, injects a `throttle_behavior_test.go` package-local test, runs it. Verifies:

- `WaitForBudget` returns immediately when `!seeded`
- `computeWait` projects forward (status 100/1000 with restoreRate 50, requested 800 after 5s elapsed → projected 350, deficit 450, sleep 9s)
- `HandleThrottleError` detects HTTP 429 with `Retry-After`
- `HandleThrottleError` detects GraphQL `THROTTLED` extension code body
- `HandleThrottleError` caps at `MaxThrottleRetries`
- Concurrent `Update`/`computeWait`/`Status` calls are race-free under `-race`
- Off-mode `awaitBudget` short-circuits even when bucket is drained
- `updateThrottleFromBody` parses Shopify's shape; no-ops when `extensions.cost` is missing
- `reserveBudget` debits concurrent callers — two `reserveBudget(50)` calls on a 100-bucket admit immediately, the third waits ~1s

Spec-validation coverage in `internal/spec/spec_test.go`:
- `TestThrottlingValidate` — off case, off-with-stray-shape, valid shopify, enabled-without-shape (rejected), unknown shape (rejected)
- `TestThrottleShapeShopifyValue` — pins `"shopify"` so a constant rename surfaces immediately

## Test plan

- [x] `go test ./...` green (2541 passed)
- [x] `go vet ./...` clean
- [x] `go fmt ./...` clean
- [x] `golangci-lint run ./...` clean (CI lint failure resolved)
- [x] `scripts/golden.sh verify` — all 9 cases byte-identical
- [x] New tests: `TestGenerateNoCostThrottlingByteCompat`, `TestGenerateCostThrottlingPrimitives` (with -race lane on the concurrent test), `TestThrottlingValidate`, `TestThrottleShapeShopifyValue`
- [x] Codex review pass — 2 P2 findings addressed

## Out of scope

- A `catalog/shopify.yaml` spec that opts in (recommended as a follow-on PR — proves the primitives end-to-end against Shopify's actual API)
- Additional shapes (`github-graphql`, `datadog`) — each is a constant + a gated parser block + an extension to `validateThrottling`'s switch
- REST-API cost-bucket headers (when an API exposes one, the surface extends; today the parser is GraphQL-only)


tmchow:
@cathrynlavery hey! Pushed two commits on top of yours and updated the PR description. Quick rundown of what changed and why:

1. **Fixed the lint failure.** `internal/spec/spec.go:108` had `json:"throttling,omitempty"` on a struct value — modernize lint (correctly) flagged that omitempty doesn't do anything on Go struct values. Dropped `,omitempty` from the json tag to match how `Cache`/`Share`/`MCP` are tagged in the same struct.

2. **Promoted the pluggable parser into v1.** Original PR noted "making it pluggable per-spec is a follow-up." Took that on now because the "generic primitives for cost-throttled GraphQL CLIs" framing only really lands once spec authors can pick a shape. Added a `Shape` field on `ThrottlingConfig` with one valid value today (`shopify`); the Shopify parser block in `graphql_client.go.tmpl` is now gated on `Shape == "shopify"`. Adding GitHub or Datadog later becomes a new gated block + a constant, not a refactor of the existing path. `validateThrottling` rejects `Enabled=true` without a recognized Shape, so misconfigurations fail at spec-load instead of silently no-op'ing at runtime.

3. **Polish from a self-review pass.** `validateThrottling` is a free function now (matches `validateMCP`/`validateExtraCommands` in the same file). `TestThrottleShapeShopifyValue` pins the constant's wire value to `"shopify"` so a silent rename of the Go constant would break the test rather than silently breaking the template gate. Trimmed a duplicate doc paragraph between `ThrottlingConfig` and `ThrottleShape`.

Goldens still byte-identical, all 2541 tests pass, lint clean.

PR #457, opened 2026-05-01
refactor(cli): extract body-map block to a shared helper (#450)

### LINKED ISSUE #450: Generator: extract body-map-building block to a shared template partial
## Context

Surfaced by /simplify reuse review on PR #449.

## Problem

The body-map-building template block is now duplicated 4 times across 2 templates:

- `internal/generator/templates/command_endpoint.go.tmpl:217-240` (POST branch)
- `internal/generator/templates/command_endpoint.go.tmpl:266-289` (PUT branch)
- `internal/generator/templates/command_endpoint.go.tmpl:309-332` (PATCH branch)
- `internal/generator/templates/command_promoted.go.tmpl:163-181` (PR #449 added this site)

Each block is ~22 lines of `{{range .Endpoint.Body}}` with the same three-branch shape:
1. `object`/`array` types: parse JSON, store parsed value
2. `jsonStringParam` types: parse JSON, store the raw string
3. Scalar types: zero-check, store directly

The blocks differ only in indentation level. If a fourth body-shape category gets added (e.g. file uploads, multipart), all 4 sites must be updated in lockstep.

## Proposed change

Extract to a `{{define "body_map"}}` partial. Two implementation options:

**Option A — `text/template` named blocks.** Currently each template parses independently via `template.New(tmplName).Funcs(g.funcs).Parse(...)`. To share a partial, either:
1. Move toward `template.ParseFS` with all templates as one parsed set (broader change), or
2. Add a parse-time include helper that reads a `_partials/body_map.tmpl` file and substitutes it.

**Option B — FuncMap helper that returns rendered Go code.** A `bodyMapBuilder` template func that receives the body params and returns a string of Go code. Templates call it with `{{bodyMapBuilder .Endpoint.Body}}`. Less idiomatic than `{{define}}` but works without restructuring how templates are parsed.

## Why deferred from PR #449

The duplication was 3 sites pre-PR (all in `command_endpoint.go.tmpl`); PR #449 added a 4th site in `command_promoted.go.tmpl`. The right time to extract is when the duplication is bothering us enough to do the parsing-architecture work, not as a side-quest in a feature PR. Filed so it doesn't drift past 4 sites without notice.

## References

- PR #449 (added the 4th duplicate site)
- `internal/generator/templates/command_endpoint.go.tmpl`
- `internal/generator/templates/command_promoted.go.tmpl`
- `internal/generator/generator.go:2168` (where templates are parsed today)

🤖 Filed by Claude Code as a follow-up to PR #449's /simplify pass

tmchow:
## Summary

- Extract the 22-line `body := map[string]any{}` building block — duplicated 4× across `command_endpoint.go.tmpl` (POST/PUT/PATCH) and `command_promoted.go.tmpl` — into a single `bodyMap(body, indent)` Go helper registered in the generator's FuncMap.
- Each template now invokes `{{bodyMap .Endpoint.Body "<indent>"}}`, parameterizing the indent so the same logic produces correctly indented output at both the 4-tab endpoint sites and the 3-tab promoted site.
- Add `body_map_test.go` covering the object/array and JSON-string branches that the existing golden harness doesn't exercise.

## Why FuncMap helper instead of `{{define}}` partial

Considered text/template named blocks but rejected: `{{template "name" .}}` doesn't preserve per-line indentation across a multi-line block — only the first line picks up the surrounding context's indent. Handling indent through the data context would require prefixing every line in the partial with `{{.Indent}}`, which is noisier than a Go helper. The FuncMap helper avoids restructuring how templates are parsed (the issue's "parsing-architecture work" note) and gets correct indentation for free.

## Test plan

- [x] `go test ./...` (2585 passed, +9 from new body_map_test cases)
- [x] `go vet ./...`
- [x] `golangci-lint run ./...`
- [x] `scripts/golden.sh verify` — all 9 cases pass byte-identical (proves the helper produces the same output as the previous templates for every API in the harness)
- [x] Unit tests cover all three branches: scalar (string + int zero values), object/array JSON parse, JSON-string validate-but-store-raw, and multi-param concatenation

Closes #450

### DIFF
--- a/internal/generator/body_map_test.go
+++ b/internal/generator/body_map_test.go
@@ -127,3 +127,19 @@ func TestBodyMap_DashIdentifier(t *testing.T) {
 		t.Errorf("expected JSON key with dash preserved, got: %s", got)
 	}
 }
+
+// TestBodyMap_IdentName verifies the dedup pass's output: when IdentName
+// is set (because two params would otherwise collide on the same Go
+// identifier), the variable name uses IdentName but body[key] keeps the
+// wire Name. Without this, the generated CLI would either fail to compile
+// or send the wrong field name to the server.
+func TestBodyMap_IdentName(t *testing.T) {
+	t.Parallel()
+	got := bodyMap([]spec.Param{{Name: "start", IdentName: "StartGT", Type: "string"}}, "\t")
+	if !strings.Contains(got, "bodyStartGT") {
+		t.Errorf("expected variable to use IdentName, got: %s", got)
+	}
+	if !strings.Contains(got, `body["start"]`) {
+		t.Errorf("expected wire key to use Name (not IdentName), got: %s", got)
+	}
+}
--- a/internal/generator/body_map_test.go
+++ b/internal/generator/body_map_test.go
@@ -1,0 +1,129 @@
+package generator
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/mvanhorn/cli-printing-press/v3/internal/spec"
+)
+
+// TestBodyMap pins the rendered Go code for each of the three body-param
+// shapes (object/array, JSON-string, scalar). The generator's golden
+// harness only exercises the scalar branch, so this test guards the
+// other two against silent drift after the bash → helper extraction.
+func TestBodyMap(t *testing.T) {
+	t.Parallel()
+
+	cases := []struct {
+		name   string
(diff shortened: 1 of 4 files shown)


tmchow:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

PR #467, opened 2026-05-01
fix(generator): gate table creation on hasDomainUpsert

tmchow:
## Summary

Brings table creation in line with typed-Upsert generation. Both now share the same gate:

```
(gt (len .Columns) 3) (ne .Name "sync_state") (hasDomainUpsert .Name)
```

Previously tables were created unconditionally while typed Upserts were gated by column count. The asymmetry produced dead tables: resources whose names collide with framework cobra commands (`auth`, `search`, `version`) get renamed `<api-slug>-<name>`, the renamed table ends up with only 3 columns (extraction runs against the original name and finds nothing), the typed Upsert gets filtered out by the column gate, UpsertBatch's dispatch has no branch for the renamed resource — so the typed table gets created but never written to.

## What this unblocks

Surfaced by the SIG-DRIFT-LIGHT sweep work on cal-com, pokeapi, coingecko, docker-hub. Each had pub-side `Upsert<X>` + `case "<x>":` switches that fresh stopped emitting because of this gate, leaving 3-col tables that nothing populated. After this fix, the dead tables simply don't exist; data still flows via UpsertBatch + upsertGenericResourceTx into the generic `resources` table.

## Verification

- `go test ./...` — 2621 pass (1 new test added)
- `go vet`, `golangci-lint` clean
- Golden verify passed
- Regenerated cal-com — 6 dead tables removed: `api_keys`, `cal_com_auth`, `cal_com_auth_2`, `conferencing`, `destination_calendars`, `organizations`

Test: `TestStoreSkipsDeadTablesForResourcesWithoutTypedUpsert` generates a spec with both a typed-column resource and a collision-renamed bare resource, asserts the renamed dead table is not created and the typed-column resource keeps its table.

## Test plan

- [x] Existing test suite passes
- [x] Golden tests updated (no diff this round — change is purely conditional emission)
- [ ] Library follow-up: re-sweep the deferred CLIs to drop their dead typed-Upsert helpers and tables

### DIFF
--- a/internal/generator/generator.go
+++ b/internal/generator/generator.go
@@ -297,6 +297,15 @@ func New(s *spec.APISpec, outputDir string) *Generator {
 		"hasDomainUpsert": func(name string) bool {
 			return domainUpsertMethodName(name) != "UpsertBatch"
 		},
+		// hasTypedTable is the single source of truth for "this table gets a
+		// typed Upsert<X>." Table creation, typed-Upsert generation, the
+		// UpsertBatch dispatch switch, and the populated-table tests must all
+		// gate on the same predicate; otherwise dead tables (created but never
+		// written to) leak in for resources whose names hit the framework-cobra
+		// rename and end up with only id/data/synced_at columns.
+		"hasTypedTable": func(t TableDef) bool {
+			return len(t.Columns) > 3 && t.Name != "sync_state" && domainUpsertMethodName(t.Name) != "UpsertBatch"
+		},
 		"pathContainsParam": func(path, name string) bool {
 			return strings.Contains(path, "{"+name+"}")
 		},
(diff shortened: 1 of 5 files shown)


tmchow:
Simplify pass applied:

1. **Extracted gate into a `hasTypedTable` FuncMap helper** at `internal/generator/generator.go:300`. The 6 template sites that all checked `(gt (len .Columns) 3) (ne .Name "sync_state") (hasDomainUpsert .Name)` now collapse to `{{- if hasTypedTable .}}` — a future contributor adding a 4th clause can change one Go function instead of remembering all six template files.
2. **Tightened the comment block** in `store.go.tmpl` from 8 lines to 3.
3. **Strengthened test assertions** — replaced `assert.NotContains` with a count-based check that pins the exact CREATE TABLE set (`resources`, `sync_state`, `items`) and asserts `upsertGenericResourceTx` is still invoked so renamed resources keep flowing through the generic path.

Verified: 2621 tests still pass, vet/lint clean, golden verify clean.

PR #488, opened 2026-05-02
fix(scorer): live-check treats CLI's graceful empty-handling as PASS, not FAIL

### LINKED ISSUE #484: scorer: live-check treats novel_features.example placeholder slugs as authoritative, fails 5/12 on producthunt
`scorecard --live-check` samples each novel feature by running its `example` field from `research.json`. When the example uses a placeholder slug that doesn't exist on the live API, the sample fails — counted as a CLI bug even though the CLI works correctly.

## What I observed

producthunt's first live-check after rate limit cleared:
```
Live Check (behavioral sample)
  Passed: 7/12  (58% pass rate)
  Failures:
    - Launch-day tracker: exit 1: Error: post not found: my-launch-slug
    - Vote trajectory: exit 1: Error: post not found: my-launch-slug
    - Comments → genuine questions: output does not contain any token from query "my-launch-slug"
    - Side-by-side launch compare: exit 1: Error: post not found: cursor-ide
    - Lookalike launches: output does not contain any token from query "notion"
```

All 5 failures are research.json's `novel_features[].example` field using slugs that the LLM-authored examples treated as illustrative ("my-launch-slug"), or slugs that exist on PH but as something different than the example implies (`notion` returns the original 2014 launch, not a current Notion product).

The sampler interpreted these as CLI failures, dragging the apparent pass rate from 12/12 to 7/12.

## Root cause: novel_features.example field is LLM-authored prose, not test data

From the printing-press SKILL.md guidance for novel_features:
> `example` is a ready-to-run invocation an agent can copy-paste. Use realistic arguments from the API's domain (e.g. `AAPL`, `customer_42`), not `<placeholder>`. Include the `--agent` flag when the feature benefits from structured output.

This says "use realistic arguments" but doesn't say "use slugs that exist right now on the live API." For static reference data (`AAPL`, `customer_42`) this works; for time-series content like Product Hunt launches, today's top-ranked slug rotates daily.

## Fix options

**Option A — sampler-side:** Allow `novel_features` entries to mark `example` as `placeholder: true` or use a separate `live_test_args` field with known-stable slugs. Sampler skips entries marked placeholder when computing the pass rate.

**Option B — skill-side:** Tell the LLM authoring novel_features to mark examples that use rotating content with a `# placeholder` comment or to use known-stable slugs (e.g. for PH, `notion-2`/`figma-2`/etc. — products that have been #1 multiple times). This still lies about live correctness for any slug that 404s in the future.

**Option C — sampler-side smart fallback:** When a sample fails with "not found"/"no results" against the live API, distinguish "CLI failed" (exit code 1 with a CLI-level error) vs "API returned empty" (exit code 0 with empty results). Currently both look like failures. Many of producthunt's failures (3/5) were "post not found" which is the API correctly returning nothing for a non-existent slug — not a CLI bug.

Option C is the most agent-grade: it would correctly score the CLI on whether it handles "no results" gracefully (which producthunt does — returns a clear `post not found: <slug>` error and exits non-zero, exactly the right behavior).

Together Option B + Option C would make the scorer correctly distinguish "broken CLI" from "broken example data" from "API has no matching record."

## Why this matters beyond producthunt

Any CLI for a content-rotating API (Product Hunt, Hacker News, Reddit hot, GitHub trending, etc.) hits this. The sampler currently penalizes them for using realistic-looking example slugs that decay.

## Evidence
- Run: `~/printing-press/.runstate/linear-herding-lollipop-44b2f933/runs/20260501-170103/`
- Polish scorecard output: `Passed: 7/12 (58% pass rate)` — but Live API Verification still scored 10/10 because the cap fires at >=3 passes. So the headline score wasn't affected here, but the "5 failures" line is misleading and would mislead a reviewer about CLI quality.
- Public PR: https://github.com/mvanhorn/printing-press-library/pull/184

tmchow:
## Summary

`scorecard --live-check` invoked the CLI with `novel_features[].example` args verbatim and treated **any non-zero exit as a CLI failure** — even when the CLI was correctly reporting "your input maps to no record on the live API." On producthunt this turned 3 of 5 reported failures into noise: each one was the CLI gracefully printing `Error: post not found: my-launch-slug` and exiting 1, which is the right behavior for a placeholder/decayed slug from research.json's LLM-authored example field.

This fixes the conflation by requiring **two** signals before treating non-zero exit as PASS: (1) stderr matches a known empty-result phrase, AND (2) stderr echoes one of the user-supplied positional args. The arg-echo gate is the boundary that keeps generic auth/config errors failing.

Closes #484 (H1 — see "Out of scope" below).

## What changed

`internal/pipeline/live_check.go`:

- New `isGracefulEmptyResponse(exitCode, stderr, args)` helper.
- New `positionalCandidates(args)` helper — filters out flags but intentionally keeps flag values (so `--query notion` → "notion" is still a valid arg-echo signal for query-driven features).
- New branch in `runOneFeatureCheck`: when `*exec.ExitError` triggers and the helper matches, set `Status=Pass` with `Reason: "graceful empty: ..."` instead of failing.

`internal/pipeline/live_check_test.go`:

- `TestLiveCheck_PassOnGracefulEmpty` — 5 sub-cases, one per phrase, end-to-end through `RunLiveCheck`.
- `TestLiveCheck_FailOnGenericExitErrorWithoutArgEcho` — 3 sub-cases preserving the original fail contract: phrase-without-echo, no-phrase-no-echo, auth-failure-with-phrase-no-echo all still fail.
- `TestIsGracefulEmptyResponse` — 11 table-driven cases covering exit code, phrase presence, arg echo, short-token threshold (`go` skipped), flag-only args (no positional candidates), case-insensitivity, and the flag-value-as-positional case.

The original `TestLiveCheck_FailOnExitError` (exit 5 with generic stderr) keeps passing — confirms no regression on the negative case.

## Why the arg-echo gate matters

Without it, `"config file not found"` would silently mask broken auth setups across the library. The CLI not finding its config isn't the same shape of contract as the CLI not finding the requested record. The arg-echo signal narrows graceful-empty to "the CLI processed the user's input and reported nothing matched" — which is the only contract the live-check should be praising.

Side benefit: when a flag value carries the query (e.g., `--query notion`), the helper picks that up too. So `recipes search --query notion` returning `"no match for query: notion"` correctly counts as graceful empty.

## Out of scope

The other 2 of producthunt's 5 failures hit `outputMentionsQuery` (line 351), not the exit-code path. Those happen when:

- "lookalike" / "similar" features by design return *different* items than the query
- A rotating slug like `notion` returns valid data that doesn't echo the query token

That's a separate failure mode. Fix needs either `research.json` schema (`placeholder: true` per feature) or a per-feature opt-out (`expects_query_in_output: false`). I'll either spin a follow-up issue or leave the existing #484 open with a "H2" carve-out, your call.

## Test plan

- [x] `go test ./internal/pipeline/ -run 'GracefulEmpty|IsGracefulEmptyResponse|FailOnGenericExitErrorWithoutArgEcho' -v` — 19/19 PASS
- [x] `go test ./internal/pipeline/ -run 'LiveCheck'` — full live_check suite PASS, no regression on `TestLiveCheck_FailOnExitError`
- [x] `go test ./...` — full repo PASS
- [x] `go vet ./...` — 0 issues
- [x] `golangci-lint run ./internal/pipeline/...` — 0 issues
- [x] `scripts/golden.sh verify` — 9/9 PASS

### DIFF
--- a/internal/pipeline/live_check.go
+++ b/internal/pipeline/live_check.go
@@ -338,7 +338,19 @@ func runOneFeatureCheck(cliDir, binaryPath string, f NovelFeature, timeout time.
 	if runErr != nil {
 		var exitErr *exec.ExitError
 		if errors.As(runErr, &exitErr) {
-			return fail(fmt.Sprintf("exit %d: %s", exitErr.ExitCode(), trimOutput(stderrCap.String())))
+			stderr := stderrCap.String()
+			if isGracefulEmptyResponse(exitErr.ExitCode(), stderr, args) {
+				// CLI exited non-zero gracefully on "no record matches this
+				// input" — that's the CORRECT behavior for an unknown slug
+				// (research.json's LLM-authored example slugs decay over
+				// time on content-rotating APIs). Don't penalize a feature
+				// that handled the empty case well. See issue #484.
+				result.Status = StatusPass
+				result.Reason = "graceful empty: " + trimOutput(stderr)
+				result.OutputSample = sampleOutput(stderr)
+				return result
+			}
+			return fail(fmt.Sprintf("exit %d: %s", exitErr.ExitCode(), trimOutput(stderr)))
 		}
 		return fail("run error: " + runErr.Error())
 	}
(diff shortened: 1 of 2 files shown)


tmchow:
### Code review

Found 1 issue:

1. Code comments embed the issue number (`#484`) in three places. AGENTS.md says ["No dates, incidents, or ticket numbers in code comments. Belongs in the PR description and commit message, not the code. Comments stay forever; incidents fade."](https://github.com/mvanhorn/cli-printing-press/blob/bb7950fcdd7b24042115b51aa6646dd9da749dcb/AGENTS.md#L286-L297)

https://github.com/mvanhorn/cli-printing-press/blob/bb7950fcdd7b24042115b51aa6646dd9da749dcb/internal/pipeline/live_check.go#L344-L348

https://github.com/mvanhorn/cli-printing-press/blob/bb7950fcdd7b24042115b51aa6646dd9da749dcb/internal/pipeline/live_check.go#L654-L657

https://github.com/mvanhorn/cli-printing-press/blob/bb7950fcdd7b24042115b51aa6646dd9da749dcb/internal/pipeline/live_check_test.go#L607-L611

PR #505, opened 2026-05-02
feat(cli): JSON:API support for OpenAPI parser (envelope, page[cursor], x-prefix)

cathrynlavery:
## Summary
- flatten JSON:API resource envelopes into usable generated typedef fields
- detect JSON:API page[cursor]/page[size] pagination and follow links.next during sync
- map header apiKey x-prefix into AuthConfig.Format and render {token}-based API key headers
- document x-prefix in SPEC-EXTENSIONS

## Verification
- go test ./...
- scripts/golden.sh verify
- go vet ./...

### DIFF
--- a/docs/SPEC-EXTENSIONS.md
+++ b/docs/SPEC-EXTENSIONS.md
@@ -17,6 +17,7 @@ in the same change as any new `Extensions["x-*"]` lookup in that file.
 | `x-proxy-routes` | `info` | `APISpec.ProxyRoutes` | No |
 | `x-auth-type` | `components.securitySchemes.<name>` | `APISpec.Auth.Type` | No |
 | `x-auth-format` | `components.securitySchemes.<name>` | `APISpec.Auth.Format` | No |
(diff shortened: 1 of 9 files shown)


tmchow:
Pulled the branch and reviewed for code simplification. 3 small follow-up simplifications, all behavior-preserving:

1. **`nextCursorFromLinks` (`sync.go.tmpl`)** — removed the `strings.Index` + `url.ParseQuery` fallback. `url.Parse` already handles unencoded brackets (`?page[cursor]=raw` — verified empirically), so the fallback was unreachable. The unencoded-brackets test still passes through the primary path.
2. **`mapAuth` x-prefix block (`parser.go`)** — flattened from 4-level nested `if`s to 2 by combining the map lookup and string type assertion. Nil-interface-to-string asserts to `("", false)`, so behavior is identical.
3. **`TestIsJSONAPIResource_Discriminator`** — dropped the `tc := tc` loop capture; unnecessary under Go 1.22+ per-iteration scoping (caught by `golangci-lint modernize`).

Pushed as `94f0d82d` on top of your branch, all tests pass.

Issue #525, opened 2026-05-03
Retro: FedEx — 7 findings, 7 work units


tmchow:
## Session Stats

| Metric | Value |
|--------|-------|
| API | FedEx |
| Spec source | Hand-authored internal YAML (no vendor OpenAPI; 51 endpoints across 14 FedEx APIs) |
| Scorecard | 72/100 (Grade B) |
| Verify pass rate | 100% |
| Live pass rate | 5/6 mandatory items (rate shop + address validate + address cache + sql + OAuth lifecycle PASS; 2 endpoint commands had body-shape gaps that are per-CLI) |
| Fix loops | 2 (initial shipcheck + post-Cloudflare-pattern iteration) |
| Manual code edits | 6 substantive (OAuth login, AuthHeader precedence, Track 403 hint, dual-creds path-routing, MCP cache-leak backport, post-regen-merge surgery) |
| Features built from scratch | 13 novel commands, 1 SQLite store package (~530 lines), 1 OAuth client_credentials login flow (~165 lines) |

## What the Printing Press Got Right

- `shipcheck` umbrella verdict caught all the right things at the right times
- `regen-merge` classification correctly identified TEMPLATED-CLEAN vs NOVEL vs TEMPLATED-WITH-ADDITIONS for 100+ files
- `mcp-audit` recommendations were spot-on (per #517 description) — the binary already knows what to do
- Cobratree walker automatically picked up all 13 novel commands as MCP tools
- `cliutil.AdaptiveLimiter` handled FedEx's 1 req/s burst limit without bespoke code
- Lock + promote workflow worked cleanly across both initial and post-iteration promotes
- `tools.go` / `code_orch.go` template emission for the Cloudflare pattern worked perfectly on first generation — gap was upstream skill, not template

## P1 — High priority

| Finding | Title | Component | Frequency | Fix summary |
|---------|-------|-----------|-----------|-------------|
| F1 | OAuth2 `client_credentials` grant + AuthHeader precedence fix | `internal/generator/templates/auth.go.tmpl` + `config.go.tmpl` | subclass: oauth2_client_credentials APIs (FedEx now, future Auth0 / MS Graph / Stripe Connect) | New `auth_client_credentials.go.tmpl` gated on `auth.oauth2_grant=client_credentials`; reorder `AuthHeader()` so cached AccessToken takes precedence over env-var-as-bearer fallback |

## P2 — Medium priority

| Finding | Title | Component | Frequency | Fix summary |
[...]

PR #528, opened 2026-05-03
feat(cli): OAuth2 client_credentials grant + bearer_token AuthHeader precedence fix

tmchow:
## Summary

Implements WU-1 from the FedEx retro: OAuth2 `client_credentials` grant (2-legged server-to-server flow) for the printing-press generator, plus a precedence fix in the generated `bearer_token` `AuthHeader()` so a cached `AccessToken` wins over the env-var-as-bearer fallback.

Why this matters: APIs like FedEx, Auth0 Management, and Microsoft Graph daemon apps use `client_credentials` where the env var is the **Client ID**, not a usable bearer JWT. The previous env-var-first precedence sent the Client ID as `Bearer <client-id>` on every request and 401'd.

Cloes #525 

## Implementation units

- **U1** \`feat(cli): AuthConfig.OAuth2Grant field + validation\` — new spec field \`auth.oauth2_grant\` with constants \`authorization_code\` (default) and \`client_credentials\`, plus \`EffectiveOAuth2Grant()\` helper and \`validateOAuth2Grant()\` mirroring existing throttling/transport patterns.
- **U2** \`feat(cli): OpenAPI parser detects oauth2 client_credentials flow\` — \`flows.clientCredentials\` wins when both \`authorizationCode\` and \`clientCredentials\` are declared (most M2M APIs declare both for completeness).
- **U3** \`feat(cli): generator emits OAuth2 client_credentials login command\` — new \`auth_client_credentials.go.tmpl\` emits \`auth login\` (form-encoded POST), \`auth status\`, \`auth set-token\`, \`auth logout\`. Selected when \`OAuth2Grant=client_credentials && TokenURL!=""\`.
- **U4** \`feat(cli): config + client templates honor oauth2 client_credentials\` — \`AuthHeader()\` precedence flip for \`bearer_token\` (AccessToken wins over env var); grant-aware refresh in \`client.go.tmpl\` (\`needsClientCredentialsMint\` + \`mintClientCredentials\` re-mint inside a 60s safety window).
- **simplify** \`refactor(cli): simplify OAuth2 client_credentials surface\` — drop local \`truncateForError\` for \`cliutil.SanitizeErrorBody\` (also redacts credential-shaped strings); drop unused \`tokenResponse\` fields; trim over-documented comments.

## Backward compatibility

- \`OAuth2Grant\` defaults to \`authorization_code\`; existing OAuth CLIs continue to use the existing \`auth.go.tmpl\` (3-legged callback flow). Verified by \`TestGenerateOAuth2AuthorizationCodeRegression\` and \`TestGenerateOAuth2AuthorizationCodeClientRefreshUnchanged\`.
- The \`bearer_token\` precedence flip changes order so AccessToken wins; CLIs where the env var IS a usable bearer JWT (GitHub PATs, etc.) still work via the same fallback — the only behavior difference is that an explicitly-set AccessToken is no longer ignored.
- \`cookie\` and \`composed\` types keep env-first precedence; verified by \`TestAuthHeader_EnvVarWinsOverFileTokenForCookieComposed\`.

## Test plan

- [x] \`go test ./...\` (2804 passing across 30 packages)
- [x] \`go fmt ./...\`
- [x] \`go build -o ./printing-press ./cmd/printing-press\`
- [x] New unit tests cover spec validation (6 cases), \`EffectiveOAuth2Grant\` (4 cases), parser branches (3 tests including precedence when both flows declared), and generator output (4 tests covering template selection, regression, and grant-aware refresh)
- [ ] Regenerate FedEx CLI against this branch and verify token endpoint round-trips correctly (manual, post-merge)


tmchow:
### Code review

Self-review pass via /review surfaced several findings. Two were fixed in [b200d97e](https://github.com/mvanhorn/cli-printing-press/pull/528/commits/b200d97e6572abbee01afec05188e9503cb94111); two are deferred as follow-ups.

**Fixed in b200d97e:**

1. **Bearer_token precedence regression for non-client_credentials CLIs.** The U4 commit's AuthHeader flip was emitted unconditionally for `Type=bearer_token`, regressing the env-first convention established in [`aab364be`](https://github.com/mvanhorn/cli-printing-press/commit/aab364be) (1 day before this PR) for plain bearer / PAT-style flows. A user who rotated their env var would see it silently shadowed by a stale on-disk AccessToken. Now gated on `EffectiveOAuth2Grant=client_credentials`.

   https://github.com/mvanhorn/cli-printing-press/blob/b200d97e6572abbee01afec05188e9503cb94111/internal/generator/templates/config.go.tmpl#L148-L201

2. **`generator.go` read `OAuth2Grant` raw instead of via `EffectiveOAuth2Grant()`.** The spec.go doc mandates the helper (`AGENTS.md` Code Hygiene rule on default-lives-in-one-place). Harmless today (empty falls through), but a maintenance hazard.

   https://github.com/mvanhorn/cli-printing-press/blob/b200d97e6572abbee01afec05188e9503cb94111/internal/generator/generator.go#L1597-L1610

3. **Misleading doc comment on `mintClientCredentialsToken`** claimed it was used by both the auth login command and the client refresh path; the refresh path has its own parallel `mintClientCredentials` method. Comment trimmed.

**Deferred follow-ups:**

4. **Golden fixture not extended** for the new auth shape. Per `AGENTS.md` Golden Output Harness: "When the machine gains deterministic generation capabilities... new auth shapes... extend this fixture and add the smallest useful artifact comparison." Worth a separate PR adding a `golden-api-oauth2-cc.yaml` fixture and matching case.

5. **`mintClientCredentials` race under high concurrency** (no mutex around the 60s-window mint check). Planner accepted as out-of-scope — extra mints under the race, not corruption (file write is atomic). Adding `sync.Mutex` is real correctness work but warrants its own PR.

Other minor findings (parser scope-merging when both flows declared, expires_in=0 edge case, vendor names in const docstrings) scored below the noise floor for action.

[Build green: 2805 tests pass. Golden verify: 9 cases pass.](https://github.com/mvanhorn/cli-printing-press/actions/workflows/test.yml)

PR #574, opened 2026-05-04
docs(cli): consolidate AGENTS.md via progressive disclosure

cathrynlavery:
This PR consolidates `AGENTS.md` because Codex was warning on repo open that the agent instruction file was too large.

The goal here is to keep the always-loaded instruction surface smaller without making the work worse: the write-time rules stay inline in `AGENTS.md`, and the longer reference material moves into focused docs under `docs/`.

That keeps the important guidance in front of the agent at the moment it needs it, while reducing the amount of repo-specific prose that loads on every turn.


docs/CATALOG.md
```diff
@@ -0,0 +1,20 @@
+# Catalog Entry Validation
+
+Catalog entries in `catalog/` are validated by [`internal/catalog/catalog.go`](/Users/cathrynlavery/Developer/mvanhorn/cli-printing-press/internal/catalog/catalog.go). Keep the inline `AGENTS.md` rule in sync with that validator; when the validator's applicability or allowed values change, update the inline trigger sentence in the same PR.
```
tmchow: @cathrynlavery fix the absolute path, this won't work for anyone else :)

tmchow (dismissed):

docs/CATALOG.md
cathrynlavery: amateur hour over here 🙄

PR #647, opened 2026-05-06
fix(phase5): explain accepted marker levels

### LINKED ISSUE #644: lock promote phase5 gate rejects unrecognized 'level' values without documenting the allow-list
## Problem

`printing-press lock promote` validates `phase5-acceptance.json` and refuses to promote when the marker's `level` field is anything other than `quick` or `full`. The error message names the rejected value but does not document the accepted set, so callers writing the marker manually (e.g., when `dogfood --live` can't be used because of [the SKIP-as-FAIL issue tracked separately]) hit a wall.

## Reproduction

Write `phase5-acceptance.json` to `$PROOFS_DIR` with these values (a manual matrix run, not a `dogfood --live` invocation):

```json
{
  "schema_version": 1,
  "api_name": "<api>",
  "run_id": "<run>",
  "status": "pass",
  "level": "manual-flagship",
  "matrix_size": 14,
  "tests_passed": 14,
  "tests_failed": 0,
  "auth_method_proved": "none",
  "completed_at": "<iso>"
}
```

Run `printing-press lock promote --cli <api>-pp-cli --dir <dir>`. Observed:

```
Error: promoting CLI: phase5 gate failed: unknown phase5 acceptance level "manual-flagship"
```

The error names the rejected value but doesn't enumerate what *would* be accepted. The user has to grep the source (`internal/pipeline/phase5_gate.go` or similar) to discover the allow-list is `{quick, full}`.

## Why this matters

When `dogfood --live` works, the `--write-acceptance` flag emits `level: "quick"` or `"full"` and the gate accepts it. But when `dogfood --live` *can't* be used (e.g., a SKIP-on-no-positional flake, or the matrix runner doesn't fit the CLI's shape), the manual fallback has no documented set of valid `level` values to choose from. The user picks something descriptive ("manual-flagship", "manual-quick", "custom"), gets a hard reject, and has to read the binary's source to recover.

## Proposed fix

Two options:

**Option A (recommended): widen the allow-list with an "agent-authored" escape hatch**
Add `"manual"` (or `"custom"`) as a documented third valid value. Marker fields like `matrix_size`, `tests_passed`, `tests_failed` already self-describe the run, so the gate doesn't need `level` to be a closed enum to enforce safety.

**Option B: improve the error message**
If the closed enum is the right design, the error should enumerate accepted values:

```
Error: phase5 gate failed: invalid level "manual-flagship"
       Accepted values: quick, full
       (run printing-press dogfood --live --write-acceptance to emit these automatically)
```

## Acceptance criteria

- Either accept a documented `level` value for manual fallbacks (preferred), OR
- The rejection error names the accepted set and points at how to generate a valid marker.

## Cross-reference

Pairs with the SKIP-as-FAIL issue (filed separately) — the two together force a manual fallback path that has no obvious correct value for `level`.

dinakars777:
## Summary

Phase5 acceptance markers now report invalid `level` values with an actionable diagnostic instead of naming only the bad value. The gate remains closed to runner-emitted `quick` and `full` pass markers, and the error points operators back to `printing-press dogfood --live --write-acceptance` for generating the required marker.

## Testing

- `go test ./internal/pipeline ./internal/cli`

Fixes mvanhorn/cli-printing-press#644

### DIFF
--- a/internal/pipeline/live_dogfood.go
+++ b/internal/pipeline/live_dogfood.go
@@ -1088,9 +1088,9 @@ func normalizeLiveDogfoodLevel(level string) (string, error) {
 		return "full", nil
 	}
 	switch level {
-	case "quick", "full":
+	case phase5AcceptanceLevelQuick, phase5AcceptanceLevelFull:
 		return level, nil
 	default:
-		return "", fmt.Errorf("invalid live dogfood level %q (expected quick or full)", level)
+		return "", fmt.Errorf("invalid live dogfood level %q (expected %s)", level, strings.Join(phase5AcceptedAcceptanceLevels, " or "))
 	}
 }
--- a/internal/pipeline/lock_test.go
+++ b/internal/pipeline/lock_test.go
@@ -557,6 +557,40 @@ func TestPromoteWorkingCLI_RequiresPhase5GateForRunstatePromote(t *testing.T) {
 	assert.ErrorIs(t, statErr, os.ErrNotExist)
 }
 
+func TestPromoteWorkingCLI_RejectsManualPhase5Marker(t *testing.T) {
+	tmp := t.TempDir()
+	t.Setenv("PRINTING_PRESS_HOME", tmp)
+	t.Setenv("PRINTING_PRESS_SCOPE", "test-scope")
+	t.Setenv("PRINTING_PRESS_REPO_ROOT", tmp)
+
+	workDir := filepath.Join(tmp, "working", "test-pp-cli")
+	require.NoError(t, os.MkdirAll(workDir, 0o755))
+	require.NoError(t, os.WriteFile(filepath.Join(workDir, "go.mod"), []byte("module test-pp-cli\n\ngo 1.21\n"), 0o644))
+	require.NoError(t, os.WriteFile(filepath.Join(workDir, "main.go"), []byte("package main\nfunc main() {}\n"), 0o644))
+
+	state := NewStateWithRun("test", workDir, "run-manual-phase5", "test-scope")
+	writePhase5GateMarker(t, state.ProofsDir(), Phase5AcceptanceFilename, Phase5GateMarker{
+		SchemaVersion: 1,
+		APIName:       state.APIName,
+		RunID:         state.RunID,
+		Status:        "pass",
+		Level:         "manual",
+		MatrixSize:    1,
+		TestsPassed:   1,
+		TestsFailed:   0,
+		AuthContext:   Phase5AuthContext{Type: "none"},
+	})
+
+	err := PromoteWorkingCLI("test-pp-cli", workDir, state)
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "unknown phase5 acceptance level")
+	assert.Contains(t, err.Error(), "quick, full")
+	assert.Contains(t, err.Error(), "dogfood --live --write-acceptance")
+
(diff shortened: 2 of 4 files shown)


tmchow (approved):
This looks good to me.

I reviewed this against #644 and the existing Phase 5 gate contract. The change implements the preferred documented fallback (`level: "manual"`) while keeping the important safety checks: marker identity is still required, manual markers need positive counts, failed tests still reject, and manual uses the full-style `tests_passed == matrix_size` rule rather than quick's thresholded rule. The improved unknown-level error also covers the alternate acceptance criterion from the issue.

The SKILL.md update is the right companion change: it keeps runner-generated quick/full markers as the normal path, documents `manual` as an operator fallback only, and points users back to `dogfood --live --write-acceptance` when possible. The new tests cover direct gate validation, failure rejection, the error message, and the `lock promote` path.

CI currently reports all checks passing. I attempted a focused local `go test` in an isolated worktree, but the local test process hung in this environment; I do not see a code issue from review.

tmchow (changes requested):
I need to revise my earlier approval after re-checking this against the project contract rather than treating #644's preferred option as sufficient justification. I don't think this should merge as proposed.

Required changes:

1. Do not add `manual` as an accepted Phase 5 pass level in `phase5AcceptancePassed` without a stronger, machine-checkable proof contract. As implemented, a hand-authored marker with `level: "manual"`, `matrix_size: 1`, `tests_passed: 1`, and `tests_failed: 0` is enough for `lock promote` to publish a runstate-backed CLI. That materially weakens the Phase 5 gate: the repo's skill contract says Phase 5 live dogfood is the behavioral gate and normal `phase5-acceptance.json` markers come from `printing-press dogfood --live --write-acceptance`, not from an operator-authored escape hatch.

2. The salvageable part is the error-message improvement. Please keep the unknown-level diagnostic that enumerates accepted values and points users to `printing-press dogfood --live --write-acceptance`, but keep the accepted pass levels closed to runner-emitted `quick`/`full` unless this PR also introduces a real manual-evidence schema that the gate can validate. The narrower fix satisfies the issue's alternate acceptance criterion without adding a bypass.

3. Update the tests and SKILL.md accordingly. The new promote test currently demonstrates the bypass (`matrix_size: 1` manual marker promotes successfully); that should be removed or inverted. The skill should continue to make the runner the source of truth for pass markers. If a future manual fallback is desired, it should require auditable evidence fields/files and should not be a bare level string plus counts.

The motivation is understandable, but the current implementation changes the safety boundary rather than just improving operator ergonomics.

dinakars777:
Addressed the revised review feedback in 5711761:

- removed `manual` as an accepted Phase 5 pass level
- kept accepted pass levels closed to `quick` and `full`
- preserved the improved unknown-level diagnostic with the dogfood marker hint
- inverted the promote coverage so a `manual` marker is rejected

Verified with `go test ./internal/pipeline ./internal/cli`.

tmchow:
@dinakars777 

I made one small cleanup after review: the accepted Phase 5 levels are now defined once as `quick` / `full` constants, and both the Phase 5 gate diagnostic and live dogfood `--level` validation use that shared list.

The behavior is unchanged: `manual` and other unknown levels are still rejected, and the error still tells operators the accepted values and points them back to `printing-press dogfood --live --write-acceptance`. The cleanup just avoids having the same allow-list duplicated as separate string literals that could drift later.

tmchow (approved):

PR #681, opened 2026-05-07
feat(cli): emit AGENTS.md for printed CLIs

silverstein:
## Summary

This adds a root `AGENTS.md` to newly generated printed CLIs. After review feedback, the guide is intentionally tiny and invariant: it identifies the tree as generated, points systemic fixes back upstream, and routes agents to runtime discovery instead of duplicating generated docs.

The generated guide now focuses on:

- `doctor --json` for health/auth/connectivity truth
- `agent-context --pretty` for the live structured command/auth surface
- `which "<capability>" --json` and `<command> --help` for runtime discovery
- `--agent` for non-interactive machine-friendly invocation
- help/dry-run checks before unfamiliar mutating commands
- README/SKILL as the source for install, auth, examples, and longer product guidance

This keeps the per-CLI artifact small while preserving the local operating contract for shell agents working inside a copied or checked-out printed CLI directory.

## Validation

- `gofmt`
- `git diff --check`
- `go test ./internal/generator -run TestGeneratedAgentsGuide`
- `scripts/golden.sh verify`
- `go test ./...`

### DIFF
--- a/internal/generator/templates/agents.md.tmpl
+++ b/internal/generator/templates/agents.md.tmpl
@@ -1,0 +1,36 @@
+# {{.ProseName}} Printed CLI Agent Guide
+
+This directory is a generated `{{.Name}}-pp-cli` printed CLI. It was produced by CLI Printing Press, so treat systemic fixes as upstream Printing Press fixes first. Keep local edits narrow and document why a generated-tree patch belongs here.
+
+## Local Operating Contract
+
+Start by asking the generated CLI for current runtime truth:
+
+```bash
+{{.Name}}-pp-cli doctor --json
+{{.Name}}-pp-cli agent-context --pretty
+```
+
+Use runtime discovery instead of relying on a copied command list:
+
+```bash
+{{.Name}}-pp-cli which "<capability>" --json
+{{.Name}}-pp-cli <command> --help
+```
+
+Add `--agent` to command invocations for JSON, compact output, non-interactive defaults, no color, and confirmation-safe scripting:
+
+```bash
+{{.Name}}-pp-cli <command> --agent
+```
+
+Before running an unfamiliar command that may mutate remote state, inspect its help and prefer a dry run:
+
+```bash
+{{.Name}}-pp-cli <command> --help
+{{.Name}}-pp-cli <command> --dry-run --agent
+```
+
+Use `--yes --no-input` only after the target, arguments, and side effects are clear.
+
+For install, auth, examples, and longer product guidance, read `README.md` and `SKILL.md`. This file intentionally stays small so repo-local agents get invariant local guidance without duplicating the generated docs.
--- a/testdata/golden/expected/generate-golden-api-rich-auth/printing-press-rich-auth/AGENTS.md
+++ b/testdata/golden/expected/generate-golden-api-rich-auth/printing-press-rich-auth/AGENTS.md
@@ -1,0 +1,36 @@
+# Printing Press Rich Printed CLI Agent Guide
+
+This directory is a generated `printing-press-rich-pp-cli` printed CLI. It was produced by CLI Printing Press, so treat systemic fixes as upstream Printing Press fixes first. Keep local edits narrow and document why a generated-tree patch belongs here.
+
+## Local Operating Contract
+
+Start by asking the generated CLI for current runtime truth:
+
+```bash
+printing-press-rich-pp-cli doctor --json
+printing-press-rich-pp-cli agent-context --pretty
+```
+
+Use runtime discovery instead of relying on a copied command list:
+
+```bash
+printing-press-rich-pp-cli which "<capability>" --json
+printing-press-rich-pp-cli <command> --help
+```
+
+Add `--agent` to command invocations for JSON, compact output, non-interactive defaults, no color, and confirmation-safe scripting:
+
+```bash
+printing-press-rich-pp-cli <command> --agent
+```
+
+Before running an unfamiliar command that may mutate remote state, inspect its help and prefer a dry run:
+
+```bash
+printing-press-rich-pp-cli <command> --help
+printing-press-rich-pp-cli <command> --dry-run --agent
+```
+
+Use `--yes --no-input` only after the target, arguments, and side effects are clear.
+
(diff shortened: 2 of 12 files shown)


internal/generator/templates/agents.md.tmpl
```diff
@@ -0,0 +1,150 @@
...
+{{- if or $resource.Endpoints $resource.SubResources}}
+
+### {{$name}}
+
+{{$resource.Description}}
+{{range $eName, $endpoint := $resource.Endpoints}}
+{{- if index $.PromotedResourceNames $name}}
+- `{{$.Name}}-pp-cli {{$name}}{{positionalArgs $endpoint}} --agent` - {{oneline $endpoint.Description}}
+{{- else}}
+- `{{$.Name}}-pp-cli {{$name}} {{$eName}} --agent` - {{oneline $endpoint.Description}}
```
tmchow: This command reference drops required positional args for non-promoted endpoint commands, and the subresource loop below has the same problem. The new golden fixture demonstrates it: `projects tasks update-project --agent` is rendered in `AGENTS.md`, but the emitted Cobra command is `update-project <projectId> <taskId>`, so an agent following this guide will hit help/arg errors instead of executing the command. Please either remove the generated command catalog from `AGENTS.md` and point agents at `which` / `--help` / `agent-context`, or render these entries from the actual command surface including `positionalArgs` for every endpoint/subresource and add regression coverage for that.

tmchow (changes requested):
Thanks for putting this together. I think the underlying feature has merit: a tiny local operating contract in each generated CLI directory is useful, especially when a printed CLI is checked out or copied outside `mvanhorn/printing-press-library`.

I do want this artifact to be lighter before merge. The repo-level or `library/AGENTS.md` in `printing-press-library` can cover shared repository workflow, while the generated per-CLI file should stick to invariant local guidance: generated tree, prefer upstream Printing Press fixes for systemic changes, run `doctor --json`, use `which` / `--help` / `agent-context`, use `--agent`, and inspect help before mutating commands. The current version duplicates a command catalog and enough install/auth/flag prose that it is already prone to drift.

Requesting changes for the concrete correctness issue below: the generated command reference can emit commands an agent cannot actually run. Please either remove that catalog from `AGENTS.md` and route agents to runtime discovery, or generate it from the same actual command surface with required positionals and regression coverage for normal endpoints and subresources.

silverstein:
Thanks, this direction makes sense. I pushed `c374a759` to make the generated file much lighter:

- removed the generated command catalog entirely, so AGENTS.md no longer risks emitting non-runnable command paths
- removed duplicated install/auth/common-flag prose and point agents back to README/SKILL for that
- kept the invariant local contract: generated-tree warning, upstream-first systemic fixes, `doctor --json`, `agent-context --pretty`, `which`, `--help`, `--agent`, and help/dry-run before mutating commands
- refreshed the AGENTS.md goldens for both the main generated CLI fixture and the rich-auth fixture

Re-ran:

- `go test ./internal/generator -run TestGeneratedAgentsGuide`
- `scripts/golden.sh verify`
- `go test ./...`

testdata/golden/expected/generate-golden-api-rich-auth/printing-press-rich-auth/AGENTS.md
```diff
@@ -0,0 +1,36 @@
+# Printing Press Rich Printed CLI Agent Guide
+
+This directory is a generated `printing-press-rich-pp-cli` printed CLI. It was produced by CLI Printing Press, so treat systemic fixes as upstream Printing Press fixes first. Keep local edits narrow and document why a generated-tree patch belongs here.
```
tmchow, line 3: I would add a repo link on the CLI Printing Press, without it, this probably has no meaning to most agent contexts.

tmchow (changes requested):

internal/generator/templates/agents.md.tmpl
```diff
@@ -0,0 +1,36 @@
...
+
+Before running an unfamiliar command that may mutate remote state, inspect its help and prefer a dry run:
+
+```bash
+{{.Name}}-pp-cli <command> --help
+{{.Name}}-pp-cli <command> --dry-run --agent
+```
+
+Use `--yes --no-input` only after the target, arguments, and side effects are clear.
+
```
tmchow, line 35: add this verbatim section after this. it's related to another PR. I want to slipstream this 

```
## Local Customizations

If you modify this CLI beyond what the generator produced, record each customization so it isn't lost on the next regen and is visible to the next reader.

1. **Mark every changed site** in source with a comment summarizing the deviation:

    ```
    // PATCH: <one-line summary>
    ```

    Include an upstream reference inline when there is one (e.g. `// PATCH(upstream cli-printing-press#<issue>): …`). `grep -rn 'PATCH' .` from this directory then surfaces every customization.

2. **Catalog the change** in a `.printing-press-patches.json` at this CLI's root (parallel to `.printing-press.json`). Minimum shape:

    ```json
    {
      "schema_version": 1,
      "applied_at": "YYYY-MM-DD",
      "base_run_id": "<copy from .printing-press.json>",
      "base_printing_press_version": "<copy from .printing-press.json>",
      "patches": [
        {
          "id": "short-identifier",
          "summary": "What changed (one sentence).",
          "reason": "Why this customization was needed (one or two sentences).",
          "files": ["internal/cli/foo.go"],
          "validated_outcome": "Optional: non-obvious test result that confirms the fix.",
          "upstream_issue": "Optional: https://github.com/mvanhorn/cli-printing-press/issues/<n>"
        }
      ]
    }
    ```

This file is an **index of customizations**, not a second copy of the diff. Diffs live in `git`; code lives in the source files; the inline `// PATCH:` comment carries the local semantics. Keep `summary` and `reason` short — if you find yourself writing tables of field renames or code transformations, that detail belongs in the source comment or commit message, not here.
```

tmchow (changes requested):
one thing here

tmchow:
Closing in favor of #728, which fast-tracks this with the three pieces of feedback already applied:

- Adds the upstream repo link to `CLI Printing Press`.
- Adds the `## Local Customizations` section (`// PATCH:` comments + `.printing-press-patches.json` index).
- Picks up your latest revision that drops the in-template command catalog.

Original commit was cherry-picked, so authorship and `Co-Authored-By` are preserved on the merge. Thanks @silverstein!

silverstein:
Awesome, my pleasure, Trevin!

Issue #717, opened 2026-05-08
Scorecard token-presence checks penalize dead-code removal


heyimjames:
> Filed by `/printing-press-retro` from a run of **Google Search Console**
> (scorecard 71/100, 100% verify pass).

## Summary

Two scorecard dimensions (`sync_correctness` and `output_modes`) grade on lexical token presence — `strings.Contains(content, "paginatedGet")` and a similar check on `page_fetch`. When the polish skill removes genuinely dead helpers (rewarded by dogfood's `dead_code` dimension), these scorers regress. The CLI is contradicting itself: dogfood awards points for less dead code; scorecard takes them away because one specific helper name is no longer present. Polish itself flagged this as a retro candidate.

## Where to look

- **Component:** scorer
- **Likely area:** scorecard scoring rules in `internal/cli/` (search for `strings.Contains` against `"paginatedGet"` and `"page_fetch"` or equivalent literal tokens).
- **Triggered when:** a CLI's pagination or output-mode code uses different helper names than the literal tokens the scorer greps for, OR when polish removes a helper whose name happens to match the scorer's token.

## What we observed

From the polish skill's result block on the google-search-console run:

```
sync_correctness: 5 → 0
output_modes:    10 → 9
dead_code (dogfood): 3 → 5
go_vet issues: 1 → 0
```

Polish notes verbatim:

> **Scorer regression `sync_correctness` 5→0 and `output_modes` 10→9.** Caused by removing `paginatedGet`. Both scorers grade on token presence (`strings.Contains(content, "paginatedGet")` and `"page_fetch"`), not actual usage. Removing genuinely dead code (which dogfood rewards via `dead_code` 3→5) is punished by these dimensions. Net delta -3 score for one less dead function — scorer is contradicting itself. Retro candidate: scorer should grade on actual call sites, not lexical token presence.

The two helpers polish removed (`paginatedGet`, `extractResponseData`) were genuinely unused — verified by dogfood's static dead-code analysis. Polish also fixed a real SQL placeholder undercount in `sitemap-health` in the same pass; the net effect was a real quality improvement. The scorecard reported a regression anyway.

## Suspected root cause

Certain on the mechanism (token-presence check on a literal helper name); some uncertainty on the right replacement strategy. Two candidate directions:

- **(A) Multi-pattern grep:** small set of equivalent helper names per dimension. Cheap to implement, brittle in the long run as new patterns emerge.
[...]

PR #727, opened 2026-05-08
fix(cli): score sync_correctness on usage, not dead helper presence

heyimjames:
## Summary

`scoreSyncCorrectness` was awarding +2 whenever `paginatedGet` appeared in `internal/cli/*.go` — including when it was a dead helper with no callers. So when polish removed the dead helper (rightly, dogfood rewards `dead_code` removal), the scorer regressed by 2 even though pagination behavior didn't change. The pre-polish 5 was inflated; post-polish 0 was the correct value, just delivered as an apparent regression.

This is the dead-code-rewards-points half of #717.

## Change

Adds `funcIsCalled(content, name string) bool` in `internal/pipeline/scorecard.go`:

```go
func funcIsCalled(content, name string) bool {
    total := strings.Count(content, name)
    if total == 0 {
        return false
    }
    defs := strings.Count(content, "func "+name+"(")
    return total > defs
}
```

Applies it to the `paginatedGet` branch of the `scoreSyncCorrectness` OR. The other three tokens in the same OR (`hasNextPage`, `endCursor`, `cursor`) are GraphQL field names / variable names, not functions, so bare `strings.Contains` stays correct for those.

## Test plan

- [x] `TestScoreSyncCorrectness` fixtures use `paginatedGet(path, cursor)` as a call with no definition in scope. With the change: `total=1, defs=0 → true → still scores +2`. Existing assertion `assert.Equal(t, 10, scoreSyncCorrectness(dir))` should still pass.
- [x] `TestScoreSyncCorrectness_NonSyncFilename` same pattern, same outcome.
- [ ] Could not run `go test ./internal/pipeline/...` locally (no Go on the dev box). Relying on CI.
- [ ] Could not run `scripts/golden.sh verify`. Goldens *should* be unaffected because their fixtures presumably exercise pagination via real call sites, but a reviewer with Go available should confirm.

## What this PR does **not** fix

`scoreOutputModes` (line 311) checks the string literal `"page_fetch"`. That's not a function name, so `funcIsCalled` doesn't apply. Tightening that to "appears inside an `Fprintln`/`Fprintf` call site, not just any string occurrence" is regex/AST work and merits a separate, more carefully-tested PR.

Refs #717 (closes the sync_correctness half; output_modes half remains open).

### DIFF
--- a/internal/pipeline/scorecard.go
+++ b/internal/pipeline/scorecard.go
@@ -2161,7 +2161,7 @@ func scoreSyncCorrectness(dir string) int {
 	if strings.Contains(content, "SaveSyncState") {
 		score++
 	}
-	if strings.Contains(content, "paginatedGet") || strings.Contains(content, "hasNextPage") || strings.Contains(content, "endCursor") || strings.Contains(content, "cursor") {
+	if funcIsCalled(content, "paginatedGet") || strings.Contains(content, "hasNextPage") || strings.Contains(content, "endCursor") || strings.Contains(content, "cursor") {
 		score += 2
 	}
 	// URL path parameters only count when other sync signals are present,
@@ -2543,6 +2543,22 @@ func readFileContent(path string) string {
 func fileExists(path string) bool {
 	_, err := os.Stat(path)
 	return err == nil
+}
+
+// funcIsCalled reports whether content references a Go function name in a
+// way that's not just its own definition. The bare-token check (used by
+// some scorers as a behavior signal) rewards a CLI that *defines* a helper
+// regardless of whether anything calls it — so polish removing a genuinely
+// dead helper would regress the score even though the CLI's pagination
+// behavior didn't change. Counting non-definition occurrences ties the
+// signal to actual use. See cli-printing-press#717.
+func funcIsCalled(content, name string) bool {
+	total := strings.Count(content, name)
+	if total == 0 {
+		return false
+	}
+	defs := strings.Count(content, "func "+name+"(")
+	return total > defs
 }
 
 func computeGrade(percentage int) string {


tmchow (changes requested):
Thanks for tackling this — the diagnosis in #717 is right and starting with the `paginatedGet` half is good scope. But I think the predicate as written doesn't actually fix the scenario the issue describes, and the existing tests don't catch it.

## The bug

`funcIsCalled` counts *every* occurrence of `name` against only `func name(`:

```go
total := strings.Count(content, name)
defs  := strings.Count(content, "func "+name+"(")
return total > defs
```

Real generated CLIs ship a godoc on `paginatedGet` in `helpers.go`:

```go
// paginatedGet fetches pages and concatenates array results. The headers
// argument carries per-endpoint required headers ...
func paginatedGet(c interface { ... }) (json.RawMessage, error) { ... }
```

Verified against `~/printing-press/library/cal-com/internal/cli/helpers.go` (line 258 godoc, 262 def). Running the predicate against this content:

- **Dead case** (godoc + def, zero callers): total=2, defs=1 → `true` → still scores +2.
- **Alive case** (godoc + def + 1 call): total=3, defs=1 → `true` → scores +2.

So in the exact pre-polish state from the google-search-console polish proof — `paginatedGet` defined with its godoc, no callers — `funcIsCalled` still returns `true` because the comment counts as a "use." Polish removes the helper post-merge → 0 occurrences → `false`. Net regression after polish: still 2 points. The contradictory-scoring loop in #717 isn't closed.

## Why the existing tests don't catch this

Both fixtures (`scorecard_tier2_test.go:284`, `:379`) have a *call* to `paginatedGet` with no definition in scope: total=1, defs=0 → `true`. That's an unrealistic shape — production CLIs always emit the def in `helpers.go`. So CI passing here doesn't tell us the fix works on real input.

## Suggested changes

1. Count call expressions, not bare name occurrences:

   ```go
   calls := strings.Count(content, name+"(")
   defs  := strings.Count(content, "func "+name+"(")
   return calls > defs
   ```

   `paginatedGet(` appears in both the def signature and any call site; subtracting defs leaves true call sites. Comments like `// paginatedGet fetches pages` no longer count. Still lexical, still cheap.

2. Add a regression test that mirrors real `helpers.go` shape: godoc comment + `func paginatedGet(...)` body + **no** call site. Assert the `paginatedGet` branch contributes 0. Without this, future changes to `funcIsCalled` can silently regress again.

Scope-limiting to the `paginatedGet` half (deferring `page_fetch` / `scoreOutputModes`) is fine and you flagged it clearly. Once the predicate actually distinguishes dead-with-godoc from called, this is mergeable.

tmchow:
However, this PR has been superseded by #732 ("fix(cli): score structural scorer behavior"), which was merged after this was opened. That PR replaced the entire `strings.Contains(content, "paginatedGet") || ...` check with `hasSyncPaginationStructure(cliDir)`, an AST-based function that parses the actual Go syntax tree and looks for a loop containing a real fetch call, a cursor/pagination signal, and an exit condition. Structural AST analysis is immune to the dead-code-rewards-points problem by construction — a helper that's defined but never called inside a loop body won't match.

As a result, this PR now has a merge conflict (the target line no longer exists), and the `funcIsCalled` helper it introduces would have no callers in the current codebase.

Closing as superseded by #732.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #735, opened 2026-05-08
fix(cli): populate manifest.Category from spec when catalog misses

jheitzeb:
## Summary

Synthetic CLIs (not in the embedded catalog) silently lost their `spec.Category` at manifest-write time, breaking `verify-skill`'s canonical-sections check downstream.

The README/SKILL templates correctly read `.Category` from the pipeline state and render `library/<category>/...` URLs. But the manifest writer in `internal/pipeline/climanifest.go` only sources `m.Category` from the embedded catalog. For a CLI with `category: travel` in `spec.yaml` and no catalog entry, the lookup fails silently and `m.Category` stays empty.

`internal/cli/verify_skill.go:163` then reads `manifest.Category` (empty), passes it to `generator.CanonicalSkillInstallSection`, which falls back to `"other"` — and the verifier reports an install-section drift between the SKILL (rendered with the real category) and the expected canonical.

Each component was doing the right thing in isolation; only the manifest writer dropped category on the floor.

## Fix

Two sites — both follow the same pattern: **catalog wins, spec is the fallback** for synthetic CLIs:

- `WriteManifestForGenerate` (`climanifest.go`): fall back to `p.Spec.Category` after the catalog lookup.
- `PublishWorkingCLI` (`publish.go`): same fallback, sourced from the spec already parsed for MCP metadata population later in the function.

## Tests

Three new tests in `climanifest_test.go`:

| Test | Pins |
|---|---|
| `TestWriteManifestForGeneratePopulatesCategoryFromSpec` | The fix: synthetic CLI with `spec.Category=travel` → `manifest.Category=travel` |
| `TestWriteManifestForGenerateCatalogCategoryWinsOverSpec` | Precedence: catalog wins when an API has both a catalog entry and a spec.Category |
| `TestWriteManifestForGenerateNoCategoryAnywhere` | Empty-case: manifest.Category stays empty when neither source provides one (so `install_section.go`'s "other" fallback still fires for genuinely-uncategorized CLIs) |

## Verification

- `go test ./...` — clean (26/26 packages pass)
- `scripts/golden.sh verify` — 13/13 PASS (no fixture drift; every existing golden either has a catalog entry or empty `spec.Category`)
- `go vet ./...` — clean
- `gofmt` — clean

## How this was found

This surfaced while I (an AI assistant — Claude Code) was running the `/printing-press` skill for a user on `wanderlust-goat`, a synthetic multi-source CLI with `category: travel` in its spec. `printing-press publish validate` reported a `verify-skill` canonical-sections failure that required hand-editing the SKILL's install URL from `library/other/` to `library/travel/` before publish would proceed (see [printing-press-library#291](https://github.com/mvanhorn/printing-press-library/pull/291)). Tracing the chain backward isolated this issue to the manifest writer.

## Note from the AI

I'm filing this on the user's behalf with their approval. The user trusts the diagnosis. I added test coverage, ran the full test suite + golden harness, and confirmed the fix doesn't shift any existing fixture output. The PR body above is my own write-up — please review the actual diff carefully rather than relying on the prose. Any review feedback or rejection is welcome; happy to revise or close if a different approach is preferred.

## Test plan

- [x] Unit tests pass (`go test ./internal/pipeline/...`)
- [x] Full repo tests pass (`go test ./...`)
- [x] Golden harness clean (`scripts/golden.sh verify` → 13/13)
- [x] `go vet ./...` clean
- [ ] CI green

### DIFF
--- a/internal/pipeline/climanifest.go
+++ b/internal/pipeline/climanifest.go
@@ -469,6 +469,17 @@ func WriteManifestForGenerate(p GenerateManifestParams) error {
 			m.DisplayName = entry.DisplayName
 		}
 	}
+	// Fall back to spec.Category for synthetic CLIs that aren't in the
+	// embedded catalog. Without this, manifest.Category stays empty even
+	// when the spec sets `category: travel`, and verify-skill's canonical-
+	// sections check then expects the install URL to use "other" — putting
+	// the rendered SKILL (which read category from the spec via the
+	// template's .Category) and the manifest-derived expected SKILL out of
+	// sync. The README/SKILL templates already resolve category through the
+	// spec; the manifest writer was the lone holdout.
+	if m.Category == "" && p.Spec != nil && p.Spec.Category != "" {
+		m.Category = p.Spec.Category
+	}
 
 	// Record the API version from the spec for provenance (not the CLI version).
 	if p.Spec != nil && p.Spec.Version != "" {
--- a/internal/pipeline/climanifest_test.go
+++ b/internal/pipeline/climanifest_test.go
@@ -1350,3 +1350,76 @@ func TestPopulateMCPMetadataCLIDescription(t *testing.T) {
 		assert.Equal(t, "Catalog description.", m.Description)
 	})
 }
+
+// TestWriteManifestForGeneratePopulatesCategoryFromSpec pins the fallback
+// that lets synthetic CLIs (not in the embedded catalog) carry their
+// spec.Category through to .printing-press.json. Without this fallback,
+// verify-skill's canonical-sections check expects the install URL to use
+// "other" while the rendered SKILL (which reads category from the spec
+// via the template's .Category) uses the real category — a structural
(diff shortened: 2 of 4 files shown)


tmchow (changes requested):
Thanks for the careful diagnosis here. I agree the underlying bug is real: generated SKILL.md renders the install path from the parsed spec category, while verify-skill derives the expected block from .printing-press.json, so synthetic/non-catalog CLIs can drift to library/other when the manifest drops spec.Category.

Please add coverage for the publish path as well as generate. The implementation changes writeCLIManifestForPublish in internal/pipeline/publish.go to rehydrate Category from the archived spec, but the new tests only exercise WriteManifestForGenerate. A small test in internal/pipeline/publish_test.go with a synthetic/non-catalog state plus archived spec.yaml containing category: travel would lock the second half of the fix and prevent publish-time manifest rewrites from regressing independently.

PR #756, opened 2026-05-08
fix(cli): sync sibling list endpoints

### LINKED ISSUE #701: Multi-endpoint per resource: novel-feature SQL needs data from sibling list endpoints
## Summary

Split out from #689 (Bug 2). Deferred for design before code.

The printing-press currently maps one resource family to a single sync endpoint (`resourcePathMap` in the generated `internal/cli/sync.go` is 1:1, e.g. `portfolio` → `/portfolio/fills`). When a novel-feature command's SQL depends on data from a *sibling* endpoint under the same prefix (Kalshi exposes `/portfolio/{fills,settlements,positions,orders}`), that data is unreachable through `sync` without manual additions to the printed CLI.

Concrete symptom on the Kalshi CLI: `portfolio winrate` filters on `WHERE type='settlement' OR settled_at IS NOT NULL` but the only synced portfolio data is fills, which lack those fields. Workaround in #689 added `portfolio-settlements` as a separate sync resource.

## Why this is deferred

The two candidate fix shapes both have design weight:

- **(a) Profiler detects sibling list endpoints under a shared path prefix and emits one sync resource per endpoint.** Could over-split or break APIs that intentionally use a single canonical list endpoint. Needs a clear heuristic for what counts as a "sibling list" (path-prefix match alone is too loose; response-shape similarity may be needed).
- **(b) Novel-feature codegen carries an explicit \"this command's SQL targets resource X\" link, and dogfood asserts column overlap at gen-time.** This is closer to the meta-fix proposed in #689 Bug 4 (SQL ↔ schema validation in dogfood) and probably the better landing — but it depends on Bug 4 being in place first.

## Recommended sequencing

Land #689 Bug 4 (dogfood SQL ↔ schema validator) first. Once novel-feature SQL is mechanically tied to a target resource and validated against synced data, revisit this issue with a concrete fix shape — likely (b).

## References

- Parent issue with full context: #689
- Bug 4 (meta-fix this depends on): see #689

tmchow:
## Summary

- keep the canonical sync resource on the shortest standalone list endpoint while preserving sibling list endpoints as additional sync resources
- derive stable sibling names from the endpoint path, e.g. `portfolio-settlements`, without splitting generic aliases like `/things/all`
- add profiler and generator regression coverage for sibling portfolio list endpoints

Closes #701

## Verification

- `go test ./internal/profiler`
- `scripts/golden.sh verify`
- `go fmt ./...`
- `go test ./...`
- `golangci-lint run ./...`

### DIFF
--- a/internal/generator/generator_test.go
+++ b/internal/generator/generator_test.go
@@ -6158,6 +6158,59 @@ func TestGenerateSyncRejectsUnknownResourcePath(t *testing.T) {
 		"sync must not request fake /<resource> paths")
 }
 
+func TestGenerateSyncIncludesSiblingListResources(t *testing.T) {
+	t.Parallel()
+
+	apiSpec := &spec.APISpec{
+		Name:    "trading",
+		Version: "0.1.0",
+		BaseURL: "https://api.example.com",
+		Auth:    spec.AuthConfig{Type: "none"},
+		Config: spec.ConfigSpec{
+			Format: "toml",
+			Path:   "~/.config/trading-pp-cli/config.toml",
+		},
+		Resources: map[string]spec.Resource{
+			"portfolio": {
+				Description: "Portfolio",
+				Endpoints: map[string]spec.Endpoint{
+					"fills": {
+						Method:      "GET",
+						Path:        "/portfolio/fills",
+						Description: "List fills",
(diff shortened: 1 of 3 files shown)


tmchow:
@mergifyio queue

PR #784, opened 2026-05-09
feat(cli): add Domino's catalog entry under food-and-dining

joeynyc:
## Intent

Establish a catalog entry for Domino's so the Printing Press can produce a CLI for it. First entry in the `food-and-dining` category from a planned food/fast-food set (Papa John's, Pizza Hut, McDonald's, Burger King, Wendy's to follow as separate PRs).

Issue: N/A

## Approach

Follows the `kayak.yaml` precedent: Domino's has no official API, so the entry is `tier: community` with a synthetic `html-scrape` wrapper pointing at `order.dominos.com/power/`. The `power/*` JSON endpoints (store-locator, store-profile, menu, validate-order, price-order, track) are the same surface the web/mobile apps consume and have been stable for years. The long-running JS wrapper `RIAEvangelist/node-dominos-pizza-api` is referenced for endpoint shapes only — not a runtime dependency. The printed Go CLI should reimplement the HTTP calls natively.

## Scope

Primary area: `catalog/`

Why this belongs in this repo: Catalog YAML is generator input; adding it here lets `/printing-press dominos` and `printing-press catalog` discover the entry. No printed-CLI-only changes here.

## Risk

N/A — single new YAML file. Does not modify generator templates, embedded packages, MCP surface, or scorer.

## Output Contract

N/A — adds a new catalog entry; does not change rendering of existing entries or any generated artifact.

## Verification

- `go test ./internal/catalog/...` — passes (validates `TestParseFSEmbeddedCatalog` and the schema rules: category in allowed set, integration_mode in {native, subprocess, html-scrape}).
- `printing-press catalog list | grep dominos` — entry appears with correct description.

## AI / Automation Disclosure

### DIFF
--- a/catalog/dominos.yaml
+++ b/catalog/dominos.yaml
@@ -16,12 +16,6 @@ notes: |
   JSON and are the same surface the web and mobile apps consume. They've
   been stable enough that community wrappers have tracked them for years.
 
-  The spec at catalog/specs/dominos-spec.yaml is the Printing Press internal
-  YAML format (spec_format: custom), not OpenAPI. It mirrors the spec already
-  used by the published dominos-pp-cli in the printing-press-library repo so
-  this catalog entry stays continuous with prior generation history rather
-  than introducing a divergent shape.
-
   Surface scope:
     - Browse / read: store-locator, store/{id}/profile, store/{id}/menu,
       tracker, deals, menu categories, cart pricing — anonymous, safe by
--- a/catalog/dominos.yaml
+++ b/catalog/dominos.yaml
@@ -1,36 +1,36 @@
 name: dominos
 display_name: Domino's
-description: Pizza ordering, store lookup, and order tracking. No public API — reached through reverse-engineered JSON endpoints powering the order.dominos.com web flow.
+description: Pizza ordering, store lookup, menu browsing, deal lookup, and order tracking. No public API — reached through reverse-engineered JSON endpoints under order.dominos.com/power.
 category: food-and-dining
+spec_url: https://raw.githubusercontent.com/mvanhorn/cli-printing-press/main/catalog/specs/dominos-spec.yaml
+spec_format: custom
+spec_source: sniffed
+auth_required: false
 tier: community
 verified_date: "2026-05-09"
 homepage: https://www.dominos.com
 notes: |
   Domino's does not publish an API. The internal endpoints under
-  `https://order.dominos.com/power/` (store-locator, store-profile, menu,
-  validate-order, price-order, place-order, track) return JSON and are the same
-  surface the web and mobile apps consume. They've been stable enough that
-  community wrappers have tracked them for years.
+  https://order.dominos.com/power/ (store-locator, store/{id}/profile,
+  store/{id}/menu, validate-order, price-order, place-order, tracker) return
+  JSON and are the same surface the web and mobile apps consume. They've
+  been stable enough that community wrappers have tracked them for years.
 
-  Treat this as html/json-sniff implementation backing. The generator should
-  emit a thin HTTP client targeting the `power` endpoints rather than searching
-  for an OpenAPI spec. Auth is anonymous for browse/track flows; placing real
-  orders requires a Domino's account and is out of scope for the printed CLI's
-  default surface.
+  The spec at catalog/specs/dominos-spec.yaml is the Printing Press internal
+  YAML format (spec_format: custom), not OpenAPI. It mirrors the spec already
+  used by the published dominos-pp-cli in the printing-press-library repo so
+  this catalog entry stays continuous with prior generation history rather
+  than introducing a divergent shape.
 
-  Suggested compound queries the printed CLI should make easy:
-    - "nearest open delivery store to <address> with current wait time"
-    - "full menu + active coupons for store <id>, cheapest large pepperoni"
-    - "track order <phone or order key> until delivered"
-wrapper_libraries:
-  - name: power-endpoint-client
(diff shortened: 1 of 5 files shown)


tmchow (commented):
Fix failing CI

tmchow:
Thanks for the PR. I think Domino's has merit for the catalog, and the live `power` surface is still real: I spot-checked `https://order.dominos.com/power/store-locator` with a delivery address today and it returned useful JSON store data. There is also prior Printing Press history for Domino's, including an existing published-library spec, so this is not just an aspirational food entry.

I would not merge this as-is, though. Follow-up I would want before merge:

- Fix the PR title to the required conventional format, e.g. `feat(cli): add Domino's catalog entry`. Figure out why your agent is not following the AGENTS.md in this repo as that convention is clearly stated.
- Update the `catalog-list` golden. CI is red because adding the first `food-and-dining` catalog row intentionally changes the golden output.
- Do not leave this as only a wrapper-only entry if the goal is that `printing-press generate dominos` works from the catalog. Wrapper-only entries currently have no `spec_url`, and the discovery path rejects them with “try providing a URL with --spec”. Since Domino's already has a generated/library spec shape, the stronger follow-up is to add `catalog/specs/dominos-spec.yaml` (or otherwise point at a real reproducible spec) and have `catalog/dominos.yaml` reference it.
- Keep mutating/externally visible ordering actions scoped carefully. Store lookup, store profile, menu, pricing/validation, and tracking are good candidates. Actual `place-order` behavior should require explicit user confirmation/opt-in in the printed CLI and should not be implied as a default anonymous browse surface.

So: worth including, but it should be a reproducible spec-backed catalog entry rather than another wrapper-only scrape note.

joeynyc:
Thanks for the detailed review @tmchow. Pushed two commits addressing all four asks:

**1. PR title** — fixed to `feat(cli): add Domino's catalog entry under food-and-dining`. Sorry for missing the convention; the AGENTS.md commit-style rule applies to PR titles too and I should have caught it before opening.

**2. Catalog-list golden** — regenerated via `scripts/golden.sh update` in commit `c638608d`. Diff is exactly one new `food-and-dining:` section with the dominos line, no other case changed.

**3. Spec-backed instead of wrapper-only** — commit `a6612e84` rewrites `catalog/dominos.yaml` to mirror the `postman-explore.yaml` shape:
- `spec_url` points at `catalog/specs/dominos-spec.yaml` (raw GitHub URL)
- `spec_format: custom` (the spec is the Printing Press internal YAML format, not OpenAPI)
- `spec_source: sniffed`, `auth_required: false`
- The new `catalog/specs/dominos-spec.yaml` is copied directly from the published `dominos-pp-cli` in `mvanhorn/printing-press-library` (`library/food-and-dining/dominos/spec.yaml`) — keeps continuity with prior generation history rather than introducing a divergent shape.

**4. place-order scoping** — the spec retains `place-order` so the published CLI's existing surface stays reproducible from this catalog entry, but the entry's `notes` block now explicitly flags it as a side-effect command that the printed CLI MUST gate behind an explicit `--confirm`/`--place` flag and short-circuit on `PRINTING_PRESS_VERIFY=1`, per the AGENTS.md "Side-effect commands" rule. The default browse surface (store-locator, store profile, menu, validate-order, price-order, tracker, deals) stays anonymous and read-only.

Verification I ran locally on the branch:
- `go test ./internal/catalog/...` ✅
- `go test ./...` ✅ (no failures)
- `scripts/golden.sh verify` ✅ after the update commit
- `printing-press catalog list` shows the new `food-and-dining:` section with `dominos` correctly grouped

Let me know if you'd like the spec slimmed (e.g., dropping place-order from the spec entirely rather than just gating it at the CLI layer) — happy to push another commit.

joeynyc:
Update — pushed `7d85b72c` fixing a generator bug surfaced by this PR's `validate` job.

**The bug:** `helpers.go.tmpl` had the `truncateJSONArray` function block nested inside the `{{- if .HasDataLayer}}` guard, so the helper was only emitted when both `HasClientLimit` AND `HasDataLayer` were true. The corresponding call site (in `command_endpoint.go.tmpl` and `command_promoted.go.tmpl`) is gated only on `endpointNeedsClientLimit`. The dominos spec has `customer.orders` (GET, non-positional `limit` param, no Pagination block) but profiles to no data layer, so it emitted a call against an undefined function — `go build` fails with `undefined: truncateJSONArray`.

**The fix:** Moved the `{{- if .HasClientLimit}}` block out of `HasDataLayer` so it's a peer rather than nested. Added `TestClientLimitGenerationWithoutDataLayer` as a regression test — the existing `TestClientLimitGenerationEmitsHelperAndBuilds` didn't catch this because its spec shape happens to trip the profiler heuristic that auto-enables Store. The new test pre-sets `g.VisionSet` so `HasDataLayer` stays false at template render time.

**Verification:**
- `go test ./internal/generator/...` ✅ (including the new regression test, which fails without the template fix)
- `scripts/golden.sh verify` ✅ (all 16 cases still pass — no existing CLI changes since they all had `HasDataLayer=true`)
- Local `printing-press generate --spec catalog/specs/dominos-spec.yaml --output /tmp/dominos-cli` now passes all 8 quality gates: `go mod tidy`, `govulncheck`, `go vet`, `go build`, binary build, `--help`, `version`, `doctor`

Open question for you: this is bundling a machine fix with a catalog PR. Per AGENTS.md "Default to machine changes", happy to split the generator fix into its own PR and rebase this one on top — just say the word.

tmchow:
Stale with no progress, closing.

PR #785, opened 2026-05-09
feat(cli): add McDonald's catalog entry

joeynyc:
## Summary
- Adds `catalog/mcdonalds.yaml` as a wrapper-only entry under `food-and-dining`, community tier
- McDonald's publishes no API; generator targets the public locator endpoints and the GMA mobile-app JSON surface, mirroring the Domino's entry's shape
- Browse flows (locator, menu, deals) are in scope; ordering / Rewards redemption requires real auth and is out of scope for the default surface

## Test plan
- [ ] `go build -o ./printing-press ./cmd/printing-press` succeeds (catalog is a Go embed, requires rebuild)
- [ ] `go test ./internal/catalog/...` passes (validates required fields, category enum, tier, wrapper-only shape)
- [ ] Spot-check `./printing-press catalog list` shows the McDonald's entry

### DIFF
--- a/catalog/mcdonalds.yaml
+++ b/catalog/mcdonalds.yaml
@@ -1,0 +1,40 @@
+name: mcdonalds
+display_name: McDonald's
+description: Restaurant lookup, menu browsing, and Deals/MyMcDonald's Rewards offers. No public API — reached through reverse-engineered JSON endpoints powering the McDonald's mobile app and mcdonalds.com.
+category: food-and-dining
+tier: community
+verified_date: "2026-05-09"
+homepage: https://www.mcdonalds.com
+notes: |
+  McDonald's does not publish an API. The internal endpoints under
+  `https://www.mcdonalds.com/googleapps/` and the GMA (Global Mobile App)
+  hosts (`https://gmqsrmobile.mcd.com/`, `https://api.mcdonalds.com/`) return
+  JSON and are the same surface the web restaurant locator and the iOS/Android
+  apps consume. The mobile-app endpoints require an OAuth2 client credential
+  pair scoped to the GMA tenant; the public web locator endpoints are
+  anonymous.
+
+  Treat this as html/json-sniff implementation backing. The generator should
+  emit a thin HTTP client targeting the locator and menu endpoints rather than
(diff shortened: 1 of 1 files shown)


tmchow:
Thanks for putting this together. I think this has merit as a community, scrape-backed catalog entry, but I do not think it is ready to merge as written.

The live shape checks out in the corrected direction: the legacy `/googleapps/GoogleRestaurantLocAction.do` path is gone, `api.mcd.com` is not anonymously usable, and the public `mcdonalds.com` HTML/sitemap surface is reachable. The strongest supported surface is restaurant sitemap + store detail pages + full menu/product nutrition pages.

Follow-up I would want before merge:

- Update the `catalog-list` golden. CI is red because adding the entry intentionally changes catalog output.
- Narrow the wording so this does not imply `printing-press generate mcdonalds` is generator-ready from the catalog alone. Wrapper-only entries currently have no `spec_url`; the discovery path rejects them and asks for `--spec`. Either add an actual internal spec/source path, or frame this as catalog/discovery backing for a future scrape spec.
- Keep the supported surface to what is proven from public HTML. Locator/store details, menu pages, product nutrition/allergens are solid. Public deals should be described carefully unless the PR proves extractable offer objects rather than just the marketing/app page. Ordering, Rewards, personalized offers, and mobile-app API key/JWT flows should stay explicitly out of scope.

So: worth including, but as an honest public-HTML extraction candidate, not as a ready API/spec source yet.

catalog/mcdonalds.yaml
```diff
@@ -0,0 +1,59 @@
+name: mcdonalds
```
tmchow, line 1: This new catalog entry changes `printing-press catalog list`, so the golden fixture has to move in the same PR. CI is already red on `catalog-list stdout.txt` for exactly that missing update, which makes this unmergeable as-is.

tmchow (changes requested):
The entry itself is still worth carrying, but one blocking fix is needed before this is mergeable.

tmchow:
Stale with no progress, closing.

tmchow:
Stale with no progress, closing.

PR #799, opened 2026-05-09
fix: doctor reports 'not configured' for query-param auth APIs

### LINKED ISSUE #339: youtube: doctor reports auth 'not configured' when API key is set via env var
## Bug

`youtube-pp-cli doctor` reports `FAIL Auth: not configured` even when `YOUTUBE_DATA_OAUTH2C` is set in the environment and working.

## Root Cause

In `doctor.go`, the auth check relies on `cfg.AuthHeader()`:

```go
header := cfg.AuthHeader()
if header == "" {
    report["auth"] = "not configured"
}
```

But `AuthHeader()` correctly returns `""` for YouTube because the Data API uses `?key=` query params, not `Authorization: Bearer` headers. The client adds the key via `q.Set("key", c.Config.YoutubeDataOauth2c)` in `client.go:222-224`.

The doctor doesn't account for the API-key-as-query-param auth pattern.

## Fix

```go
if header == "" && cfg.YoutubeDataOauth2c == "" {
    report["auth"] = "not configured"
    report["auth_hint"] = "export YOUTUBE_DATA_OAUTH2C=<your-key>"
} else if header == "" && cfg.YoutubeDataOauth2c != "" {
    // YouTube Data API uses ?key= query param, not Authorization header
    report["auth"] = "configured (API key)"
    report["auth_source"] = cfg.AuthSource
} else {
    report["auth"] = "configured"
    report["auth_source"] = cfg.AuthSource
}
```

## Broader Implication

This may affect other Printing Press-generated CLIs where the API uses query-param auth instead of header-based auth. The `doctor` codegen template could check for query-param auth patterns generically.

## Reproduction

```bash
export YOUTUBE_DATA_OAUTH2C=AIzaSy...
youtube-pp-cli doctor
# FAIL Auth: not configured     <-- wrong
# OK Env Vars: OK 1/1 available <-- contradicts Auth line
```

## After Fix

```bash
youtube-pp-cli doctor
# OK Auth: configured (API key)
# OK Env Vars: OK 1/1 available
```

shekerkamma:
## Summary

- When `Auth.In == "query"`, `AuthHeader()` deliberately returns `""` because the credential goes as a `?key=` query parameter, not an `Authorization` header
- The `doctor.go.tmpl` template only checked `AuthHeader()`, causing it to report `FAIL Auth: not configured` even when the API key env var is set and working
- Fix: when `Auth.In == "query"`, check `cfg.AuthSource` (populated during `config.Load()` when the env var is present) instead of `AuthHeader()`
- Reports `"configured (API key)"` for query-param auth with valid credentials

## Affected APIs

YouTube Data API v3 and any future CLI where the OpenAPI spec declares `apiKey` auth with `in: query`.

## Reproduction

```bash
export YOUTUBE_DATA_OAUTH2C=AIzaSy...
youtube-pp-cli doctor
# Before: FAIL Auth: not configured
#         OK Env Vars: OK 1/1 available   <-- contradicts Auth line
# After:  OK Auth: configured (API key)
#         OK Env Vars: OK 1/1 available
```

## Golden test impact

None — all three golden fixtures (`golden-api`, `rich-auth`, `tier-routing`) use `in: header` for top-level auth. The new `{{if eq .Auth.In "query"}}` branch only triggers for query-param auth specs.

Closes mvanhorn/printing-press-library#339

## Test plan

- [ ] Verify golden tests still pass (`go test ./internal/generator/...`)
- [ ] Generate a CLI from a spec with `in: query` auth and verify doctor reports correctly
- [ ] Verify existing `in: header` CLIs are unaffected

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/internal/generator/templates/doctor.go.tmpl
+++ b/internal/generator/templates/doctor.go.tmpl
@@ -190,6 +190,29 @@ func newDoctorCmd(flags *rootFlags) *cobra.Command {
 {{- else}}
 			if cfg != nil {
 				header := cfg.AuthHeader()
+{{- if and .Auth .Auth.In (eq .Auth.In "query")}}
+				// Auth is sent as a query parameter, not a header — AuthHeader()
+				// returns "" by design. Check the credential field directly.
+				if header == "" && cfg.AuthSource != "" {
+					report["auth"] = "configured (API key)"
+					report["auth_source"] = cfg.AuthSource
+				} else if header == "" {
+{{- if .Auth.Optional}}
+					report["auth"] = "optional — not configured"
+{{- else}}
+					report["auth"] = "not configured"
+{{- end}}
+{{- with $canonicalEnvVar}}
+					report["auth_hint"] = "export {{.Name}}=<your-key>"
+{{- end}}
+{{- if .Auth.KeyURL}}
+					report["auth_key_url"] = "{{.Auth.KeyURL}}"
+{{- end}}
+				} else {
+					report["auth"] = "configured"
+					report["auth_source"] = cfg.AuthSource
+				}
+{{- else}}
 				if header == "" {
 {{- if .Auth.Optional}}
 					report["auth"] = "optional — not configured"
@@ -206,6 +229,7 @@ func newDoctorCmd(flags *rootFlags) *cobra.Command {
 					report["auth"] = "configured"
 					report["auth_source"] = cfg.AuthSource
 				}
+{{- end}}
 			}
 {{- end}}
 


internal/generator/templates/doctor.go.tmpl
```diff
@@ -190,6 +190,29 @@ func newDoctorCmd(flags *rootFlags) *cobra.Command {
 {{- else}}
 			if cfg != nil {
 				header := cfg.AuthHeader()
+{{- if and .Auth .Auth.In (eq .Auth.In "query")}}
```
tmchow, line 193: This adds a new doctor template branch, but there is no regression test covering top-level query-param auth. The repo already pins doctor emission in `internal/generator/auth_status_test.go` and `internal/generator/auth_optional_test.go`; without a focused test, this path can regress silently because the existing golden fixtures do not exercise `Auth.In == "query"` here.

tmchow (changes requested):
The bug is real and the template direction looks right, but one blocking gap remains.

tmchow:
no activity, closing

PR #854, opened 2026-05-09
fix(skills): correct update preflight compatibility

tmchow:
## Summary

Printing Press preflight now separates compatibility blocks from update advice. Repo checkouts still check `origin/main` every run, but skipping an update records that exact SHA so the same revision is not prompted again. Standalone installs compare against `@latest`, and binaries below the declared floor stop before any optional freshness prompt can offer a misleading skip path.

## Details

- Sets the user-invocable Printing Press skill floor to `min-binary-version: "4.0.0"`, including polish.
- Makes `min-binary-version` the hard compatibility gate and keeps minor or patch freshness updates optional.
- Adds a release test that keeps internal skill binary floors aligned with the current binary major and their setup-contract comments.

## Test Plan

- `go test ./internal/cli -run 'Test(VersionConsistencyAcrossFiles|InternalSkillMinimumBinaryVersionsTrackMajor|ModulePathMatchesMajorVersion|ReleasePleaseAnnotationExists)'`
- Embedded `/printing-press` setup shell block passes `bash -n`.
- Standalone simulation reports `installed=4.0.0 latest=4.2.1` and detects an update.
- `git diff --check`

---

GPT-5

### DIFF
--- a/internal/cli/release_test.go
+++ b/internal/cli/release_test.go
@@ -78,6 +78,38 @@ func TestVersionConsistencyAcrossFiles(t *testing.T) {
 		"plugin.json and version.go hardcoded version must match")
 }
 
+func TestInternalSkillMinimumBinaryVersionsTrackMajor(t *testing.T) {
+	// Skill frontmatter `version` values are not release-managed. The
+	// executable compatibility contract is `min-binary-version`; keep the
+	// frontmatter and the duplicated setup-contract comment in sync.
+	want := fmt.Sprintf("%d.0.0", majorVersion(t, version.Version))
+	paths := []string{
+		"../../skills/printing-press/SKILL.md",
+		"../../skills/printing-press-catalog/SKILL.md",
+		"../../skills/printing-press-polish/SKILL.md",
+		"../../skills/printing-press-publish/SKILL.md",
+		"../../skills/printing-press-score/SKILL.md",
+	}
+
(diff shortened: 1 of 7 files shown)


tmchow:
@greptileai review this

tmchow:
> @greptile-apps flagged in [the review summary](https://github.com/mvanhorn/cli-printing-press/pull/854#issuecomment-3543500481): "Continue setup" conflicts with "reload agent session" — `skills/printing-press/references/setup-checks.md:89-103`
>
> An AI agent following instructions linearly could interpret the reload directive as a stop signal for the current run, abandoning the rest of the setup-checks sequence (min-binary-version check, etc.). [...] Making this explicit — e.g., "Tell the user the above out-of-band steps for their skills, then continue setup with the newly-installed binary" — removes the ambiguity.

Addressed in f3634f3c. Re-ordered Section 4's post-install wording so the "continue this run with the new binary" directive lands immediately after the `printing-press version --json` confirmation (with explicit "do not stop, do not reload the session, do not skip the remaining checks") *before* the user-facing `gh skill update` / `npx skills update` block. Those commands are now framed up-front as out-of-band advice for the user's *next* session, with the reload-after-this-run language clearly scoped to "for next time" so an agent reading linearly can't misread it as a stop signal mid-run.

PR #861, opened 2026-05-09
fix(generator): preserve multi-spec server prefixes

### LINKED ISSUE #831: generator: multi-spec generation drops base-URL path prefix when input specs declare different servers[].url paths
Surfaced by `/printing-press-retro` after generating a multi-spec OpenAPI CLI from two specs whose `servers[]` declared different base-URL prefixes. Half the generated CLI's commands hit 404 because their endpoints' base prefix was lost at merge time.

## Where to look

- **Component:** generator (primary). May touch spec-parser if base-URL detection happens there.
- **Likely area:** the spec-merge logic when `--spec foo.json --spec bar.json` is passed. Single-spec generation is unaffected.

## What we observed

Two OpenAPI specs:

- Spec A: `servers: [{url: https://{tenant}}]` + endpoint `/projects` (typical "API at root" shape)
- Spec B: `servers: [{url: https://{tenant}/wiki/api/v2}]` + endpoint `/pages` (different product on same tenant, different path prefix)

The user passes both: `printing-press generate --spec a.json --spec b.json --name combo`.

The generator collapses to ONE `BaseURL` field on the Config struct (`https://{tenant}`). Resource paths from each spec are emitted verbatim into `client.Get(...)` calls. Spec A's commands reach `https://{tenant}/projects` correctly. Spec B's commands reach `https://{tenant}/pages` (wrong) instead of `https://{tenant}/wiki/api/v2/pages` (right). The `/wiki/api/v2` prefix from spec B's `servers[0].url` was silently discarded at merge time.

Result: every command from spec B returns 404. No generator warning. No build error. Caught only at runtime via live smoke test.

## Suspected root cause

The merge collapses N specs to a single `BaseURL` field by taking the host portion of one spec's `servers[0].url` and dropping any path component. There is no per-command source-spec tracking and no path-rewrite step to absorb the discarded prefix into endpoint paths.

Confidence: high. Direct repro: hand-patched the runtime client with a `resolveURL(baseURL, path)` helper that prefixes the missing path component for B's endpoints. Fix in 12 lines, but hardcoded for one specific shape.

## Suggested direction

Two viable approaches; recommend (a):

- **(a) Path-rewrite at merge time.** When N>1 specs are passed and any spec's `servers[0].url` contains a non-empty path component, rewrite that spec's endpoint paths to include the prefix at merge time. Then collapse all server URLs to the bare host portion. Single-pass, deterministic, no runtime cost. Emit a generator info log: `[multi-spec] prefixed N endpoints from <spec-id> with /<path-component>`.

- **(b) Per-command source-spec tagging.** Annotate each generated command with the source-spec id. Emit a `client.resolveURL(specID, path)` helper that routes to the correct base URL at runtime. More flexible but more runtime complexity; harder to audit.

(a) is small, deterministic, and matches what every OpenAPI client generator already does for single-spec runs (server URL prefix concatenated to operation path). It just generalizes that step to multi-spec input.

## Acceptance criteria

- **positive:** feed two fixture specs — A with `servers: [{url: https://example.com}]` + endpoint `/foo`; B with `servers: [{url: https://example.com/api/v2}]` + endpoint `/bar`. Generate with `--name combo`. Assert generated `bar` command's emitted Go calls `client.Get("/api/v2/bar", ...)`. Assert generated `foo` command's emitted Go is unchanged (`client.Get("/foo", ...)`).
- **negative (idempotent):** two specs with the SAME `servers[0].url` prefix → endpoints emitted unchanged. No double-prefixing.
- **regression:** single-spec generation unchanged.
- **log signal:** when merge prefixing fires, generator stderr shows `[multi-spec] prefixed N endpoints from <spec-id> with /<path-component>`.

## Frequency

`subclass:multi-spec-different-base-prefixes`. Concrete evidence today:

- Atlassian Cloud (this run, confirmed: every Confluence command 404'd until patched). Jira v3 at `https://{tenant}` + Confluence v2 at `https://{tenant}/wiki/api/v2`.

Plausible cross-API where this would also recur:

- Twilio combo CLIs across product surfaces (Programmable Voice at `/2010-04-01/...` vs Studio at `/v1/...`)
- Microsoft Graph multi-version (`/v1.0/...` vs `/beta/...`) — different versions, same shape

Today only one API directly verifies; predictable as combo CLIs (single binary, multiple product specs) become more common.

## Complexity

small — single pass over each spec's endpoints at merge time, prefix the path, collapse server URL to bare host. ~30-50 lines.

## Scope boundary

Does NOT handle the case where one spec's `servers[]` has multiple entries with different prefixes (still 1 base URL per spec). Does NOT auto-detect overlap conflicts (e.g. both specs claim `/api`). Does NOT solve cross-spec authentication where each spec needs different auth (out of scope; rare).

## Dependencies

None.

## Related issues

None matched. Step D scan: 0 matches across 89 retro issues for "multi-spec", "spec merge", or "baseURL collision".

## Anti-pattern observed

Silent 404s at runtime with no generator warning. The generator KNOWS when N>1 specs have non-overlapping server-URL prefixes — that signal is detectable at merge time. Even if path-rewrite isn't the chosen fix, the generator should at minimum emit a hard warning at generate time when this shape is detected: "merge spec base URLs disagree: spec A=`<host>`, spec B=`<host>/api/v2`. Endpoints from spec B will be relative to spec A's base. Confirm with --multi-spec-base-strategy=<flatten|prefix|reject>."

---

*Generated by `/printing-press-retro`*

hnshah:
## Summary
- Preserve each input spec's server base URL when merging multiple specs into one generated CLI
- Apply the source spec's base URL as a resource-level override when it differs from the merged CLI base URL
- Add a regression test for mixed root and path-prefixed OpenAPI servers

## Why
When `printing-press generate --spec a.yaml --spec b.yaml --name combo` merges specs with different `servers[].url` path prefixes, the merged CLI keeps only the first spec's `BaseURL`. Resources from later specs then emitted relative paths against the wrong base URL and hit 404s at runtime.

## Verification
- `go test ./internal/cli -run TestMergeSpecsPreservesPerSpecBaseURLPrefixes -count=1`
- `go test ./internal/cli ./internal/generator ./internal/openapi ./internal/spec`
- `go test ./...`
- Manual generate smoke test with two OpenAPI specs confirmed:
  - root spec emits `path := "/projects"`
  - prefixed spec emits `path := "https://tenant.example.com/wiki/api/v2/pages"`

Fixes #831


Champworks:
The fix direction looks right. I’d add two focused tests before calling this complete:

1. A nested/sub-resource case showing `BaseURL` is propagated recursively, not just to top-level resources.
2. A tiny golden/generated fixture or equivalent generation test proving the preserved server path actually reaches emitted commands/client calls.

That would close the main confidence gap: the code says it recurses, but the current coverage only proves the top-level merge behavior.

tmchow:
@hnshah can you resolve the conflicts?

PR #863, opened 2026-05-09
feat(graphql): cluster flat snake_case mutations like xxxCreate variants

NK4478:
## Summary

The GraphQL parser only recognized PascalCase mutation names (`xxxCreate` / `xxxUpdate` / `xxxDelete`) for resource clustering. APIs using flat snake_case mutations — Monday.com, older GraphQL APIs — failed with `at least one resource is required` even though the SDL parsed cleanly.

This PR fixes 3 root causes so flat snake_case clusters into the same `resourceEntities` map as PascalCase:

1. **`classifyMutation`** — added a verb-prefix table for snake_case (`create_`, `update_`, `change_`, `delete_`, `archive_`, `move_`, `add_`, `remove_`) that fires BEFORE the existing PascalCase substring scan. Fixes `delete_update → "delete"` (was matching `update`).

2. **`entityFromDirectReturnType`** — new helper that checks the mutation's return type directly before walking its fields. Fixes `create_item → Item` (was incorrectly walking to `Item.board → Board`).

3. **`stripSDLComments`** — strips standalone `"..."` description lines before resource extraction. Without it, descriptions like `"Root query type for the Dependencies service"` followed by `type Query { ... }` caused the block regex to capture all of Query as the body of a phantom `type for {}` block, leaving Query with 0 fields → 0 resources.

## Test plan

- [x] Added `testdata/graphql/monday-flat.graphql` — 200-line subset of Monday SDL exercising flat snake_case mutations
- [x] Added test in `internal/graphql/parser_test.go` asserting parser returns ≥3 resources (boards, items, updates) — fails on current main, passes after this PR
- [x] `go test ./...` — all existing tests still pass
- [x] Empirical: `printing-press generate --spec monday.graphql --dry-run` now produces **108 resources / 173 endpoints** (was 0 before the patch)

## Context

Found while building pp-monday for an internal MCP→CLI migration project. Same failure pattern documented in `docs/retros/2026-04-09-trigger-dev-retro.md` — both retros suggested the press should handle non-Linear-shaped GraphQL natively; this PR is the upstream fix.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/internal/graphql/parser.go
+++ b/internal/graphql/parser.go
@@ -16,11 +16,16 @@ import (
 )
 
 var (
-	docBlockRE = regexp.MustCompile(`(?s)""".*?"""`)
-	commentRE  = regexp.MustCompile(`(?m)#.*$`)
-	blockRE    = regexp.MustCompile(`(?s)\b(type|input|enum)\s+([A-Za-z_][A-Za-z0-9_]*)[^{=]*\{(.*?)\}`)
-	fieldRE    = regexp.MustCompile(`^\s*([A-Za-z_][A-Za-z0-9_]*)\s*(?:\((.*)\))?\s*:\s*([A-Za-z0-9_\[\]!]+)`)
-	scalarRE   = regexp.MustCompile(`(?m)^\s*scalar\s+([A-Za-z_][A-Za-z0-9_]*)\s*$`)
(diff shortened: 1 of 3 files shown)


tmchow:
Fix greptile code review feedback

tmchow:
Poor quality, closing.

PR #901, opened 2026-05-10
fix(cli): isolate generic resources by type

### LINKED ISSUE #890: Generic resources table uses single-column PRIMARY KEY, silently overwriting rows when resource types share natural IDs
> Filed by `/printing-press-retro` from a run of **Find Me Gluten Free**
> (scorecard 82/100, 100% verify pass).

## Summary

The generic `resources` table emitted by `internal/generator/templates/store.go.tmpl` uses `id TEXT PRIMARY KEY` instead of `PRIMARY KEY (resource_type, id)`. When a printed CLI legitimately stores multiple resource types keyed on the same natural ID — a primary resource and any local-only sub-resource (bookmark, watch, favorite, star, like, save) keyed on the primary's ID — the second write silently overwrites the first via `ON CONFLICT(id) DO UPDATE SET data = excluded.data`. The collision is data-corruption-grade: no error, no warning, and `resource_type` doesn't even change to reflect the new row's true type.

## Where to look

- **Component:** `generator`
- **Likely area:** `internal/generator/templates/store.go.tmpl` (table DDL at lines ~245-251 and the `upsertGenericResourceTx` function below it).
- **Triggered when:** A printed CLI defines two or more `resource_type` strings whose natural IDs overlap. This is the exact shape of every transcendence pattern that adds personal-state on top of remote data: bookmarks, watches, stars, likes, saved-items, notes, tags.

## What we observed

After running `bookmark add 6118524171452416` (a real Find Me Gluten Free biz_id, where the biz row was already cached from `places hydrate`), the biz row was silently replaced with the bookmark JSON:

```
$ sqlite3 ~/.local/share/find-me-gluten-free-pp-cli/data.db \
    "SELECT resource_type, length(data) FROM resources WHERE id LIKE '6118524171452416%';"
biz|135                       # was 9000+ bytes (the full Restaurant JSON-LD blob)
```

The 135-byte payload is the bookmark JSON (`{"biz_id":"6118524171452416","biz_name":"Pinky's","city":"Richmond","added_at":...,"note":"Anniversary spot"}`) — the original biz blob is gone. Subsequent `places stats --city richmond` returned `total: 0` and `places near` returned `[]`. Read-side queries had no idea the row had been corrupted.

The DDL (`internal/generator/templates/store.go.tmpl:245-251`):

```sql
CREATE TABLE IF NOT EXISTS resources (
    id TEXT PRIMARY KEY,
    resource_type TEXT NOT NULL,
    data JSON NOT NULL,
    synced_at DATETIME DEFAULT CURRENT_TIMESTAMP,
    updated_at DATETIME DEFAULT CURRENT_TIMESTAMP
)
```

The upsert at `store.go.tmpl:413` then does `ON CONFLICT(id) DO UPDATE SET data = excluded.data, synced_at = excluded.synced_at, updated_at = excluded.updated_at` — `resource_type` is never in the conflict clause and never gets updated either, leaving a row whose `resource_type` says one thing and whose `data` is something else entirely.

## Suspected root cause

Certain. The schema was authored under the assumption that `id` alone would be unique across all resource types in a single CLI's database. That holds for the generator's typed tables (one table per resource type, with the table's own UUID-style ID space) but breaks for the generic `resources` table once a CLI uses it for more than one resource type with overlapping natural IDs.

## Suggested direction

Change the table to a composite PK and update the upsert SQL accordingly:

```sql
CREATE TABLE IF NOT EXISTS resources (
    id TEXT NOT NULL,
    resource_type TEXT NOT NULL,
    data JSON NOT NULL,
    synced_at DATETIME DEFAULT CURRENT_TIMESTAMP,
    updated_at DATETIME DEFAULT CURRENT_TIMESTAMP,
    PRIMARY KEY (resource_type, id)
)
```

Update `upsertGenericResourceTx` to `ON CONFLICT(resource_type, id) DO UPDATE`. Bump `StoreSchemaVersion` from 1 to 2 and add a migration that rebuilds the PK in-place for existing CLIs (drop the lone-id PK, recreate with the composite, rebuild `resources_fts` since its rowid scheme depends on id).

The `Get(resourceType, id)`, `List(resourceType, limit)`, and `Search` helpers already filter by `resource_type` correctly — their behavior doesn't change.

## Acceptance criteria

- **positive:** Generated CLI inserts `(resource_type='biz', id='X')` then `(resource_type='bookmark', id='X')`. `Get("biz", "X")` returns the biz blob; `Get("bookmark", "X")` returns the bookmark blob. Both rows persist, both retrievable independently.
- **negative (migration):** Existing CLI running schema v1 with rows of one type only — after binary upgrade, schema migrates to v2, all rows still readable, FTS index rebuilt cleanly. No data loss.
- **negative (regen):** Running the generator against an existing CLI's source produces no diff to user-authored code outside generated `internal/store/` files.

## Frequency

every printed CLI that uses the generic resources table for more than one resource_type with overlapping natural IDs; subclass:`personal-state-on-resource`

## Complexity

medium

## Scope boundary

Don't change the `Get` / `List` / `Upsert` Go signatures — they already filter by `resource_type` correctly; only the underlying SQL changes. Don't touch typed-table emission for spec-defined resources (those are independent tables with their own PKs).

## Related issues

None.

## Artifacts

| Artifact | Link |
|----------|------|
| Retro document | https://files.catbox.moe/3lbh3o.md |
| Manuscripts (research + proofs) | https://files.catbox.moe/wj6l9d.zip |
| Generated CLI source code | https://files.catbox.moe/3qcbrg.zip |

---

tmchow:
## Summary

Generated CLIs now keep generic resources distinct by `(resource_type, id)`, so personal-state rows such as bookmarks can share a natural ID with the primary resource without overwriting it.

Existing v1 stores upgrade to schema v2 by rebuilding the generic `resources` table with the composite key and rebuilding `resources_fts` with type-scoped row IDs. FTS joins now include `resource_type` so search does not cross-return same-ID rows from other resource types.

## Test plan

- `rtk go test ./internal/generator -run 'TestGenerateStoreUpsertBatchDispatchesToTypedTable|TestGenerateStoreMigrateUsesBeginImmediate' -count=1`
- `rtk scripts/golden.sh verify`
- `rtk go test ./...`
- `rtk git diff --check`
- `rtk golangci-lint run ./...`

Closes #890

### DIFF
--- a/internal/generator/generator_test.go
+++ b/internal/generator/generator_test.go
@@ -2557,6 +2557,89 @@ func TestGenerateStoreUpsertBatchDispatchesToTypedTable(t *testing.T) {
 	runGoCommand(t, outputDir, "test", "./internal/store")
 }
 
+func TestGenerateSimilarCommandUsesCompositeResourceKey(t *testing.T) {
+	t.Parallel()
+
+	apiSpec := minimalSpec("similar-composite")
+	outputDir := filepath.Join(t.TempDir(), naming.CLI(apiSpec.Name))
+	gen := New(apiSpec, outputDir)
+	gen.VisionSet = VisionTemplateSet{
+		Store:    true,
+		Insights: []string{"insights/similar.go.tmpl"},
+	}
+	require.NoError(t, gen.Generate())
+
+	inlineTest := `package cli
+
+import (
+	"bytes"
+	"encoding/json"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"` + naming.CLI(apiSpec.Name) + `/internal/store"
+)
+
+func TestSimilarDisambiguatesOverlappingIDs(t *testing.T) {
+	dbPath := filepath.Join(t.TempDir(), "data.db")
+	db, err := store.Open(dbPath)
+	if err != nil {
+		t.Fatalf("open store: %v", err)
+	}
+	if err := db.Upsert("biz", "shared", []byte(` + "`" + `{"id":"shared","name":"Pinky restaurant"}` + "`" + `)); err != nil {
+		t.Fatalf("upsert biz: %v", err)
+	}
+	if err := db.Upsert("bookmark", "shared", []byte(` + "`" + `{"id":"shared","name":"Anniversary bookmark"}` + "`" + `)); err != nil {
+		t.Fatalf("upsert bookmark: %v", err)
+	}
+	if err := db.Upsert("biz", "other", []byte(` + "`" + `{"id":"other","name":"Pinky restaurant north"}` + "`" + `)); err != nil {
+		t.Fatalf("upsert peer: %v", err)
+	}
+	db.Close()
+
+	cmd := newSimilarCmd(&rootFlags{asJSON: true})
+	cmd.SetArgs([]string{"shared", "--db", dbPath})
+	err = cmd.Execute()
+	if err == nil || !strings.Contains(err.Error(), "matches multiple resource types") {
+		t.Fatalf("similar without --type err = %v, want overlap error", err)
+	}
+
+	cmd = newSimilarCmd(&rootFlags{asJSON: true})
+	var out bytes.Buffer
+	cmd.SetOut(&out)
+	cmd.SetArgs([]string{"shared", "--db", dbPath, "--type", "biz"})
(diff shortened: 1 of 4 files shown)


tmchow:
@greptile code review

PR #933, opened 2026-05-10
fix(generator): route receiver JSON helper through filters

### LINKED ISSUE #826: Eliminate flags.printJSON foot-gun for hand-written commands
## Goal

Hand-authoring a transcendence command using the natural-feeling JSON-emit call cannot silently drop `--select` / `--compact` / `--csv` / `--quiet`.

## Symptom

Two helpers exist with similar names:
- `flags.printJSON(cmd, v)` — does NOT honor any of the four flags
- `printJSONFiltered(cmd.OutOrStdout(), v, flags)` — honors all four

Spec-emitted endpoint commands use `printJSONFiltered` correctly. But `flags.printJSON` is exported, shorter to type, reads better in code review (receiver style), and is the natural choice when an agent hand-writes a transcendence command. Hand-written commands using `flags.printJSON` silently drop all four user-facing filter flags — output looks correct, but `--select id,name` returns full objects, `--compact` returns full payloads, etc.

## Why this matters across CLIs

Universal. Every printed CLI is required to have transcendence commands per the Phase 3 mandate in the skill. Every transcendence command that uses `flags.printJSON` is silently broken for agent-UX flags.

The Agent Build Checklist rule #2 in `skills/printing-press/SKILL.md` already documents this:
> Hand-written novel commands that build a Go-typed slice/struct and emit JSON MUST call `printJSONFiltered(cmd.OutOrStdout(), v, flags)` instead of `flags.printJSON(cmd, v)`.

But documentation is a patch over an API foot-gun. The rule didn't prevent the failure — agents follow receiver-style ergonomics naturally and reach for `flags.printJSON`.

## Evidence

`openrouter-pp-cli` build (2026-05-09). 8 hand-built transcendence commands all initially called `flags.printJSON`. Phase 5.5 polish skill caught it and migrated 9 files (12 call sites) to `printJSONFiltered`. Pre-polish, an agent calling `openrouter usage cost-by --select cron,cost --json` would have silently received unfiltered output.

The polish skill exists for taste-level fixes. Silent-correctness fixes shouldn't depend on it.

## Three options ranked by cost/risk

(a) **Rename + alias.** Rename `flags.printJSON` → `flags.printJSONUnfiltered`. Add `flags.printJSON = flags.printJSONFiltered` (the receiver-style call now does the right thing). Update generator templates to use the receiver style consistently.

(b) **Delete the unfiltered version.** If no caller actually needs unfiltered output, remove `flags.printJSON` entirely. Compile-time error guides the migration.

(c) **Add a vet check.** Generator emits a custom `go vet` analyzer that flags `flags.printJSON` calls. Adds tooling weight.

Option (b) is cleanest if no real caller needs unfiltered. Option (a) is the safe path. Skill instruction (rule #2) becomes redundant after either fix — that's the goal.

## Acceptance criteria

- positive (preferred): `flags.printJSON` is removed; `go build` fails on any caller; migration path is to `printJSONFiltered`
- positive (alternative): `flags.printJSON` is renamed `flags.printJSONUnfiltered`; a new `flags.printJSON` aliases `printJSONFiltered`
- negative: existing `printJSONFiltered` callers continue to work without changes
- The Agent Build Checklist rule #2 in the SKILL becomes redundant; remove the rule when the fix lands.

## Component

`internal/generator/templates/` (helpers.go template) + spec-emitted endpoint command templates if they reference the helper.

## Complexity

small (rename + alias, or delete + propagate)

hnshah:
## Summary

- route the generated receiver-style `flags.printJSON(...)` helper through `printJSONFiltered(...)`
- keep existing `printJSONFiltered(...)` callers working while making the natural helper honor `--select`, `--compact`, `--csv`, and `--quiet`
- add a generated-CLI regression test that exercises `flags.printJSON(...)` with `selectFields`
- update the affected golden fixture

Closes #826.

## Test plan

- `go test ./internal/generator -run TestRootFlagsPrintJSONHonorsOutputFlags -count=1`
- `go test ./internal/generator ./internal/cli`
- `scripts/golden.sh verify`
- `git diff --check`
- `go test ./...`


tmchow:
@hnshah can you fix linting please

PR #938, opened 2026-05-10
fix(cli): infer resource-prefixed IDField from item-schema properties

### LINKED ISSUE #915: profiler: populate resourceIDFieldOverrides from {resource}_id response-schema fields when x-resource-id is absent
## Problem

The profiler emits `resourceIDFieldOverrides` empty for any spec that doesn't carry `x-resource-id` annotations. When the API also has no plain `id` field on its response objects, the runtime falls back to the generic list (`id, ID, name, uuid, slug, key, code, uid`), fails extraction on every item, and silently stores 0 rows.

Surfaced in podscan: `Category` is `{category_id, category_name, category_display_name, ...}`. No `id`. No `x-resource-id`. Result: `sync` consumed 163 items, stored 0, exited 0. Same shape was latent on alerts, exports, exports-podcasts.

## Reported externally

mvanhorn/printing-press-library#383

## Library-side workaround

mvanhorn/printing-press-library#394 patches the published podscan copy by hand-populating both override maps (one in `internal/cli/sync.go`, one in `internal/store/store.go`). Recorded in `.printing-press-patches.json` so a regen knows what to keep.

## Suggested fix in the profiler

When no `x-resource-id` annotation is present on a path-item, derive the override from the response schema:

1. For each flat resource (and dependent resource), inspect the array-item schema returned by the resource's list endpoint.
2. If the schema has a property literally named `{singular_resource_name}_id` (e.g. `category_id` for resource `categories`), use that as the override.
3. Fallback chain after the singular-id heuristic, before the generic list: `{resource}_id`, `{resource}_uuid`, `{resource}_guid`.

This is a heuristic, not a contract — `x-resource-id` should still win when present. But the heuristic catches the common case (REST APIs that prefix every key field with the resource name) without requiring spec annotations.

## Affected published CLIs

Worth grepping the library tree for empty `resourceIDFieldOverrides{}` blocks and auditing each:

\`\`\`
grep -rl 'var resourceIDFieldOverrides = map\[string\]string{$' library/*/*/internal/
\`\`\`

CLIs where the generic fallback (`id, ID, name, ...`) doesn't match the API's actual key field will exhibit the same silent-drop behavior.

## Severity

High visibility — silent failure with exit 0. Most users won't notice unless they read the JSON event stream or compare wire counts to stored counts. Promoting `all_items_failed_id_extraction` to "critical" by default (instead of warning) is a separate but related ask from the reporter.

tmchow:
## Summary

Adds a resource-prefixed key heuristic to the OpenAPI parser's IDField fallback chain so APIs whose item schemas key off `<singular_resource>_id` (instead of a bare `id`) get a non-empty `Endpoint.IDField` at parse time. Without this, sync silently consumes pages and stores zero rows, exiting with code 0 — exactly the failure mode reported in #915 against podscan's `Category` resource.

The heuristic slots in between the existing `id` tier and the existing `name` tier:

1. `x-resource-id` extension (unchanged)
2. bare `id` (unchanged)
3. **NEW**: `<singular>_id`, then `<singular>_uuid`, then `<singular>_guid` against item-schema properties
4. bare `name` (unchanged)
5. first scalar in `required[]` (unchanged)

Resource name and property name are both normalized via `toSnakeCase` before comparison, so `categoryId` ↔ `category_id` and `/auth-tokens` ↔ `auth_token_id` match through the same path. `singularizeIdentifier` handles the irregular plurals (`categories` → `category`, `entries` → `entry`) and only singularizes the trailing token of a snake-cased identifier so multi-word resources keep their prefix.

## Behavior

- Item schema with `id` and `category_id` → `id` wins (REST convention preserved).
- Item schema with only `category_id` → `category_id` wins (was empty before).
- Item schema with `name` and `category_id` → `category_id` wins.
- `_id` is preferred over `_uuid`, which is preferred over `_guid`.
- Synthetic endpoints with no resource name skip the new tier and fall through unchanged.

## Out of scope

- Promoting `all_items_failed_id_extraction` to a critical severity (called out in the issue as a related-but-separate ask).
- Library-wide regen sweep to refresh published CLIs that ship empty `resourceIDFieldOverrides{}`.

## Test plan

- [x] `rtk go test ./internal/openapi/ -run TestParseIDFieldResourcePrefixedHeuristic -v` — 10 new sub-cases covering plural, singular, kebab path, camelCase property, suffix precedence, and existing-tier interactions.
- [x] `rtk go test ./...` — 3428 tests in 34 packages, all passing.
- [x] `rtk scripts/golden.sh verify` — 16 cases, all passing (no goldens shifted).
- [x] `rtk go vet ./...`, `rtk golangci-lint run ./...`, `rtk git diff --check` — clean.

Closes  #915

### DIFF
--- a/internal/openapi/parser.go
+++ b/internal/openapi/parser.go
@@ -2865,17 +2865,16 @@ func resourcePrefixedIDField(schema *openapi3.Schema, resourceName string) strin
 	if singular == "" {
 		return ""
 	}
-	suffixes := []string{"_id", "_uuid", "_guid"}
-	for _, suffix := range suffixes {
+	// Sort property names so behavior is deterministic across Go map
+	// iteration order; this only matters when a schema declares multiple
+	// keys that snake-case to the same target, which is unusual.
+	propNames := make([]string, 0, len(schema.Properties))
+	for name := range schema.Properties {
+		propNames = append(propNames, name)
+	}
+	sort.Strings(propNames)
+	for _, suffix := range []string{"_id", "_uuid", "_guid"} {
 		target := singular + suffix
-		// Sort property names so behavior is deterministic across Go map
-		// iteration order; this only matters when a schema declares multiple
-		// keys that snake-case to the same target, which is unusual.
-		propNames := make([]string, 0, len(schema.Properties))
-		for name := range schema.Properties {
-			propNames = append(propNames, name)
-		}
-		sort.Strings(propNames)
 		for _, propName := range propNames {
 			if toSnakeCase(propName) == target {
 				return propName
@@ -2906,6 +2905,13 @@ func singularizeIdentifier(s string) string {
 		"statuses":   "status",
 		"addresses":  "address",
 		"analyses":   "analysis",
+		// `<stem>+s` plurals on `ie`/`ix`-ending stems that the generic
+		// `ies → y` rule would otherwise mangle (`movies → movy`).
+		"movies":   "movie",
+		"series":   "series",
+		"matrices": "matrix",
+		"indices":  "index",
+		"vertices": "vertex",
 	}
 	if singular, ok := irregulars[last]; ok {
 		return prefix + singular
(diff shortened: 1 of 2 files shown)


tmchow:
@greptileai code review

tmchow:
@greptileai code review

PR #946, opened 2026-05-10
fix(cli): infer resource-prefix IDField when generic chain returns empty

tmchow:
## Summary

The OpenAPI parser's IDField fallback chain (`id`, `name`, required-scalar) returned `""` for APIs whose item schemas prefix every key field with the resource name (`{category_id, category_name, ...}`) and whose specs don't annotate `required:`. Empty IDField → empty `resourceIDFieldOverrides` map → runtime `extractID` falls through the generic list (`id, ID, name, uuid, slug, key, code, uid`) → `all_items_failed_id_extraction` → 0 rows stored, exit 0. Surfaced on Podscan (`Category` shape) and Spypoint runs.

This adds a **tier 5** to `resolveIDFieldFromResponseSchema` that probes the array-item schema for `{singular_resource_name}_id`, `{resource}_id`, and the same patterns with `_uuid` / `_guid` suffixes. Kebab-case resource slugs (e.g. `markers-types`) map to snake_case field names (`markers_types_id`) so the schema lookup matches typical JSON shapes.

## Behavior

- **Tier 1 (`x-resource-id`)** — unchanged, still wins.
- **Tiers 2-4 (`id` / `name` / required-scalar)** — unchanged.
- **Tier 5 (NEW)** — fires only when tiers 2-4 return `""`. APIs with a plain `id` field keep their current IDField; the test `tier 2 (id) still wins when both id and {resource}_id present` pins this contract.
- Property values must be scalar to match (object-typed `widget_id` is skipped, falls through to `_uuid` / `_guid`).

## Files

- `internal/openapi/parser.go` — extends `resolveIDFieldFromResponseSchema` and adds `resolveIDFieldFromResourceNamePrefix` helper. Call site passes `targetResourceName`.
- `internal/spec/spec.go` — exports `Singularize` as a public wrapper over the existing private `singularize` so the parser can reuse the irregulars table without duplicating it.
- `internal/openapi/parser_test.go` — eight new table-driven cases covering plural→singular, already-singular, kebab-case, `_uuid` / `_guid` fallbacks, scalar-only requirement, the bottom-out case, and the existing-tier-wins guarantee.

## Verification

- `go test ./...` — 3414 passed
- `go vet ./...` — clean
- `go fmt ./...` — clean
- `scripts/golden.sh verify` — 16/16 cases passed (no generator-output drift)

Refs #915

### DIFF
--- a/internal/cli/release_test.go
+++ b/internal/cli/release_test.go
@@ -78,6 +78,38 @@ func TestVersionConsistencyAcrossFiles(t *testing.T) {
 		"plugin.json and version.go hardcoded version must match")
 }
 
+func TestInternalSkillMinimumBinaryVersionsTrackMajor(t *testing.T) {
+	// Skill frontmatter `version` values are not release-managed. The
+	// executable compatibility contract is `min-binary-version`; keep the
+	// frontmatter and the duplicated setup-contract comment in sync.
+	want := fmt.Sprintf("%d.0.0", majorVersion(t, version.Version))
+	paths := []string{
+		"../../skills/printing-press/SKILL.md",
+		"../../skills/printing-press-catalog/SKILL.md",
+		"../../skills/printing-press-polish/SKILL.md",
+		"../../skills/printing-press-publish/SKILL.md",
+		"../../skills/printing-press-score/SKILL.md",
+	}
+
+	frontmatterRe := regexp.MustCompile(`(?m)^min-binary-version:\s*"?([^"\n]+)"?\s*$`)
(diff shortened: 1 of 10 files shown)


tmchow:
@greptileai code review

tmchow:
Closing in favor of a clean branch — the original was branched from a worktree base that had not been rebased onto main, so the PR diff included orphan commits from the squash-merged #854. Reopening from a fresh branch off main with the same fix plus the Greptile dead-guard fix.

PR #947, opened 2026-05-10
fix(cli): emit form-encoded request bodies

### LINKED ISSUE #921: Generator can't emit form-encoded request bodies (OAuth tokens, reverse-engineered APIs)
> Filed by `/printing-press-retro` from a run of **Resy**
> (scorecard 85/100, 100% verify pass).

## Summary

The generator can only emit JSON-encoded request bodies via the auto-generated
`c.Post(path, body)`, which silently breaks against any endpoint that requires
`application/x-www-form-urlencoded`. This includes every OAuth 2.0 token
endpoint (RFC 6749 mandates form encoding for `/oauth/token`) and every
reverse-engineered API that mirrors a Laravel/Rails-style backend. For Resy,
both `/3/venuesearch/search` and `/3/book` return HTTP 400 with "Invalid data
received" until the body is form-encoded with the JSON payload as a single
`struct_data` form param. The fix path the agent had to take — write a
`post_form.go` helper bypassing `c.Post()` entirely, then rewrite the
generated `venues_search.go` to use it — is exactly what the spec annotation
should drive.

## Where to look

- **Component:** generator (primary) + spec-parser (secondary)
- **Likely areas:**
  - `internal/spec/spec.go` — add `RequestFormat string` field to `Endpoint`; parse `request_format` from internal YAML.
  - `internal/openapi/parser.go` — auto-detect `application/x-www-form-urlencoded` in `requestBody.content` and set `Endpoint.RequestFormat = "form"`.
  - `internal/generator/templates/client.go.tmpl` — add a `PostForm(path, url.Values)` method (or extend `Post` with a format switch).
  - `internal/generator/templates/<resource>_<endpoint>.go.tmpl` — branch on `Endpoint.RequestFormat` to call the right client method and assemble form values from body fields.
- **Triggered when:** any spec declares a POST endpoint with `application/x-www-form-urlencoded` content type, or the agent annotates an internal-YAML endpoint with `request_format: form`.

## What we observed

- Resy `/3/venuesearch/search` direct probe with form-encoded `struct_data=<json>` body: 200 OK, 36 hits returned for "carbone" query.
- Same endpoint via auto-generated `c.Post()` (which JSON-marshals): HTTP 400 "Invalid data received."
- The agent had to write `internal/cli/post_form.go` (52 lines) using `c.HTTPClient` + `url.Values.Encode()` directly, then rewrite `internal/cli/venues_search.go` (~155 lines) to call it. Total bypass: ~92 lines of agent-authored code that the generator should have absorbed.
- The OAuth subset alone is broad: GitHub Apps token exchange, Slack `/oauth.v2.access`, every Stripe Connect refresh, every internal SSO using OAuth 2.0 client_credentials/refresh_token grants. RFC 6749 §4.4.2 mandates form encoding for these.

## Suspected root cause

The generator's request-body emission path treats all bodies as JSON. The
internal-YAML `Endpoint` schema has no field to declare otherwise. The OpenAPI
parser doesn't read `requestBody.content`'s media-type keys (or reads them but
the generator doesn't act on the result). The client template only emits
`json.Marshal` + `Content-Type: application/json` for non-GET requests.

## Suggested direction

Three coordinated changes:

1. **Spec parser:** add `RequestFormat string` to `Endpoint` (internal YAML field `request_format`, values `json` or `form`).
2. **OpenAPI parser:** auto-detect `application/x-www-form-urlencoded` in `requestBody.content` and populate `Endpoint.RequestFormat = "form"`.
3. **Generator client template:** branch on `RequestFormat`. Add either a `PostForm(path, url.Values)` method or extend `Post` with a format switch. Per-endpoint emission picks the right call.
4. **Per-field encoding (optional follow-up):** when `request_format: form`, allow individual body fields to declare `encoding: json-string` so a single field can carry a JSON-stringified payload (Resy's `struct_data` shape).

## Acceptance criteria

- positive: a generated CLI with an OAuth token endpoint (`POST /oauth/token` declaring `application/x-www-form-urlencoded`) successfully exchanges credentials.
- positive: an internal YAML endpoint declaring `request_format: form` with a single `struct_data` string field sends `Content-Type: application/x-www-form-urlencoded` and `struct_data=<value>` URL-encoded.
- negative: a spec without any form-encoded endpoints generates byte-identical output to today (golden tests for ≥3 representative existing CLIs unchanged).
- regression: every generated client_test.go and per-endpoint test continues to pass.

## Frequency

subclass: APIs with at least one form-encoded endpoint (~30-40% of catalog,
including all OAuth-using APIs)

## Complexity

medium

## Scope boundary

REST endpoints only. Not handling `multipart/form-data` (file uploads) — that's
a separate template gap. Not auto-rewriting JSON-shaped specs to form just
because a single endpoint requires form (only flagged endpoints change behavior).

## Dependencies

None.

## Related issues

None.

## Artifacts

| Artifact | Link |
|----------|------|
| Retro document | https://files.catbox.moe/61t2ll.md |
| Manuscripts (research + proofs) | https://files.catbox.moe/rmdlm6.zip |
| Generated CLI source code | https://files.catbox.moe/6cowi3.zip |

---

tmchow:
## Intent

Generator can't emit `application/x-www-form-urlencoded` bodies, so generated CLIs JSON-marshal POST bodies and break against any endpoint that requires form encoding. RFC 6749 mandates form encoding for `/oauth/token` so this hits every OAuth 2.0 client_credentials/refresh_token grant; it also hits reverse-engineered APIs that mirror Laravel/Rails-style backends (e.g. Resy `/3/venuesearch/search`).

Issue: Closes #921

## Approach

Mirrors the multipart precedent (#904). The shape is identical: `Endpoint.RequestContentType` already flows through the OpenAPI parser; the generator now also branches on `application/x-www-form-urlencoded`.

- `internal/generator/generator.go` — adds `endpointUsesForm`, `hasFormRequest`, and `formBodyMaps` helpers; the existing `mcpParamBindings` carries `RequestContentType` for both multipart and form so MCP handlers can pick the right wire shape; `clientTemplateData.HasFormRequest` gates the new client-side declarations.
- `internal/generator/templates/client.go.tmpl` — emits `PostForm`/`PutForm`/`PatchForm` (and `*WithHeaders` variants), `formRequestBody` type, and `encodeFormBody` only when `.HasFormRequest`. The `do()` body-encoding switch now composes multipart and form: a single `contentType` variable is gated on `or .HasMultipartRequest .HasFormRequest`.
- `internal/generator/templates/command_endpoint.go.tmpl` & `command_promoted.go.tmpl` — adds `$isForm` and a parallel POST/PUT/PATCH branch that builds `url.Values{}` via `formBodyMaps` and calls `c.PostForm` / `c.PutForm` / `c.PatchForm`. The stdin-body flag is suppressed for form endpoints (form bodies aren't JSON), and the required-flag check moves to top-level alongside the multipart branch.
- `internal/generator/templates/mcp_tools.go.tmpl` — MCP handler now branches on `RequestContentType` for form too, populating `url.Values` and dispatching to `c.PostForm`/`c.PutForm`/`c.PatchForm`.

Per-field `encoding: json-string` (Resy `struct_data` style) is listed as optional in the issue and is intentionally out of scope; `formBodyMaps` validates JSON for object/array/json-string fields before writing them as a single form value, which covers the Resy shape.

## Scope

Primary area: generator (templates, client emission, MCP wiring) + parser test coverage.

Why this belongs in this repo: every printed CLI that wraps an OAuth 2.0 token endpoint or a form-encoded reverse-engineered API breaks the same way; the right fix is in the generator so future prints don't have to hand-roll a `post_form.go` workaround.

## Risk

- Specs that declare no form-encoded endpoints generate byte-identical output. `scripts/golden.sh verify` passes all 16 cases (acceptance criterion: regression goldens unchanged).
- A new `TestParseJSONPreferredOverFormUrlencoded` asserts that when a spec declares both `application/json` and `application/x-www-form-urlencoded`, the parser still picks JSON, so existing OpenAPI specs that offer both content types keep their current wire shape.
- Form endpoints don't go through stdin or print bodies in `--dry-run` (the dry-run pretty-printer still tries to JSON-decode the body, so form payloads won't render in preview). Acceptable for now; can be a follow-up if needed.
- No changes to existing multipart paths.

## Test plan

- [x] `rtk go test ./internal/openapi -run "TestParseFormUrlencoded|TestParseJSONPreferredOverFormUrlencoded" -count=1`
- [x] `rtk go test ./internal/generator -run "TestGenerateFormRequestBodyUsesFormClient|TestGenerateNonFormSpecOmitsFormHelpers" -count=1`
- [x] `rtk go test ./...` (3421 tests pass)
- [x] `rtk scripts/golden.sh verify` (all 16 cases pass; non-form output byte-identical)
- [x] `rtk golangci-lint run ./...`
- [x] `rtk git diff --check`
- [x] `rtk go build -o ./printing-press ./cmd/printing-press`
- [x] Generator test compiles and runs the generated CLI under `go mod tidy && go build ./...`.


tmchow:
@greptileai code review

tmchow:
@greptileai code review

PR #957, opened 2026-05-10
feat(cli): emit nested-object body fields as parent-prefixed flags

### LINKED ISSUE #942: generator: emit nested-object body fields when spec declares type: object with fields
## Goal

Make the generator emit nested-object body fields when the internal spec declares `body[].type: object` with a populated `fields` array. Today the templates serialize every body field as a flat string assignment regardless, so structured payloads like Microsoft Graph's `start: {dateTime, timeZone}` break.

## Evidence

Surfaced during the **outlook-calendar** run (Microsoft Graph v1.0). The first live POST to `/me/events` returned HTTP 400 with `Property end in payload has a value that does not match schema.` because the wire shape was `"start": "2026-06-09T15:00:00"` (string) instead of `{ "dateTime": ..., "timeZone": "UTC" }`. Five generated commands (`events_create`, `events_update`, `events_forward`, `events_snooze`, `availability_schedule`, `availability_find`) all needed hand-built body shapers in a new `internal/cli/graph_body.go`.

## Cross-API stress test

Concrete APIs with the same shape, with evidence:

- **Microsoft Graph** (this run) — every event/message/contact/drive endpoint
- **Stripe** — `address: {line1, line2, city, country}` on customer/charge create
- **Notion** — `parent: {database_id}` on page create; `properties: {Name: {title: [...]}}` everywhere
- **GitHub** — `branch_protection: {required_status_checks: {strict, contexts}}` on protection-update endpoints

## Proposed fix

`internal/generator/templates/command_endpoint.go.tmpl` and `command_promoted.go.tmpl` recurse on `Param.Fields` when emitting body construction. Spec parser already accepts `fields` per the format doc; templates currently ignore them for body params.

## Acceptance criteria

- positive: spec with `body: [{name: start, type: object, fields: [{name: dateTime, type: string}, {name: timeZone, type: string}]}]` emits `body["start"] = map[string]any{"dateTime": flagStart, "timeZone": flagTimeZone}`.
- negative: flat `body: [{name: name, type: string}]` still emits `body["name"] = flagName`.
- golden: end-to-end fixture exercises a Microsoft-Graph-shaped spec.

## Counter-check

Does NOT hurt APIs that don't use nested-object bodies — gated on `Param.Fields` being populated.

## Source

Retro for **outlook-calendar** (2026-05-10): `~/printing-press/manuscripts/outlook-calendar/20260510-094714/proofs/20260510-141437-retro-outlook-calendar-pp-cli.md`. Finding F1 + WU-1.

tmchow:
Closes #942.

## Summary

When a body `Param` declared `type: object` with non-empty `Fields`, the generator previously ignored the nested structure and emitted a single JSON-string flag — e.g. `--start='{\"dateTime\":\"...\",\"timeZone\":\"...\"}'`. Microsoft Graph (DateTimeTimeZone), Stripe (Address), Notion (parent), and similar APIs that wrap properties under an object force users to hand-write JSON for every object body field. The retro for the outlook-calendar CLI documented five hand-written body shapers in `internal/cli/graph_body.go` that should not have been needed.

After this change, `bodyMap` recurses into `Param.Fields` and builds a `map[string]any` whose parent key is only set when at least one child flag was provided. Three new template helpers consolidate the body-section loops into Go so var declarations, cobra flag registrations, and required-flag validation all recurse uniformly with parent-prefixed identifiers (so `start.dateTime` and `end.dateTime` collide on neither variable nor flag name).

For the events-create example in #942:

```
# before
events create --start='{\"dateTime\":\"2026-06-09T15:00:00\",\"timeZone\":\"UTC\"}'

# after
events create --start-date-time=2026-06-09T15:00:00 --start-time-zone=UTC
```

## Implementation

- `internal/generator/generator.go` — modify `bodyMap` to recurse into `Fields` and emit a nested `map[string]any` with a `len > 0` guard. Add `bodyVarDecls`, `bodyFlagRegs`, `bodyRequiredChecks` helpers (plus `bodyUsesFlatEmission`) and register them in the template FuncMap.
- `internal/generator/templates/command_endpoint.go.tmpl` and `command_promoted.go.tmpl` — replace the inline `{{- range .Endpoint.Body}}` loops at the var-decl, required-check, and flag-reg sites with calls to the new helpers.
- Multipart and form-encoded endpoints stay on the flat emission path. `multipartBodyMaps` / `formBodyMaps` serialize object-typed parents as JSON-string fields, so flattening parents at the var/flag site would break them.
- New tests in `internal/generator/body_map_test.go` cover nested behavior, deep nesting, scalar-and-nested coexistence, multipart/form-encoded staying flat, and parent-prefixed required checks. `internal/generator/body_nested_object_test.go` is an end-to-end check that runs the generator against a Microsoft-Graph-shaped spec, asserts the expected nested vars / flags / checks / map literals appear, and verifies the generated file parses as Go.

## Verification

- `go test ./...` — 3447 pass.
- `scripts/golden.sh verify` — all 16 cases pass (no fixture has nested-object body fields, so the helper preserves byte-identical output for every existing case).
- `go vet ./...` — clean.
- `golangci-lint run ./...` — clean.

## Impact on existing CLIs

Any spec where the OpenAPI parser auto-populates `Fields` (most APIs with object-shaped body properties) will see auto-generated flag names change on next regen — `--customer` becomes `--customer-line-1`, `--customer-line-2`, etc. The auto-generated flag set is derived from the spec, not a stable downstream contract per AGENTS.md (\"generator-output diffs that don't remove or rename a documented surface are `docs(...)` or `fix(...)` — not `feat(...)!`\"), so this is tagged `feat(cli)` rather than `feat(cli)!`. Reviewers, please flag if you disagree.

## Known limitation

If a spec marks a parent body field `required: true` but its child `Fields` are not individually required, this PR does not synthesize an \"at least one child set\" check. The CLI will pass an empty body and the API will reject. The Microsoft Graph case in #942 marks the children required too, so this does not bite the cited use case. Filed as a known follow-up.

## Test plan

- [x] `go test ./...`
- [x] `scripts/golden.sh verify`
- [x] `go vet ./...`
- [x] `golangci-lint run ./...`
- [ ] Reviewer to confirm `feat(cli)` (no `!`) is the right semver tag here.

### DIFF
--- a/internal/generator/body_map_test.go
+++ b/internal/generator/body_map_test.go
@@ -143,3 +143,278 @@ func TestBodyMap_IdentName(t *testing.T) {
 		t.Errorf("expected wire key to use Name (not IdentName), got: %s", got)
 	}
 }
+
+// TestBodyMap_NestedObject verifies that body params declaring
+// type=object with non-empty Fields render a nested-map block in
+// place of the JSON-string parse path. The wire key is the parent's
+// Name; field keys are each leaf's Name. Field-flag variables are
+// parent-prefixed (bodyStartDateTime, not bodyDateTime) so two
+// parents that share a field name do not collide.
+func TestBodyMap_NestedObject(t *testing.T) {
+	t.Parallel()
+	got := bodyMap([]spec.Param{{
+		Name: "start",
+		Type: "object",
+		Fields: []spec.Param{
+			{Name: "dateTime", Type: "string"},
+			{Name: "timeZone", Type: "string"},
+		},
+	}}, "\t")
+	want := "\t{\n" +
+		"\t\tnestedStart := map[string]any{}\n" +
+		"\t\tif bodyStartDateTime != \"\" {\n" +
+		"\t\t\tnestedStart[\"dateTime\"] = bodyStartDateTime\n" +
+		"\t\t}\n" +
+		"\t\tif bodyStartTimeZone != \"\" {\n" +
+		"\t\t\tnestedStart[\"timeZone\"] = bodyStartTimeZone\n" +
+		"\t\t}\n" +
+		"\t\tif len(nestedStart) > 0 {\n" +
+		"\t\t\tbody[\"start\"] = nestedStart\n" +
+		"\t\t}\n" +
+		"\t}\n"
+	if got != want {
+		t.Errorf("bodyMap nested mismatch.\n got:\n%s\nwant:\n%s", got, want)
+	}
+}
+
+// TestBodyMap_NestedObject_PreservesScalarSiblings verifies that
+// nested and flat body params can coexist: nested produces a block,
+// scalars keep their existing if-then-set form.
+func TestBodyMap_NestedObject_PreservesScalarSiblings(t *testing.T) {
+	t.Parallel()
+	got := bodyMap([]spec.Param{
+		{Name: "subject", Type: "string"},
+		{Name: "start", Type: "object", Fields: []spec.Param{{Name: "dateTime", Type: "string"}}},
+	}, "\t")
+	if !strings.Contains(got, `if bodySubject != "" {`) {
+		t.Errorf("scalar branch missing, got:\n%s", got)
+	}
+	if !strings.Contains(got, `body["subject"] = bodySubject`) {
+		t.Errorf("scalar wire-set missing, got:\n%s", got)
+	}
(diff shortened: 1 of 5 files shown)


tmchow:
@greptileai code review

tmchow:
@greptileai code review

Issue #918, opened 2026-05-10
Systemic output-pipeline bugs: --csv/--quiet/--plain ignored when piped, --all bails on nested-envelope pagination


gregce:
While dogfooding a freshly-printed Tella CLI ([printing-press-library#374](https://github.com/mvanhorn/printing-press-library/pull/374)), three connected bugs surfaced in the output pipeline. All three originate in the generator templates and OpenAPI parser, so they ship into every printed CLI on print.

## TL;DR

| Bug | Symptom | Affected files in public library | Affected CLIs |
|-----|---------|----------------------------------|---------------|
| **Gate forces JSON when piped** | `--csv`/`--quiet`/`--plain` ignored when stdout is not a terminal (i.e., the common case for agents and shell pipelines) | **467 of 502** `*_list.go` files | virtually all CLIs with list commands |
| **Empty pagination paths** | `--all` exits after page 1 when API uses nested pagination (e.g., `pagination.nextCursor`) | **145 of 502** `*_list.go` files | pokeapi (47), sentry (39), scrape-creators (24), roam (9), airbnb (8), mercury (7), dub (6), and 16 others |
| **Silent `--quiet`** + missing `printPlain` | `--quiet` returns nothing despite "one value per line" help text; `--plain` flag exists but no renderer | **66 of 102** `helpers.go` files | every CLI with the standard helpers template |

## Reproduction (Tella, pre-fix)

```bash
$ tella-pp-cli videos list --all --json | jq '.results | length'
15        # workspace has 232 videos; --all returned page 1 only

$ tella-pp-cli videos list --csv --select id,name | head -3
{
  "meta": {"source": "live"},
  "results": {                  # JSON envelope, not CSV

$ tella-pp-cli videos list --quiet --select id
                                # nothing
```

### 1. `command_endpoint.go.tmpl:422` and `:472` — routing gate

```go
if flags.asJSON || !isTerminal(cmd.OutOrStdout()) {
    // emit JSON envelope and return
}
```

Conflates "user explicitly chose JSON" with "stdout is piped, so default to JSON". When the user pipes `--csv`/`--quiet`/`--plain` (agents, `head`, `awk`, `grep` — all common), the gate fires before format flags are checked. JSON wins regardless.

**Fix:**
```go
if flags.asJSON || (!isTerminal(cmd.OutOrStdout()) && !flags.csv && !flags.quiet && !flags.plain) {
```

[...]

PR #968, opened 2026-05-10
fix(cli): route explicit --csv/--quiet/--plain above piped-pipe gate

tmchow:
## Intent

Fix bug #1 of [#918](https://github.com/mvanhorn/cli-printing-press/issues/918) — the `--csv`/`--quiet`/`--plain` flags were ignored when stdout is piped. The gate at the top of `command_endpoint.go.tmpl` and `command_promoted.go.tmpl` read `flags.asJSON || !isTerminal(...)`, so piped output (the common case for agents and shell pipelines) silently rendered as a JSON envelope regardless of an explicit format flag.

Issue: Refs #918

## Approach

Add `&& !flags.csv && !flags.quiet && !flags.plain` to the auto-JSON clause at every gate site so an explicit format choice opts out of the auto-JSON path and falls through to the standard pipeline (`printOutputWithFlags`). The smart-default principle (terminal=table, pipe=JSON) now yields to the user's explicit override.

The now-redundant `if flags.csv { return printOutputWithFlags(...) }` short-circuit in `command_promoted.go.tmpl` is removed: post-fix, `--csv` reaches the standard pipeline through the same path as `--quiet`/`--plain`, so the per-flag carve-out is dead code.

## Scope

Primary area: generator templates (`internal/generator/templates/command_endpoint.go.tmpl`, `command_promoted.go.tmpl`).

Bugs #2 (nested-envelope pagination introspection in `internal/openapi/parser.go`) and #3 (silent `printQuiet`, missing `printPlain` in `helpers.go.tmpl`) from the same issue are independent surfaces and ship as follow-up PRs. Bug #3 in particular needs design choices for the `printQuiet` selector and `printPlain` column shape that don't fit a surgical fix.

## Risk

Behavior change for piped consumers passing an explicit format flag:

- **`--csv` piped**: was JSON envelope → now CSV (the documented behavior).
- **`--quiet` piped**: was JSON envelope → now silent (matches the existing `printOutputWithFlags` `if flags.quiet { return nil }` branch). This is bug #3 becoming visible; tracked in the issue, will be fixed in the follow-up PR that implements `printQuiet` per the help text.
- **`--plain` piped**: was JSON envelope → now JSON via the `printOutput` self-defaulting path. Same end behavior; `--plain` is unimplemented and tracked in bug #3.
- **`--json` piped or no format flag piped**: unchanged. The gate still emits the auto-JSON envelope.

Per AGENTS.md, this is `fix(cli):` not `feat(cli)!`: it corrects incorrect behavior of an already-shipping flag and does not remove or rename a documented surface.

## Output Contract

Generator-output diff: every `*_list.go` / mutating-endpoint file and promoted-command file emits the new gate clause. Sample (`projects_list.go`):

```diff
- if flags.asJSON || !isTerminal(cmd.OutOrStdout()) {
+ if flags.asJSON || (!isTerminal(cmd.OutOrStdout()) && !flags.csv && !flags.quiet && !flags.plain) {
```

Promoted commands additionally drop the `flags.csv` carve-out block (4 lines).

12 golden fixtures across 3 cases (`generate-golden-api`, `generate-public-param-names`, `generate-tier-routing-api`) regenerated to reflect the new gate. No other golden cases touched.

## Verification

- `go test ./...` — 3436 passed
- `go vet ./...` — clean
- `scripts/golden.sh verify` — 16 cases pass post-update
- `git diff --check` — clean

## Tests

- `TestPipedJsonGateRespectsExplicitFormatFlags` — new; reads both gate templates and asserts the new gate string is present and the stale bare gate is absent.
- `TestGeneratedOutput_MutatingCommandsHaveEnvelope` — updated; existing assertion now expects the format-flag escape clause.

### DIFF
--- a/internal/generator/generator_test.go
+++ b/internal/generator/generator_test.go
@@ -3195,8 +3195,10 @@ func TestGeneratedOutput_MutatingCommandsHaveEnvelope(t *testing.T) {
 	assert.Contains(t, content, `"resource":`)
 	assert.Contains(t, content, `"status":   statusCode`)
 	assert.Contains(t, content, `"success":  statusCode >= 200 && statusCode < 300`)
-	// Envelope fires on both --json and auto-JSON (piped/non-TTY)
-	assert.Contains(t, content, `flags.asJSON || !isTerminal(cmd.OutOrStdout())`)
+	// Envelope fires on --json and on piped output, but explicit format flags
+	// (--csv, --quiet, --plain) opt out of the auto-JSON path so piped agents
+	// that asked for a non-JSON format actually get it.
+	assert.Contains(t, content, `flags.asJSON || (!isTerminal(cmd.OutOrStdout()) && !flags.csv && !flags.quiet && !flags.plain)`)
 
 	// --quiet is respected before envelope output
 	assert.Contains(t, content, "if flags.quiet {")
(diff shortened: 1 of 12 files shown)


tmchow:
@greptileai code review

tmchow:
Greptile observations noted as accepted-for-later (both stylistic, no behavior impact):

1. `NotContains` test assertion is fragile against a future template comment that contains the bare gate string. The current form correctly guards today's contract; if the template ever needs that string in a comment, the test message will guide the fix.
2. The `if flags.quiet { return nil }` guard inside the gate is now reachable only via `flags.asJSON` (the `--json --quiet` combo). Pre-existing behavior; fine to leave a docs follow-up for the broader output-pipeline cleanup that handles bug #3 of #918.

PR #970, opened 2026-05-10
fix(cli): allow runtime override of OAuth2 OIDC URLs

### LINKED ISSUE #952: OAuth2 templates: hardcoded localhost:9001 OIDC URLs ignore cfg.BaseURL (every generated CLI affected)
Spotted during Greptile review of [printing-press-library#396](https://github.com/mvanhorn/printing-press-library/pull/396) (etherpad CLI). Three real defects in the OAuth2 templates printing-press v4.2.2 emits — they affect **every CLI** the generator produces, not just etherpad:

## 1. Hardcoded OIDC URLs (P1)

In `internal/cli/auth.go` (login command):

```go
// line ~70 — authorization URL
authURL := "http://localhost:9001/oidc/auth"

// line ~129 — token-exchange URL after callback
tokenURL := "http://localhost:9001/oidc/token"
```

And in `internal/client/client.go` (refresh path):

```go
// line ~375
tokenURL := "http://localhost:9001/oidc/token"
```

All three ignore `cfg.BaseURL`. Any user pointing the CLI at a non-local API gets `auth login` connecting to `localhost:9001` and producing a confusing connection-refused error. Token refresh is broken in the same way.

## 2. Dead `if tokenURL == ""` guard (P1)

Same `client.go` block:

```go
tokenURL := "http://localhost:9001/oidc/token"
if tokenURL == "" {  // always false — literal can never be empty
    return nil
}
```

Looks like a leftover from a planned OIDC-discovery lookup (or `cfg.TokenURL`) that never landed. The guard is unreachable and the URL is irrevocably hardcoded.

## 3. `ExpiresIn==0` storage (P2)

In `auth.go` (login):

```go
expiry := time.Now().Add(time.Duration(tokenResp.ExpiresIn) * time.Second)
```

If the OIDC server returns `expires_in: 0` (non-conformant but possible), `expiry == time.Now()` (already past) — `authHeader()` immediately treats the token as expired and triggers `refreshAccessToken()` per call. Note `client.go`'s `refreshAccessToken()` already has the right guard:

```go
expiry := time.Time{}
if tokenResp.ExpiresIn > 0 {
    expiry = time.Now().Add(...)
}
```

`auth.go`'s login path should match.

## Suggested fix in templates

Add a helper on `Config` that derives OIDC URLs from `BaseURL` with an explicit `OIDCIssuer` override for split-host deployments:

```go
type Config struct {
    BaseURL    string `toml:"base_url"`
    OIDCIssuer string `toml:"oidc_issuer,omitempty"` // override
    // ...
}

func (c *Config) oidcBase() string {
    if c.OIDCIssuer != "" {
        return strings.TrimRight(c.OIDCIssuer, "/")
    }
    if u, err := url.Parse(c.BaseURL); err == nil && u.Scheme != "" && u.Host != "" {
        return u.Scheme + "://" + u.Host
    }
    return ""
}

func (c *Config) AuthURL() string {
    if base := c.oidcBase(); base != "" {
        return base + "/oidc/auth"
    }
    return "http://localhost:9001/oidc/auth"  // dev fallback
}

func (c *Config) TokenURL() string {
    if base := c.oidcBase(); base != "" {
        return base + "/oidc/token"
    }
    return "http://localhost:9001/oidc/token"
}
```

Then auth.go and client.go both call `cfg.AuthURL()` / `cfg.TokenURL()`, and the dead guard is replaced naturally.

## Workaround until templates change

Patched in PR #396 via `.printing-press-patches.json` (same mechanism notion uses for hand-fixes). Diff is small and self-contained — happy to upstream as a PR if you'd find that useful, just let me know which template file emits the OAuth2 paths and I'll match the style. The patch doc is at https://github.com/JohnMcLear/printing-press-library/blob/feat/etherpad-cli/library/productivity/etherpad/.printing-press-patches.json for reference.

tmchow:
## Intent

Closes #952.

OAuth2 templates baked AuthorizationURL/TokenURL into auth.go and client.go as string literals, so a printed CLI pointed at a non-default deployment had no way to override the OIDC endpoints without regenerating. For etherpad-style specs that declared \`http://localhost:9001\` as the default, \`auth login\` and token refresh dialed localhost:9001 even when the user pointed \`--base-url\` at a remote host. The dead \`if tokenURL == \"\"\` guard reported alongside the URL bug also disappears once the URL becomes runtime-overridable. Fold in the matching \`expires_in: 0\` guard reported as item 3 in the same issue, since the same template files own all three.

## Approach

- Emit \`AuthorizationURL\` and \`TokenURL\` as Config fields (\`omitempty\`) when the spec declares them. Same shape as \`BaseURL\`'s existing override.
- Read \`{{ENV}}_AUTHORIZATION_URL\` / \`{{ENV}}_TOKEN_URL\` env vars at \`config.Load()\` time alongside \`{{ENV}}_BASE_URL\`.
- Templates prefer \`cfg.AuthorizationURL\` / \`cfg.TokenURL\`, falling back to the spec literal so existing behavior is preserved when no override is set. The previously dead \`if tokenURL == \"\"\` guard becomes meaningful again as the final fallback.
- Auth-login expiry calc in \`auth.go\` and \`auth_client_credentials.go\` now matches \`refreshAccessToken\`'s \`if ExpiresIn > 0\` pattern: a non-conformant server returning \`expires_in: 0\` no longer marks every subsequent request as expired.

## Scope

Primary area: generator templates (\`internal/generator/templates/{auth,auth_client_credentials,client,config}.go.tmpl\`).

Why this belongs in this repo: every printed CLI with OAuth2 auth flows through these templates. Fixing in the printed CLI would compound only for that one CLI; the templates own this contract.

## Risk

- Generator output drift: only the OAuth2-cc golden changed (auth.go, client.go, config.go). Updated in this PR.
- New Config fields are \`omitempty\`, so existing config files don't change shape.
- Bearer/api_key/cookie/composed specs are unaffected: the new fields are gated on \`.Auth.AuthorizationURL\` / \`.Auth.TokenURL\` being declared in the spec, and \`client.go.tmpl\`'s \`refreshAccessToken\` no longer references \`c.Config.TokenURL\` for non-OAuth2 specs (gated by \`{{- if .Auth.TokenURL}}\`).
- New regression test (\`auth_oauth_url_override_test.go\`) pins all three behaviors plus the bearer-token negative case.

## Output Contract

Generated CLIs gain two optional Config fields and two new env-var hooks. Existing fields, flags, and commands are unchanged. Golden updated for \`generate-golden-api-oauth2-cc\`.

## Verification

- \`go build -o ./printing-press ./cmd/printing-press\` (PASS)
- \`go fmt ./... && go vet ./...\` (PASS, no issues)
- \`go test ./...\` (3438 passed)
- \`golangci-lint run ./...\` (no issues)
- \`scripts/golden.sh verify\` (16 cases passed)
- \`git diff --check\` (clean)

## AI / Automation Disclosure

### DIFF
--- a/internal/generator/auth_oauth_url_override_test.go
+++ b/internal/generator/auth_oauth_url_override_test.go
@@ -63,7 +63,7 @@ func TestOAuth2URLs_RuntimeOverrideEmittedForAuthCodeGrant(t *testing.T) {
 	clientSrc, err := os.ReadFile(filepath.Join(outputDir, "internal", "client", "client.go"))
 	require.NoError(t, err)
 	client := string(clientSrc)
-	require.Contains(t, client, "tokenURL = c.Config.TokenURL",
+	require.Contains(t, client, "tokenURL := c.Config.TokenURL",
 		"refreshAccessToken must read c.Config.TokenURL before falling back to spec default")
 }
 
@@ -103,6 +103,19 @@ func TestOAuth2URLs_RuntimeOverrideEmittedForClientCredentialsGrant(t *testing.T
 	client := string(clientSrc)
 	require.Contains(t, client, "tokenURL = c.Config.TokenURL",
 		"mintClientCredentials must read c.Config.TokenURL via the c.Config != nil guard")
+
+	// Pin the auto-refresh expiry guard: without it a server returning
+	// expires_in: 0 makes every request re-trigger mintClientCredentials in
+	// a loop. mintClientCredentials is the runtime auto-refresh path called
+	// from authHeader(), so this guard matters more than the login-path one.
+	mintIdx := strings.Index(client, "func (c *Client) mintClientCredentials")
+	require.NotEqual(t, -1, mintIdx, "mintClientCredentials must be emitted")
+	mintBody := client[mintIdx:]
+	if next := strings.Index(mintBody[1:], "\nfunc "); next != -1 {
+		mintBody = mintBody[:next+1]
+	}
+	require.Contains(t, mintBody, "if tokenResp.ExpiresIn > 0 {",
+		"mintClientCredentials must guard expiry calc against ExpiresIn==0")
 }
 
 // Pins that bearer_token / api_key specs (no OAuth2 URLs) still build cleanly:
--- a/internal/generator/auth_oauth_url_override_test.go
+++ b/internal/generator/auth_oauth_url_override_test.go
@@ -1,0 +1,136 @@
+package generator
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/mvanhorn/cli-printing-press/v4/internal/spec"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// Pins the runtime override path for the OAuth2 OIDC URLs reported in #952.
+// The generator-baked URL was treated as the only source of truth, leaving
+// users no way to point a printed CLI at a non-default deployment without
+// regenerating. The fix: emit AuthorizationURL/TokenURL as Config fields
+// (with env-var overrides) and prefer cfg.* over the spec literal.
+func TestOAuth2URLs_RuntimeOverrideEmittedForAuthCodeGrant(t *testing.T) {
+	t.Parallel()
+
+	apiSpec := minimalSpec("oauth-url-override")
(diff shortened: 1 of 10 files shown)


tmchow:
@greptileai code review

tmchow:
@greptileai code review

PR #973, opened 2026-05-10
fix(skills): preflight Go toolchain before generation runs

### LINKED ISSUE #793: skill: preflight should check for Go toolchain, not just the printing-press binary
## Summary

`/printing-press` preflight checks for the `printing-press` binary but does NOT check that the Go toolchain is on PATH. Generation runs to completion, then fails ~5+ minutes later at the `go mod tidy` quality gate with `exec: "go": executable file not found in $PATH`. This is a deterministic environment requirement that's checked late.

## Evidence (from McDonald's run, 2026-05-09)

User installed the binary via the documented `go install` happy path. Their interactive shell PATH had `~/go/bin/printing-press` but not `go` itself (WSL setup quirk; common on machines where the binary is invoked directly). Preflight passed cleanly. Phase 0 → 1.5 → 1.7 → 1.9 walked normally. After Phase Gate 1.5 user approval, `printing-press generate --spec ... --validate` produced 55 files / ~10K LOC of Go scaffolding. Then:

```
FAIL go mod tidy
Error: validating generated project: gate "go mod tidy" failed: exec: "go": executable file not found in $PATH
```

Total wall-clock cost: ~15 minutes of agent + user time before the deterministic environment failure surfaced. Build log: archived in run manuscripts.

## Why fail-fast, not fail-late

The setup contract already runs `command -v go >/dev/null 2>&1` once — but only inside the binary-not-found error fallback (where it picks between two install instructions). When the binary IS present and Go is NOT, that check never runs, and the failure punts to the post-generation validation gate.

Fail-fast preflight converts a 15-minute "generate then fail at validate" into a 30-second "abort with actionable message." The check costs nothing at runtime when Go *is* installed (single `command -v` call).

## Proposed fix

Add an unconditional Go-toolchain check to the preflight setup contract in `skills/printing-press/SKILL.md`, alongside the existing `printing-press` binary check. When `command -v go` fails, emit a `[setup-error]` block with the install URL and the minimum Go version the generator targets, then exit. Reuse the existing `references/setup-checks.md` handling that already processes `[setup-error]` for the binary-missing case.

Keep the existing in-fallback `command -v go` reference intact (it still selects between two install messages when both binary and Go are missing). The new unconditional check supersedes it for the binary-present case.

## Acceptance criteria

- **Positive (Go installed):** preflight runs unchanged. No new prompts, no measurable latency increase. Generation proceeds.
- **Negative (binary present, Go absent):** preflight prints `[setup-error] Go toolchain not found.` block with https://go.dev/dl/ and the minimum supported version, then aborts before any `AskUserQuestion`.
- **Negative (binary absent, Go absent):** existing binary-install error path is preserved. The new check fires only when the binary is present.

## Scope boundary

- Does NOT include checking the Go *version* (presence only).
- Does NOT include auto-install.
- Does NOT touch the binary's own version-throttle logic.

## Component

`skill` — fix lives in `skills/printing-press/SKILL.md` (preflight setup contract block) and `skills/printing-press/references/setup-checks.md` (already handles `[setup-error]`).

## Complexity

small (~10 lines of bash, plus minor reference doc edit).

## Step G case-against (recorded for posterity)

A maintainer could close this as "users running a Go-toolchain-dependent code generator should have Go installed; the README documents the prerequisite." That argument weakens against the actual failure mode: the binary-only install path produces `~/go/bin/printing-press` which the user invokes directly without keeping `go` itself on their interactive PATH. The README documents the install command, not "Go must remain on PATH after install." Fail-fast preflight is the right place to catch the post-install drift.

## Related issues

- #781 — Windows + spec-handling issues from notion/youtube builds (related-area: both touch Phase 0 / preflight territory; different specific gap)

---

Filed by `/printing-press-retro` after a McDonald's run that landed on HOLD due to this finding. Retro doc with full triage rationale (drops, skips, what went right) saved at `~/printing-press/manuscripts/mcdonalds/20260509-113558/proofs/20260509-115500-retro-mcdonalds-pp-cli.md` and `/tmp/printing-press/retro/20260509-115500-mcdonalds-retro.md`.

tmchow:
## Intent

Fail-fast on a missing Go toolchain in `/printing-press` preflight, before generation produces ~10K LOC of scaffolding that the post-generation `go mod tidy` quality gate cannot validate.

Issue: Closes #793

## Approach

The setup contract already runs `command -v go` once, but only inside the binary-not-found error fallback to pick between two install messages. When the binary IS present and `go` is NOT, that check never runs, and the failure punts to the post-generation `go mod tidy` gate ~5+ minutes later.

This PR adds an unconditional `command -v go` check directly after binary detection in [skills/printing-press/SKILL.md](skills/printing-press/SKILL.md). When `go` is missing, the contract emits a `[setup-error] Go toolchain not found.` block (with the install URL and minimum supported version), then exits non-zero. The existing `[setup-error]` handler in [references/setup-checks.md](skills/printing-press/references/setup-checks.md) section 1 already stops the skill and surfaces the printed message verbatim; that section is generalized to mention either missing prerequisite (binary or Go), since the same handler now covers both.

Locked in with a new contract test, `TestPrintingPressSkillPreflightChecksGoToolchain` in [internal/pipeline/contracts_test.go](internal/pipeline/contracts_test.go), that asserts the unconditional `command -v go` line, the `[setup-error]` marker, and the install URL all live inside the contract block.

## Scope

Primary area: skills (preflight setup contract)

Why this belongs in this repo: The setup contract is a checked-in skill artifact. The fix is a single-skill change to the main `printing-press` SKILL.md; the other skills (`-score`, `-catalog`, `-publish`) don't run `go mod tidy` after their commands and are intentionally untouched.

## Risk

Low. No generator code, template, or generated-output change. Only the preflight bash block runs differently — the new `command -v go` check fires once per `/printing-press` invocation. When Go is present, behavior is unchanged. When Go is missing, the user sees the new `[setup-error]` block and the skill stops before any `AskUserQuestion`.

The new check fires unconditionally, including the local-build path. That path requires Go to have been installed at build time, but a stale `./printing-press` binary plus a since-removed Go install would still be caught — strictly an improvement over the current behavior.

## Output Contract

N/A. No generator templates, manifests, command output, MCP schemas, scorecard output, catalog rendering, or pipeline artifacts changed.

## Verification

- `go test ./...` — 3451 passed in 34 packages.
- `go vet ./...` — No issues.
- `golangci-lint run ./...` — No issues.
- `go fmt ./...` — clean.
- `go build -o ./printing-press ./cmd/printing-press` — built.
- `scripts/golden.sh verify` — 16 cases passed.
- `git diff --check` — clean.

## AI / Automation Disclosure

### DIFF
--- a/internal/pipeline/contracts_test.go
+++ b/internal/pipeline/contracts_test.go
@@ -96,6 +96,20 @@ func TestPrintingPressSkillUsesRunRootStateFile(t *testing.T) {
 	assert.Contains(t, skill, `"working_dir": "$CLI_WORK_DIR"`)
 }
 
+func TestPrintingPressSkillPreflightChecksGoToolchain(t *testing.T) {
+	skillPath := filepath.Join("..", "..", "skills", "printing-press", "SKILL.md")
+	full := readContractFile(t, skillPath)
+	block := extractContractBlock(t, full)
+
+	// Unconditional Go-toolchain presence check fires regardless of whether the
+	// printing-press binary is found, so binary-present + Go-absent fails fast
+	// instead of crashing 5+ minutes later in the post-generation `go mod tidy`
+	// quality gate.
+	assert.Contains(t, block, `if ! command -v go >/dev/null 2>&1; then`)
+	assert.Contains(t, block, `[setup-error] Go toolchain not found.`)
+	assert.Contains(t, block, `https://go.dev/dl/`)
+}
+
 func TestPrintingPressSkillUsesRunstateForBuilds(t *testing.T) {
 	skill := readContractFile(t, filepath.Join("..", "..", "skills", "printing-press", "SKILL.md"))
 
--- a/internal/pipeline/contracts_test.go
+++ b/internal/pipeline/contracts_test.go
@@ -101,10 +101,10 @@ func TestPrintingPressSkillPreflightChecksGoToolchain(t *testing.T) {
 	full := readContractFile(t, skillPath)
 	block := extractContractBlock(t, full)
 
-	// Unconditional Go-toolchain presence check fires regardless of whether the
-	// printing-press binary is found, so binary-present + Go-absent fails fast
-	// instead of crashing 5+ minutes later in the post-generation `go mod tidy`
-	// quality gate.
+	// The Go-toolchain presence check fires after the binary detection block
+	// exits cleanly (binary found or PATH-augmented). It catches binary-present
+	// + Go-absent and fails fast instead of crashing 5+ minutes later in the
+	// post-generation `go mod tidy` quality gate.
 	assert.Contains(t, block, `if ! command -v go >/dev/null 2>&1; then`)
 	assert.Contains(t, block, `[setup-error] Go toolchain not found.`)
 	assert.Contains(t, block, `https://go.dev/dl/`)
--- a/skills/printing-press/SKILL.md
+++ b/skills/printing-press/SKILL.md
@@ -160,6 +160,23 @@ elif ! command -v printing-press >/dev/null 2>&1; then
   fi
 fi
 
+# Verify the Go toolchain is on PATH. Generation runs Go-based quality gates
+# (go mod tidy, go vet, etc.) after writing thousands of lines of scaffolding,
+# so a missing `go` only surfaces 5+ minutes in. Fail-fast costs one command -v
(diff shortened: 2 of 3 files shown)


tmchow:
@greptileai code review

tmchow:
@greptileai code review

PR #984, opened 2026-05-10
fix(cli): reject reserved placeholder hosts in spec validation

### LINKED ISSUE #818: spec-parser: reject or stub-disclose unresolved browser-sniff URL placeholders (no example.com leak)
## Problem
`ads detail` for the OLX CLI failed at runtime with `GET https://example.com/resource` — the literal placeholder string from the sniff-spec template was compiled into the generated client. The endpoint extracted by browser-sniff didn't have a derivable path pattern, so the spec emitter fell back to a placeholder URL that compiled fine but blew up at first call.

The placeholder is silent at generate time, silent at build time, and only fails at runtime — exactly the kind of latent bug that erodes trust in generated output.

## Root cause
`internal/spec/` (browser-sniff post-processing). When the sniff process can't determine a real path/template for a discovered endpoint (likely because `__NEXT_DATA__` extraction returned an envelope without a per-record URL key), the emitter falls back to `https://example.com/resource` rather than (a) refusing to emit the endpoint, (b) marking it as `stub: true` so the generator emits a "not yet wired" command, or (c) flagging it in research output for agent review.

## Durable fix
In the spec emitter, when a discovered endpoint's path template can't be resolved, pick exactly one path and document it:

- (a) drop the endpoint and note it in research.json's `gaps` list, OR
- (b) emit the endpoint with `stub: true` annotation; generator template renders the resulting command as a stub that prints "not yet wired" when invoked (using the existing `cliutil.IsVerifyEnv()` short-circuit pattern); research.json's gaps captures it for SKILL/README disclosure.

Reject the placeholder URL outright either way.

## Acceptance criteria
- positive: when an endpoint's URL template can't be resolved during sniff, the emitter either (a) drops the endpoint and notes it in research.json's gaps, or (b) emits the endpoint with `stub: true` so the generator scaffolds a stub command with honest "not yet wired" messaging
- negative: an endpoint with a clear URL template still emits normally
- negative: no compiled URL string in any generated client matches `example\.com`

## Scope boundary
Pick (a) or (b), not both. Document the choice. Don't try to invent URL templates the sniff can't derive.

## Evidence
OLX run 20260509-103549. Phase 5 dogfood loop 2: `ads detail` failed with `Original error: GET https://example.com/resource`. The literal string was present in the sniff-spec.yaml shipped under `research/olx-sniff-spec.yaml` and propagated through the generator into runtime client code. Patched at the printed-CLI level by adding a `--url` normalizer with an example.com guard, but the underlying spec-emitter bug remains.

---

**Artifacts** *(generated by /printing-press-retro for run olx 20260509-103549)*
- Retro doc: https://files.catbox.moe/kpuea5.md
- Manuscripts: https://files.catbox.moe/r3e6y9.zip
- CLI source: https://files.catbox.moe/djevii.zip

tmchow:
## Intent

Issue: Closes #818

Browser-sniff emitters and LLM-generated YAML can ship endpoint URLs like `https://example.com/resource` as real paths. The placeholder compiles cleanly into the runtime client and only fails at first live call (OLX retro evidence in the issue body). A defense-in-depth check at the spec layer makes the leak fail loudly at `spec.ParseBytes` regardless of the emitter that produced it.

## Approach

`Validate()` now rejects absolute URL fields on resources, sub-resources, and endpoints whose bare host is one of the RFC 2606 / RFC 6761 reserved documentation names (`example.com`, `example.org`, `example.net`, `example.test`, `example.invalid`, `example`). Subdomained forms like `api.example.com` and `geocoding-api.example.com` keep passing so existing fixtures and the openapi/docspec `api.example.com` base_url fallback are unaffected. The check runs uniformly over every URL-bearing field — resource `base_url`, endpoint `path`, endpoint `base_url`, sub-resource `base_url`, sub-endpoint `path`, sub-endpoint `base_url` — so the same leak from any of those sources is caught.

## Scope

Primary area: `internal/spec/`

Why this belongs in this repo: the bug surfaced in a printed CLI (OLX, runtime `GET https://example.com/resource`), but the placeholder originated in the spec layer and would otherwise recur on every browser-sniff CLI built off a similar template. Rejecting it at spec-parse time prevents the leak across every future printed CLI rather than patching each one.

## Risk

Low. Spec-level only. The check is keyed on the bare reserved hostnames, so:

- Legitimate test fixtures and parser fallbacks using `api.example.com` (subdomained) continue to pass.
- Relative paths (the overwhelming common case) continue to pass.
- Paths whose query strings contain `example.com` (e.g. `/proxy?url=https://example.com/x`) continue to pass because the URL is not absolute at the path level.

Acceptance-criteria mapping:

- Positive: a placeholder URL fails validation with an actionable error pointing at the offending field and host. Spec authors choose drop-and-document or annotate-as-stub before re-running generate.
- Negative 1: clear URL templates emit normally — no behavior change.
- Negative 2: no compiled URL in any generated client matches the bare reserved hosts, because validation halts generation before the client is emitted.

## Output Contract

N/A. No generator-output diffs.

## Verification

- `go test ./internal/spec/...` — 270 passed including 10 new cases in `TestValidateRejectsReservedPlaceholderHost`
- `go test ./internal/openapi/... ./internal/docspec/... ./internal/graphql/... ./internal/browsersniff/...` — 419 passed
- `go vet ./...` — clean
- `go fmt ./...` — clean
- `golangci-lint run ./internal/spec/...` — no issues
- `scripts/golden.sh verify` — 16 cases passed
- `git diff --check` — clean
- Full `go test ./...` was blocked locally by ENOSPC pressure on the host (`darwin_arm64/link: mapping output file failed: no space left on device`) on the verify-skill and generator suites; that is a host-disk issue, not a regression — the affected tests build a `printing-press` binary into TempDir and the linker cannot mmap the output. CI runs on a clean disk and will exercise the full suite.

## AI / Automation Disclosure

### DIFF
--- a/internal/cli/public_param_audit_test.go
+++ b/internal/cli/public_param_audit_test.go
@@ -18,7 +18,7 @@ func TestPublicParamAuditJSONInventoriesDecisionRequiredParams(t *testing.T) {
 name: public-param-test
 description: Public param test
 version: "0.1.0"
-base_url: https://example.test
+base_url: https://api.example.com
 auth:
   type: none
 resources:
@@ -67,7 +67,7 @@ func TestPublicParamAuditStrictFailsOnlyForUnreviewedCandidates(t *testing.T) {
 name: public-param-test
 description: Public param test
 version: "0.1.0"
-base_url: https://example.test
+base_url: https://api.example.com
 auth:
   type: none
 resources:
@@ -124,7 +124,7 @@ func TestPublicParamAuditStrictPassesWhenFlagNameIsInSpec(t *testing.T) {
 name: public-param-test
 description: Public param test
 version: "0.1.0"
-base_url: https://example.test
+base_url: https://api.example.com
 auth:
   type: none
 resources:
--- a/internal/openapi/parser_test.go
+++ b/internal/openapi/parser_test.go
@@ -745,7 +745,7 @@ info:
   title: Hybrid
   version: "1.0"
 servers:
-  - url: https://example.com
+  - url: https://api.example.com
 components:
   securitySchemes:
     OAuth2:
@@ -791,7 +791,7 @@ info:
   title: Malformed
   version: "1.0"
 servers:
-  - url: https://example.com
+  - url: https://api.example.com
 components:
   securitySchemes:
     OAuth2:
@@ -830,7 +830,7 @@ info:
   title: Sentry
   version: "1.0"
 servers:
-  - url: https://example.com
+  - url: https://api.example.com
 components:
   securitySchemes:
     auth_token:
--- a/internal/spec/spec.go
+++ b/internal/spec/spec.go
@@ -1548,6 +1548,9 @@ func (s *APISpec) Validate() error {
 	if s.BaseURL == "" && s.BasePath == "" {
 		return fmt.Errorf("base_url is required")
 	}
+	if err := validateReservedPlaceholderHost("base_url", s.BaseURL); err != nil {
+		return err
+	}
 	if len(s.Resources) == 0 {
 		return fmt.Errorf("at least one resource is required")
 	}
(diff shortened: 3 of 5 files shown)


tmchow:
@greptileai code review

tmchow:
@greptileai code review

Issue #977, opened 2026-05-10
[P1] WU-1: oneOf/anyOf body fallback to --body-json flag


paonessa:
**Source:** `/printing-press-retro` of cloudflare-pp-cli (run 20260510-153824).

## Summary

Generation emits ~26 warnings of the form `warning: skipping request body for POST "/zones/{zone_id}/dns_records": contains oneOf/anyOf` and silently emits the affected commands with no body parameter. Users cannot invoke these endpoints from the generated CLI — `dns records create` builds and runs but the API receives an empty body and rejects the call. There is no `--body-json` escape hatch and no `--help` documentation hinting at the missing body.

## Where to look

- **Component:** generator
- **Likely area:** `internal/generator/templates/` — the endpoint-body template iteration that emits typed body fields as named flags.

## Concrete cross-API evidence

- **Cloudflare** (this run): 26 endpoints with `oneOf`/`anyOf` body — DNS records (`A | AAAA | CNAME | MX | ...`), Access apps (`SaaS | Self-hosted | Browser | Infrastructure`), ruleset rules, zone settings, 8 of 13 Browser Rendering endpoints.
- **Stripe** (catalog): PaymentMethod and Source are `oneOf` discriminated unions.
- **Linear** (catalog): issues use `anyOf{null, T}` for many optional fields.
- **Notion** (catalog): block types are `oneOf` per block kind.

## Durable fix (proposed)

Generator template change. When `endpoint.Body.Schema.OneOf != nil || endpoint.Body.Schema.AnyOf != nil`:

1. Skip the field-by-field flag plumbing.
2. Emit a single `--body-json string` flag.
3. In the command's RunE, JSON-decode the flag value and send it verbatim as the request body.
4. Update help text: `complex schema; use --body-json with the API's JSON shape`.

This is the minimum-viable fallback. A follow-up could parse the discriminator and emit `--type X` plus per-branch named flags, but the bare `--body-json` flag alone unblocks the entire endpoint class.

## Acceptance criteria

- **Positive:** Regenerating cloudflare-pp-cli from a curated Cloudflare spec, `dns records create --help` shows a `--body-json string` flag. `dns records create --body-json '{"type":"A","name":"x.example.com","content":"1.2.3.4","ttl":300}' --dry-run` shows the JSON body in the dry-run preview.
- **Negative:** Specs with simple typed bodies (no `oneOf`/`anyOf`) continue to emit named flags as today — no regression.

## Scope boundary

[...]

PR #994, opened 2026-05-10
fix(cli): emit --body-json fallback for oneOf/anyOf request bodies

tmchow:
## Summary

- OpenAPI endpoints whose request body schema is a `oneOf`/`anyOf` (Cloudflare DNS records, Stripe PaymentMethod, Notion blocks, Linear filters, etc.) previously parsed with a `skipping request body` warning and shipped with **no body parameter** — the generated `create`/`update` commands built a request, sent an empty body, and the API rejected the call.
- The parser now flips a new `Endpoint.BodyJSONFallback` field for these endpoints. Command templates respond by emitting a single `--body-json string` flag that accepts the JSON object verbatim. The MCP surface exposes a matching `body_json` input so agent-native parity is preserved.
- Scope per the issue: minimum-viable opaque fallback. Discriminator-aware per-branch flags and schema-side validation are out of scope (called out as follow-ups in [#977](https://github.com/mvanhorn/cli-printing-press/issues/977)).

## Approach

- `spec.Endpoint` gets `BodyJSONFallback bool` (yaml/json tags, omitempty so existing internal specs round-trip unchanged).
- `internal/openapi/parser.go`: `mapRequestBody` returns a third bool. When `collectAllOfProperties` reports a `oneOf`/`anyOf` at the body root, it now returns `nil, "application/json", true` and downgrades the warning to "emitting --body-json fallback".
- Generator helpers (`bodyVarDecls`, `bodyFlagRegs`, `bodyRequiredChecks`) early-return the fallback shape when the flag is set. A new `bodyMapForEndpoint` template func dispatches between the typed-flag renderer and a new `bodyJSONFallbackMap` that emits the JSON-decode + map-shape check.
- `command_endpoint.go.tmpl` and `command_promoted.go.tmpl` call `bodyMapForEndpoint .Endpoint` in place of `bodyMap .Endpoint.Body`; the existing `body = map[string]any{}` line stays so the fallback only overwrites it when the user passed a value.
- MCP: `mcpParamBindings` emits a single `body_json` binding with `Location: "body_json"`. `mcp_tools.go.tmpl` adds a `mcplib.WithString("body_json", …)` input on these tools and a guarded `bodyJSONOverride json.RawMessage` variable in `makeAPIHandler` that replaces the marshaled `bodyArgs` map for POST/PUT/PATCH. The new variable is gated behind a `hasBodyJSONFallback` template predicate so it only appears in generated MCP code when at least one endpoint needs it.

## Scope

Primary area: generator (OpenAPI parser + command templates + MCP surface).

This is a machine change, not a printed-CLI patch: the bug surfaced across multiple unrelated APIs (Cloudflare, Stripe, Notion, Linear per the issue) and the right fix is the generator, not 26 hand-edits in each affected CLI.

## Risk

- The `endpoint.BodyJSONFallback` field defaults to `false`, so existing internal YAML specs and OpenAPI specs without `oneOf`/`anyOf` bodies generate the same code they did before (verified via the full golden harness).
- Fallback path intentionally restricts `--body-json` payloads to JSON objects. Top-level discriminated unions in real-world specs (every example in #977) are object-shaped; the rare array-typed body case surfaces as a clear `--body-json must be a JSON object` error rather than a silent send.
- MCP `body_json` input is unannotated for read-only-ness because POST/PUT/PATCH are inherently mutating, matching the existing typed-body MCP tools.

## Output Contract

- Generator output changes for any OpenAPI spec with a `oneOf`/`anyOf` body: a `var flagBodyJSON string` declaration and a `cmd.Flags().StringVar(&flagBodyJSON, "body-json", ...)` registration appear on the affected command; the typed body-map block is replaced with a `flagBodyJSON` parse-and-assign block.
- New `pp:body-json-fallback` behavior is implicit (no annotation); the public surface change is the `--body-json` flag itself.
- Golden harness fixtures don't currently include a `oneOf`/`anyOf` body case, so no existing golden needs updating. New parser tests pin the parser-level behavior; new helper tests pin the generator helpers.

## Verification

- `go test ./...` → 3479 passed in 34 packages.
- `go vet ./...` → clean.
- `golangci-lint run ./...` → clean.
- `scripts/golden.sh verify` → 16/16 pass (no existing golden case has a `oneOf`/`anyOf` body, so output is unchanged for non-fallback endpoints).
- End-to-end: generated a CLI from a synthetic OpenAPI spec with a `oneOf` DNS-record body. `records create --help` shows `--body-json` with the new help text; `--body-json '{"type":"A",...}' --dry-run` prints the JSON body in the dry-run preview; `--body-json '["arr"]'` returns `--body-json must be a JSON object`; `--body-json 'not json'` returns a clear parse error.

Refs #977

## AI / Automation Disclosure

### DIFF
--- a/internal/generator/body_map_test.go
+++ b/internal/generator/body_map_test.go
@@ -456,16 +456,18 @@ func TestBodyJSONFallback_VarDecls(t *testing.T) {
 }
 
 // TestBodyJSONFallback_FlagRegs registers a single --body-json flag with
-// a help string mentioning oneOf/anyOf so users know why the fallback
-// shape is in play.
+// user-facing help that also names oneOf/anyOf for spec-aware readers.
 func TestBodyJSONFallback_FlagRegs(t *testing.T) {
 	t.Parallel()
 	got := bodyFlagRegs(spec.Endpoint{BodyJSONFallback: true})
 	if !strings.Contains(got, `cmd.Flags().StringVar(&flagBodyJSON, "body-json"`) {
 		t.Errorf("expected --body-json flag registration, got:\n%s", got)
 	}
+	if !strings.Contains(got, "polymorphic schema") {
+		t.Errorf("expected user-facing help text, got:\n%s", got)
+	}
 	if !strings.Contains(got, "oneOf/anyOf") {
-		t.Errorf("expected help text to mention oneOf/anyOf, got:\n%s", got)
+		t.Errorf("expected spec-aware hint mentioning oneOf/anyOf, got:\n%s", got)
 	}
 }
 
@@ -494,7 +496,7 @@ func TestBodyJSONFallback_BodyMap(t *testing.T) {
 		`json.Unmarshal([]byte(flagBodyJSON), &parsedBodyJSON)`,
 		`asMap, ok := parsedBodyJSON.(map[string]any)`,
 		`body = asMap`,
-		`--body-json must be a JSON object`,
+		`--body-json must be a JSON object, got JSON %T`,
 	}
 	for _, s := range wantSubstrings {
 		if !strings.Contains(got, s) {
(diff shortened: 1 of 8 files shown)


tmchow:
@greptileai code review

tmchow:
@mergifyio queue

PR #995, opened 2026-05-11
fix(cli): fold per-spec base URL path prefixes on multi-spec merge

### LINKED ISSUE #831: generator: multi-spec generation drops base-URL path prefix when input specs declare different servers[].url paths

tmchow:
## Intent

Closes #831 — when `printing-press generate --spec a.yaml --spec b.yaml` merged specs whose `servers[0].url` shared a scheme+host but disagreed on path component, the merge kept only `specs[0].BaseURL` and silently dropped every later spec's path prefix. Concrete case from the retro: Atlassian Cloud combo CLI where Jira lives at `https://{tenant}` and Confluence at `https://{tenant}/wiki/api/v2`; every Confluence command 404'd at runtime because `client.Get("/pages", ...)` resolved against the bare host without the missing `/wiki/api/v2`.

## Approach

`mergeSpecs` now plans the merged BaseURL across all input specs:

- All specs share full BaseURL → keep `specs[0].BaseURL`, no rewrites (idempotent path).
- All specs share scheme+host, paths diverge → collapse merged BaseURL to the bare host, return each spec's path component for folding into endpoints.
- Hosts disagree (separate multi-host problem, out of scope) → keep the old `specs[0].BaseURL` fallback.

When rewriting fires, `prefixResourceEndpointPaths` walks each affected spec's resources (and sub-resources) and prepends the prefix to every `endpoint.Path`. Endpoints that already declare an absolute `base_url` override are skipped so the per-endpoint override mechanism keeps working unchanged.

A `[multi-spec]` stderr info line announces when prefixing fires, gated by `PRINTING_PRESS_VERIFY` to keep verifier subprocesses quiet.

## Scope

Primary area: `internal/cli/root.go` `mergeSpecs` (multi-spec merge only; single-spec early-return is untouched).

Why this belongs in this repo: this is a generator/merge-time correctness bug. The downstream printed CLI was emitting `client.Get("/pages", ...)` against `https://{tenant}` — the right fix is to prepend `/wiki/api/v2` to `/pages` before code generation runs, not to patch each printed CLI.

## Risk

- Multi-spec runs whose specs share host but diverge on path now emit different `endpoint.Path` values and a different merged `BaseURL`. Single-spec generation is unaffected (`len(specs) == 1` early return).
- Cross-host multi-spec merges (host disagreement) keep the prior fallback behavior so this change does not expand scope to a separate, already-broken case.
- Endpoint `base_url` overrides remain authoritative — the rewrite skips endpoints with non-empty `BaseURL` to avoid double-resolving against the resource/endpoint override.

## Output Contract

N/A — golden harness (`scripts/golden.sh verify`) passes unchanged; the only generated-output diff would surface in a future multi-spec fixture, and the change is gated on multi-spec input.

## Verification

- `go fmt ./...`
- `go vet ./...`
- `go build -o ./printing-press ./cmd/printing-press`
- `go test ./...` — 3491 passed in 34 packages
- `scripts/golden.sh verify` — 16 cases pass
- `rtk git diff --check` — clean

New tests:
- `TestMergeSpecsRewritesEndpointPathsForDivergentBaseURLPathPrefixes` — positive case (Atlassian-shape repro)
- `TestMergeSpecsLeavesPathsUnchangedWhenBaseURLsAgree` — idempotent case (matched prefixes, no double-prefixing)
- `TestMergeSpecsLeavesPathsUnchangedWhenHostsDiffer` — out-of-scope case (different hosts, prior fallback preserved)

## AI / Automation Disclosure

### DIFF
--- a/internal/cli/generate_test.go
+++ b/internal/cli/generate_test.go
@@ -948,6 +948,150 @@ func TestMergeSpecsPrefersReplayableBrowserTransportOverUnshippablePageContext(t
 	assert.Equal(t, spec.HTTPTransportBrowserChrome, mergedChrome.HTTPTransport)
 }
 
+// TestMergeSpecsRewritesEndpointPathsForDivergentBaseURLPathPrefixes covers the
+// Atlassian-style case from issue #831: spec A's base URL is the bare host and
+// spec B's base URL adds a path prefix on the same host. Without rewriting,
+// every spec-B command 404s because its path prefix is silently dropped by the
+// merge.
+func TestMergeSpecsRewritesEndpointPathsForDivergentBaseURLPathPrefixes(t *testing.T) {
+	t.Parallel()
+
+	specA := &spec.APISpec{
+		Name:    "a",
+		Version: "0.1.0",
+		BaseURL: "https://example.com",
+		Resources: map[string]spec.Resource{
+			"foo": {
+				Endpoints: map[string]spec.Endpoint{
+					"list": {Method: "GET", Path: "/foo"},
+				},
+			},
+		},
+		Types: map[string]spec.TypeDef{},
+	}
+	specB := &spec.APISpec{
+		Name:    "b",
+		Version: "0.1.0",
+		BaseURL: "https://example.com/api/v2",
+		Resources: map[string]spec.Resource{
+			"bar": {
+				Endpoints: map[string]spec.Endpoint{
+					"list": {Method: "GET", Path: "/bar"},
+					"get":  {Method: "GET", Path: "/bar/{id}"},
+				},
+				SubResources: map[string]spec.Resource{
+					"comments": {
+						Endpoints: map[string]spec.Endpoint{
+							"list": {Method: "GET", Path: "/bar/{id}/comments"},
+						},
+					},
+				},
+			},
+		},
+		Types: map[string]spec.TypeDef{},
+	}
+
+	merged := mergeSpecs([]*spec.APISpec{specA, specB}, "combo")
+
+	// merged BaseURL collapses to the bare host shared by both specs.
+	assert.Equal(t, "https://example.com", merged.BaseURL)
+
+	// spec A's endpoints (no path prefix on its base URL) are unchanged.
+	assert.Equal(t, "/foo", merged.Resources["foo"].Endpoints["list"].Path)
+
+	// spec B's endpoints absorb the dropped "/api/v2" path component.
+	assert.Equal(t, "/api/v2/bar", merged.Resources["bar"].Endpoints["list"].Path)
+	assert.Equal(t, "/api/v2/bar/{id}", merged.Resources["bar"].Endpoints["get"].Path)
+
+	// Sub-resource endpoints are rewritten too.
+	assert.Equal(t,
+		"/api/v2/bar/{id}/comments",
+		merged.Resources["bar"].SubResources["comments"].Endpoints["list"].Path,
+	)
+}
+
+// TestMergeSpecsLeavesPathsUnchangedWhenBaseURLsAgree guards the idempotent
+// case: when every spec's base URL is identical, no path rewriting fires and
+// the merged BaseURL keeps the shared path prefix.
+func TestMergeSpecsLeavesPathsUnchangedWhenBaseURLsAgree(t *testing.T) {
+	t.Parallel()
+
+	specA := &spec.APISpec{
+		Name:    "a",
+		Version: "0.1.0",
+		BaseURL: "https://example.com/api/v2",
+		Resources: map[string]spec.Resource{
+			"foo": {
+				Endpoints: map[string]spec.Endpoint{
+					"list": {Method: "GET", Path: "/foo"},
+				},
+			},
+		},
+		Types: map[string]spec.TypeDef{},
+	}
+	specB := &spec.APISpec{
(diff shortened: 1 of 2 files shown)


tmchow:
@greptileai code review

tmchow:
@greptileai code review

PR #1000, opened 2026-05-11
fix(cli): validate ListIDs resourceType to close SQL injection

brennaman:
## Intent

`Store.ListIDs(resourceType)` in `store.go.tmpl` interpolates the caller-supplied string directly into a `SELECT` — a classic SQL-injection sink. The method is exported, so every printed CLI ships the same flaw.

Issue: N/A (caught by Greptile on mvanhorn/printing-press-library#408, the outlook-calendar PR).

## Approach

Resolve the table name with a parameterized `sqlite_master` lookup, then substitute only the trusted name (double-quoted) into the SELECT. If the lookup misses, fall back to the existing `SELECT id FROM resources WHERE resource_type = ?` path — same semantics as before, just with the unsafe interpolation removed. Callers can now pass any string without risk.

## Scope

Primary area: `internal/generator/templates/store.go.tmpl`.

Why this belongs in this repo: the bug lives in a generator template, so every CLI rendered by the Printing Press carries it. Fixing only the printed CLI would regress on the next regen and would leave every already-printed CLI vulnerable. Per `AGENTS.md` "Default to machine changes" — this is the canonical case.

The same fix has been mirrored into the in-flight outlook-calendar PR (`mvanhorn/printing-press-library#408`) so that PR can ship clean; future regens of any CLI will pick up the template fix automatically.

## Risk

Behavior preserved: callers using existing well-known resource types (`events`, `attachments`, etc.) hit the same code path with the same results. Callers passing unknown names now fall through to the generic `resources` table — same outcome as the original `Query` error fallback. No template variable changes, no schema changes, no MCP-surface changes.

## Output Contract

N/A — generated `store.go` body changes (function internals), but no exported API, MCP shape, or scorecard output is affected. `scripts/golden.sh verify` was run and passes (16 cases).

## Verification

- `go build ./...` — clean.
- `scripts/golden.sh verify` — 16/16 PASS.
- Mirrored patch on `outlook-calendar/internal/store/store.go` in the library PR: `go build ./...`, `go vet ./...`, `go test ./...` all clean.

## AI / Automation Disclosure

### DIFF
--- a/internal/generator/templates/store.go.tmpl
+++ b/internal/generator/templates/store.go.tmpl
@@ -903,11 +903,21 @@ func (s *Store) GetSyncCursor(resourceType string) string {
 
 // ListIDs returns all IDs from a resource's domain table, or from the generic
 // resources table if no domain table exists. Used by dependent sync to iterate parents.
+//
+// resourceType is never interpolated into SQL directly. We resolve it to a real
+// table name via a parameterized sqlite_master lookup; only that trusted name is
+// substituted (double-quoted) into the SELECT. Callers may pass any string.
 func (s *Store) ListIDs(resourceType string) ([]string, error) {
-	// Try domain table first (tables are named after the resource type)
-	query := fmt.Sprintf("SELECT id FROM %s", resourceType)
-	rows, err := s.db.Query(query)
-	if err != nil {
+	var table string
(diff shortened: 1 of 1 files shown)


tmchow:
@mergifyio queue

PR #1018, opened 2026-05-11
fix(cli): bind typed upsert values in column-declaration order

### LINKED ISSUE #1014: Generator template emits FK-bearing upsert bindings in wrong column order
## Summary

The code generator's typed upsert template emits SQL binding arguments in a fixed `(id, data, synced_at, fk)` order, but the SQL column declarations use `(id, <fk>, data, synced_at)`. This causes **silent data corruption** on every typed table with a foreign key column:

- FK columns receive JSON data blobs
- Data columns receive timestamps
- synced_at columns receive FK string values

No runtime error occurs because SQLite's `TEXT NOT NULL` constraint accepts any string value.

## Affected template

`internal/generator/templates/store.go.tmpl` — the typed upsert function generator. Every CLI with FK-bearing typed tables is affected.

## Reproduction

Any OpenAPI spec that produces a typed table with a FK column. For the multimail CLI (47 endpoints, 19 command groups), 8 typed tables were affected: `verify`, `not_spam`, `report_spam`, `reply`, `request_upgrade`, `send`, `threads`, `upgrade`.

Generated code (wrong):
```go
func (s *Store) upsertVerifyTx(tx *sql.Tx, id string, obj map[string]any, data json.RawMessage) error {
    if _, err := tx.Exec(
        `INSERT INTO verify (id, domains_id, data, synced_at)
         VALUES (?, ?, ?, ?)
         ON CONFLICT(id) DO UPDATE SET domains_id = excluded.domains_id, data = excluded.data, synced_at = excluded.synced_at`,
        id,
        string(data),     // ← goes into domains_id column (wrong)
        time.Now(),       // ← goes into data column (wrong)
        lookupFieldValue(obj, "domains_id"), // ← goes into synced_at column (wrong)
    ); err != nil {
```

Correct binding order:
```go
        id,
        lookupFieldValue(obj, "domains_id"), // FK column
        string(data),                         // data column
        time.Now(),                           // synced_at column
```

## Impact

- Every `sync` run silently corrupts all FK-bearing typed tables
- Compound commands querying typed columns (`stale`, `health`, etc.) return wrong values
- Existing corrupted data persists until `sync --reset` or database deletion
- Affects every generated CLI with FK-bearing typed tables, not just multimail

## Secondary audit note: ResolveByName

`store.go.tmpl` also has a `ResolveByName` function that uses `fmt.Sprintf` to interpolate a `field` parameter into a JSON path query. While the callers may currently use safe constants, the same pattern of unvalidated string interpolation into SQL should be audited:

```go
func (s *Store) ResolveByName(resourceType, field, value string) (string, error) {
    query := fmt.Sprintf(`SELECT id FROM resources WHERE resource_type = ? AND json_extract(data, '$.%s') = ?`, field)
```

## Fix applied locally

Fixed in [multimail-dev/multimail-cli@21e3041](https://github.com/multimail-dev/multimail-cli/commit/21e3041) and in [PR #289 on printing-press-library](https://github.com/mvanhorn/printing-press-library/pull/289). The template should be updated to emit bindings matching the SQL column declaration order.

## Discovery

Found by Greptile automated review on PR #289 (mvanhorn/printing-press-library).

tmchow:
## Intent

Issue: Closes #1014

Sub-resource typed-upsert helpers emitted bindings in `(id, data, synced_at, fk)` order while the SQL column list was `(id, fk, data, synced_at)`. Every paginated sync into a sub-resource table silently corrupted rows: JSON blobs landed in the FK column, timestamps landed in `data`, and FK strings landed in `synced_at`. SQLite accepted it because the columns were all TEXT-compatible. Top-level gravity-2+ tables and dependent-sync tables happened to align because their non-base columns were appended after `data`/`synced_at`; the bug only surfaced for sub-resource tables built by `buildSubResourceTable`, which inserts the FK column between `id` and `data`.

## Approach

In `internal/generator/templates/store.go.tmpl`, replace the three hard-coded leading bindings (`id`, `string(data)`, `time.Now()`) plus the filtered loop over the rest with a single loop that walks `.Columns` in declaration order and emits the matching expression per column. The argument order now follows `columnNames` by construction, so adding a column anywhere in the slice cannot desynchronize the bindings again.

## Scope

Primary area: generator template (`store.go.tmpl`).

Why this belongs in this repo: the bug is in a Printing Press template emitted into every printed CLI with FK-bearing typed tables. Fixing the printed CLI alone (as already done in printing-press-library#289) doesn't compound to future generations.

Out of scope: the issue's secondary audit note about `ResolveByName` using `fmt.Sprintf` to interpolate a `field` parameter into a SQL JSON path. Callers currently pass safe constants; tightening that path is a separate change.

## Risk

Generator template change: every CLI regenerated after this lands will get bindings in column-declaration order. For tables where the order already aligned (top-level gravity-2+ tables, dep-sync tables with `parent_id` appended), the emitted Go is functionally identical — only sub-resource and any future mid-list-FK tables change behavior. Existing CLIs with already-corrupted typed columns still need a `sync --reset` (or DB delete) to rebuild; that's downstream work, not part of this PR.

## Output Contract

Generator template diff. Existing golden harness exercises top-level tables only, so no golden fixtures change. Verified with `scripts/golden.sh verify` (16/16 pass).

## Verification

- `go test ./internal/generator/ -count=1` — 699 passed (includes new `TestGenerateStoreSubResourceUpsertBindingOrder`)
- `go test ./... -count=1` — 3489 passed across 34 packages
- `go fmt ./...` — clean
- `go vet ./...` — no issues
- `scripts/golden.sh verify` — 16/16 pass
- `git diff --check` — clean

## AI / Automation Disclosure

### DIFF
--- a/internal/generator/generator_test.go
+++ b/internal/generator/generator_test.go
@@ -2561,6 +2561,78 @@ func TestGenerateStoreUpsertBatchDispatchesToTypedTable(t *testing.T) {
 	runGoCommand(t, outputDir, "test", "./internal/store")
 }
 
+// TestGenerateStoreSubResourceUpsertBindingOrder asserts that the typed
+// upsert for a sub-resource table binds its argument values in the same
+// order as the SQL column declarations. buildSubResourceTable puts the
+// FK column between id and data, so the bindings have to match —
+// otherwise JSON blobs land in the FK column, timestamps land in data,
+// and FK values land in synced_at, silently corrupting every row.
+func TestGenerateStoreSubResourceUpsertBindingOrder(t *testing.T) {
+	t.Parallel()
+
+	apiSpec := &spec.APISpec{
+		Name:    "subres",
+		Version: "0.1.0",
+		BaseURL: "https://api.example.com",
+		Auth: spec.AuthConfig{
+			Type:    "api_key",
+			Header:  "Authorization",
+			Format:  "Bearer {token}",
+			EnvVars: []string{"SUBRES_API_KEY"},
+		},
+		Config: spec.ConfigSpec{
+			Format: "toml",
+			Path:   "~/.config/subres-pp-cli/config.toml",
+		},
+		Resources: map[string]spec.Resource{
+			"domains": {
+				Description: "Manage domains",
+				Endpoints: map[string]spec.Endpoint{
+					"list": {Method: "GET", Path: "/domains", Description: "List domains"},
+				},
+				SubResources: map[string]spec.Resource{
+					"verify": {
+						Description: "Verify a domain",
+						Endpoints: map[string]spec.Endpoint{
+							"get": {
+								Method:      "GET",
+								Path:        "/domains/{domainId}/verify",
+								Description: "Get verification status",
+								Params:      []spec.Param{{Name: "domainId", Type: "string", Required: true, Positional: true}},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+
+	outputDir := filepath.Join(t.TempDir(), naming.CLI(apiSpec.Name))
+	gen := New(apiSpec, outputDir)
+	require.NoError(t, gen.Generate())
+
+	storeSrc, err := os.ReadFile(filepath.Join(outputDir, "internal", "store", "store.go"))
+	require.NoError(t, err)
+	src := string(storeSrc)
+
+	// buildSubResourceTable inserts the FK column between id and data, so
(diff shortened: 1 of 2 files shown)


tmchow:
Greptile auto-review returned a trial-plan-limit notice instead of a code review on this PR. CI is fully green (every non-Mergify check passed). Flipping ready and applying `ready-to-merge` so Mergify can pick it up; happy to wait on a maintainer if you'd rather a human eyeballs it first.

Issue #870, opened 2026-05-10
Generator: wrapper-only catalog entries have no end-to-end generation path


rnnh-code:
## Summary

Catalog entries marked `mode: wrapper-only` (e.g., `google-flights`, which has `wrapper_libraries` populated) have a discovery and selection flow but no end-to-end generation flow. The skill captures the wrapper choice, but `printing-press generate` has no way to consume it — the user is forced to hand-write the Go module.

## Reproduction

1. Pick a wrapper-only catalog entry — e.g., `google-flights` (`mode: wrapper-only`, `wrapper_libraries` populated).
2. **Phase 1** (`printing-press` SKILL.md): the skill walks the user through choosing a wrapper library; the choice is persisted in `state.json` under `implementation`.
3. **Phase 2** (`printing-press generate`): the command requires `--spec` (OpenAPI or internal YAML). There is no `--wrapper` mode, and the `implementation` field from `state.json` is not consumed.
4. Result: there is no generation path. To get a working CLI, the user must hand-write the Go module that imports the chosen wrapper library.

## Two paths forward

**Option A — build the missing path.** Add a `--wrapper` mode (or equivalent) to `printing-press generate` that consumes the `implementation` field from `state.json` and emits a Go module importing the chosen wrapper library, with the standard CLI scaffolding around it.

**Option B — document the limitation.** Update `SKILL.md` to state explicitly that wrapper-only catalog entries are **catalog-discovery aids only** — the generator does not support them, and users must hand-write the Go module that imports the wrapper. Keep the discovery flow but stop the skill before it implies a generation step is coming.

Either path closes the gap; the current state silently dead-ends users.

## Environment

- Version: v4.2.2
- OS: macOS (Darwin 25.3.0)

ahujasachin92:
**Train-goat retro adds evidence for this issue.** Filed by `/printing-press-retro` on the train-goat run (2026-05-10, wrapper-only catalog entry for UK National Rail OpenLDBWS).

Same finding (`mode: wrapper-only` -> `printing-press generate --spec` required but no spec exists), different catalog entry. Train-goat hand-wrote a working **synthetic-seed-spec workaround** that could be documented as the interim path under Option B until Option A lands:

1. **Seed spec shape.** One placeholder resource in the generator's internal YAML format:
   ```yaml
   resources:
     - name: status
       path: /
       methods:
         - method: GET
           operation_id: get_status
   ```
[...]

PR #1056, opened 2026-05-11
docs(skills): clarify wrapper-only catalog entries have no generator path

tmchow:
## Intent

The Wrapper-only branch in Phase 1's catalog-check section of `skills/printing-press/SKILL.md` told the agent that "Phase 3 generation reads this to decide whether to `go get` a wrapper, emit a subprocess shell-out, or emit HTML-scrape code." `printing-press generate` only consumes `--spec` and has no wrapper-mode plumbing, so the skill silently dead-ended users on `google-flights`-style catalog entries: the discovery flow runs, the choice is captured, then `generate` rejects it.

This PR is the Option B fix from the issue (document the limitation) so the skill stops promising a path the generator does not implement. A `--wrapper` generation mode (Option A) is the bigger design call that should be a separate maintainer-driven RFC.

Issue: Refs #870

## Approach

Replace the 8 lines of skill prose at the Wrapper-only branch with text that matches what the binary actually does:

- State up front that the catalog entry is a discovery aid only and that there is no direct generation path from a wrapper-only entry today.
- Rewrite the `AskUserQuestion` text to offer the two real next steps: browser-sniff the upstream to author an internal YAML spec, or hand-write a Go module that imports the wrapper.
- Reframe `state.json.implementation` as skill bookkeeping rather than generator input, and add a `next_step` field so later phases can branch on which path the user picked.
- Route the browser-sniff choice into Phase 1.7 (which already exists); call out the hand-written path as a hand-off rather than something the press continues to drive.

`docs/CATALOG.md` already says wrapper-only entries do not by themselves make `printing-press generate <name>` work; this brings `SKILL.md` in line with that contract.

## Scope

Primary area: `skills` (the printing-press skill prose only).

Why this belongs in this repo: per AGENTS.md "Skill Authoring," skill prose that promises a generator capability has to track what the generator actually does. The mismatch is in this repo's `SKILL.md`, not in any printed CLI.

## Risk

Low. Prose-only change, no generator/template/parser code touched.

- Generator behavior: unchanged. Existing `--spec` flows are unaffected.
- Catalog validator: unchanged. `wrapper_libraries` shape is unmodified.
- Contract tests in `internal/pipeline/contracts_test.go` that assert SKILL.md setup blocks, workspace variables, and the run-root state file pass against the new prose.
- Skill behavioral contract for spec-based and "no catalog hit" branches: unchanged.

## Output Contract

N/A. No generated artifacts, manifests, command output, MCP schemas, scorecard output, catalog rendering, or pipeline artifacts changed by this PR. Golden harness re-verified to confirm.

## Verification

Run on macOS, against the branch at `e0d93e91`:

- `go fmt ./...` — clean
- `go vet ./...` — clean (`Go vet: No issues found`)
- `go build -o ./printing-press ./cmd/printing-press` — passes
- `go test ./... -count=1` — `3629 passed in 34 packages`
- `golangci-lint run ./...` — clean (`No issues found`)
- `scripts/golden.sh verify` — `Golden verify passed: 16 case(s).`
- `git diff --check` — clean

## AI / Automation Disclosure

### DIFF
--- a/skills/printing-press/SKILL.md
+++ b/skills/printing-press/SKILL.md
@@ -695,14 +695,14 @@ If the catalog has an entry for this API, branch on the entry type:
 - If catalog config: use the spec_url from the catalog entry, skip the research/discovery phase
 - If full discovery: proceed with the normal research workflow
 
-**Wrapper-only entry** (no `spec_url`, `wrapper_libraries` populated) — this is a reverse-engineered API that has no official spec but has known community libraries the generator can use as implementation backing. Do not try to resolve or browser-sniff a spec. Instead, surface the wrapper options to the user via `AskUserQuestion`:
+**Wrapper-only entry** (no `spec_url`, `wrapper_libraries` populated) — this is a reverse-engineered API that has no official spec but has known community libraries. The catalog entry is a **discovery aid only**: `printing-press generate` requires `--spec` and does not consume wrapper-library metadata, so there is no direct generation path from a wrapper-only entry today. Tell the user this up front via `AskUserQuestion`:
 
-> "<API> has no official spec. The catalog knows about these community-maintained implementations:"
+> "<API> has no official spec. The catalog knows about these community-maintained wrappers, but the Printing Press cannot generate a CLI directly from a wrapper. The next step has to be either browser-sniffing the upstream to author an internal YAML spec, or hand-writing a Go module that imports the wrapper. Which path do you want?"
 
-Present each `wrapper_libraries` entry as a selectable option with language, integration mode, and notes. Example for `google-flights`:
+Present each `wrapper_libraries` entry alongside the question with language, integration mode, and notes so the user can see what implementation backing exists. Example for `google-flights`:
 - **krisukox/google-flights-api** (Go, native, MIT) — Pure Go, importable; single-binary CLI with no runtime deps.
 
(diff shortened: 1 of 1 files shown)


tmchow:
Code-review fallback note: Greptile returned a trial-limit response (`tmchow` has reached the 50-review limit for trial accounts), so per the scheduled-task spec I switched to `/pr-review-toolkit:review-pr` for the code-review gate. Both reviewers (code-reviewer and comment-analyzer) ran against the diff and reported zero critical / zero important findings. Two accepted-for-later suggestions surfaced and are not blocking:

1. The new prose says `printing-press generate` requires `--spec`; the binary's actual error at `internal/cli/root.go:266` is `--spec is required (or use --plan for plan-driven generation)`. For the wrapper-only context this distinction does not change the load-bearing claim ("does not consume wrapper-library metadata"), but a follow-up could tighten the wording to "requires `--spec` or `--plan`".
2. Line 709 ("Adding new wrapper-only APIs: drop a YAML file in `catalog/` ... No skill changes needed.") carries a faint flavor of the same implied-generation pattern this PR is fixing. A follow-up could add "the entry will appear in `catalog show`; it does not enable `generate`."

Both are scope-creep relative to a 4-line correction and the reviewers marked them as such.

tmchow:
@mergifyio queue

tmchow:
@mergifyio queue

PR #1078, opened 2026-05-11
fix(cli): preserve unknown-shape records in compactListFields

### LINKED ISSUE #1046: compactListFields silently nukes records that don't match the id/name/title allowlist
## Summary

`compactListFields` in the generated `internal/cli/helpers.go` strips JSON-array records to `{}` when none of their fields match the hardcoded allowlist. This happens by default whenever an agent uses `--agent` (which expands to `--compact`), so any printed CLI whose response shapes don't use the canonical `id/name/title/...` schema is effectively broken for agent consumers.

## Reproduction

Discovered while sanity-checking my generated `trip-com-pp-cli`. With `--agent` on these commands, every record was emitted as an empty object:

```
$ trip-com-pp-cli locale list --agent | jq '.data[0:2]'
[{}, {}]

$ trip-com-pp-cli hot-and-cheap --from HKG --agent | jq '.[0:2]'
[{}, {}]

$ trip-com-pp-cli coupons top --city Tokyo --agent | jq '.[0]'
{}
```

Without `--agent`, the same commands return real data (full locale records, real flight fares to Singapore/Bangkok/Seoul, real coupon details). The data is being marshalled correctly; `compactListFields` strips it post-marshal.

A second, milder failure mode: records that DO match the allowlist on `id`/`name` but carry useful business fields (`price`, `rating`, `score`, `address`, etc.) lose those fields. `attractions ranked --agent` returned only `{id, name}` even though the underlying response had `price` and a deeplink.

## Root cause

`internal/generator/templates/.../helpers.go.tmpl` (or wherever the generated `helpers.go` is sourced):

```go
keepFields := map[string]bool{
    \"id\": true, \"name\": true, \"title\": true, \"identifier\": true,
    \"status\": true, \"state\": true, \"type\": true, \"priority\": true,
    \"url\": true, \"email\": true, \"key\": true,
    \"created_at\": true, \"updated_at\": true, \"createdAt\": true, \"updatedAt\": true,
}

filtered := make([]map[string]any, 0, len(items))
for _, item := range items {
    compact := map[string]any{}
    for k, v := range item {
        if keepFields[k] {
            compact[k] = v
        }
    }
    filtered = append(filtered, compact)
}
```

This allowlist assumes a SaaS-like canonical schema. APIs that use domain-specific field names (travel: `fare`, `currency`, `destination`, `airline`; commerce: `price`, `discount`, `vendor`; locale: `domain`, `language`, `locale`) match nothing and get nuked.

## Suggested fix

Two-pronged, both in `compactListFields`:

1. **Per-item fallback** — if an item's compacted form is empty, preserve the original. This is the safety net: novel-shape records never silently lose all data.

   ```go
   if len(compact) == 0 {
       compact = item
   }
   ```

2. **Wider default allowlist** — add common business/travel/commerce field names so canonical-schema records still get useful compaction (id, name, price, rating, ...) instead of degrading to (id, name) only.

Patch applied locally in trip-com-pp-cli for reference: included in `.printing-press-patches.json` patch id `compact-list-allowlist`.

## Related

While investigating, I also found:
- **Hand-written novel commands using `if dryRunOK(flags) { return nil }` produce rc=0 with empty stdout+stderr on `--dry-run`** — silent. Endpoint-mirror commands routing through `callTripPost` get a request-body preview, but the novel-command short-circuit emits nothing. Trivially fixed by having `dryRunOK` print a noop envelope before returning true.
- **`whichScoreEntry` only scores exact whole-query substring matches against the description** — multi-word phrasings ('find cheap weekend trips') score zero against entries whose description mentions some-but-not-all tokens. Adding per-token substring matching (capped at 3) restores resilience.

Both have local patches in the same `.printing-press-patches.json` if useful as reference.

## Impact

Any printed CLI whose API response shape uses domain-specific field names gets degraded `--agent` output. Agent consumers reading `--agent` JSON would see those endpoints as broken (empty records) even though the network calls succeeded. Severity is high because `--agent` is the primary path for all agent integrations and the failure is silent (no error, no warning, just empty objects).

tmchow:
## Intent

`compactListFields` in every printed CLI's `internal/cli/helpers.go` reduced any list item that didn't carry one of a narrow id/name/title/status/url/timestamps allowlist to `{}`. The helper runs whenever a user passes `--agent` (which expands to `--compact`), so any printed CLI whose API responses use domain-specific field names (travel: `fare`/`airline`; commerce: `vendor`/`price`; locale: `domain`/`language`) silently emitted empty objects to agent consumers. Records that did match `id`/`name` also lost useful business scalars (`price`, `rating`, `currency`) because they were not in the allowlist.

Issue:

Closes #1046

## Approach

Two changes in [internal/generator/templates/helpers.go.tmpl](internal/generator/templates/helpers.go.tmpl):

1. **Per-item fallback.** After the keep-loop, if the compacted map is empty, preserve the original item. This is the safety net — records that don't match any allowlist key round-trip whole instead of being nuked.
2. **Wider scalar allowlist.** Extend the keepFields map with high-signal scalars across categories: identity-adjacent (`code`, `slug`), monetary (`price`, `amount`, `cost`, `fare`, `rate`, `currency`), metrics (`rating`, `score`, `count`), locale/geo (`language`, `locale`, `country`, `region`, `city`, `domain`), temporal (`date`), versioning (`version`), categorization (`kind`).

Pinned via a new template-level test ([generator_test.go](internal/generator/generator_test.go)) mirroring `TestPipedJsonGateRespectsExplicitFormatFlags` — the same helper renders into every printed CLI's `helpers.go` so a per-fixture assertion would miss future copies that drift in. Goldens regenerated to reflect the new template output.

## Scope

Primary area: generator template — affects every future regen of every printed CLI.

Why this belongs in this repo: the bug is in the machine-emitted helper, not a single printed CLI. Per AGENTS.md "Default to machine changes" — fixing the template raises the floor for every API whose response shape uses domain-specific field names.

## Risk

- `--agent`/`--compact` output shape changes for any list item that previously emitted `{}`: it now emits the full record. This is the intended fix; agent consumers parsing `--agent` JSON should see real data where they previously got empty objects.
- Items that previously emitted just `{id, name}` may now also include `price`, `rating`, etc. when present — small payload-size increase per record on canonical-schema responses; same JSON shape contract (`{key: value, ...}`).
- Typed MCP endpoint mirrors do **not** route through `compactFields` (verified at `internal/generator/templates/mcp_tools.go.tmpl:476` — returns API data verbatim via `mcplib.NewToolResultText`), so the MCP contract agents rely on is unchanged. Cobratree walker shellouts that forward `--compact` do route through this helper and benefit from the fix.
- Pre-existing asymmetry: list-side fallback now preserves verbose fields (`description`, `body`) for unknown shapes while `compactObjectFields` strips them for single-object responses. Accepted design — symmetric stripping on the fallback path would defeat the safety net for records that happen to carry a description plus only one other field.

## Output Contract

Generator template change. Goldens regenerated:
- `testdata/golden/expected/generate-golden-api/printing-press-golden/internal/cli/helpers.go`
- `testdata/golden/expected/generate-golden-api-rich-auth/printing-press-rich-auth/internal/cli/helpers.go`

Only `compactListFields` rendering changed; no other generated artifacts shifted.

## Verification

```
go fmt ./...                     # clean
go vet ./...                     # no issues
go test ./...                    # 3627 passed in 34 packages
golangci-lint run ./...          # no issues
scripts/golden.sh verify         # 16 cases pass
```

Behavioral spot-check with the new helper compiled into a scratch binary against the issue's reproduction cases:

| Input | Before | After |
|-------|--------|-------|
| `{"domain":"x","language":"en","locale":"en_US"}` | `{}` | `{"domain":"x","language":"en","locale":"en_US"}` (allowlist hits) |
| `{"airline":"JL","fare":12345,"currency":"USD","destination":"NRT"}` | `{}` | `{"currency":"USD","fare":12345}` (partial — `airline`/`destination` drop, `--select` retrieves them) |
| `{"some_obscure_field":"v","another":42}` | `{}` | `{"some_obscure_field":"v","another":42}` (fallback fires) |
| `{"id":7,"name":"Widget","description":"verbose...","price":9.99,"rating":4.5}` | `{"id":7,"name":"Widget"}` | `{"id":7,"name":"Widget","price":9.99,"rating":4.5}` (allowlist keeps business scalars) |

### DIFF
--- a/internal/generator/generator_test.go
+++ b/internal/generator/generator_test.go
@@ -3332,6 +3332,39 @@ func TestGeneratedOutput_MutatingCommandsHaveEnvelope(t *testing.T) {
 	assert.Contains(t, content, `envelope["success"] = false`)
 }
 
+// TestCompactListFieldsPreservesUnknownShapes pins the two contracts that
+// keep `--agent` / `--compact` from silently emitting {} for records that
+// don't match the canonical id/name/title allowlist: a per-item fallback
+// to the original item, and a wider scalar allowlist that covers
+// monetary, metric, locale, and identity-adjacent keys.
+//
+// Pinned at the template level because the helper renders into every
+// printed CLI's helpers.go and a per-fixture assertion would miss future
+// copies that drift in.
+func TestCompactListFieldsPreservesUnknownShapes(t *testing.T) {
+	t.Parallel()
+
+	path := filepath.Join("templates", "helpers.go.tmpl")
+	data, err := os.ReadFile(path)
+	require.NoError(t, err, "template must exist: %s", path)
+	body := string(data)
+
+	assert.Contains(t, body, "if len(compact) == 0 {",
+		"compactListFields must preserve the original item when no allowlist keys match — otherwise records with domain-specific field names get reduced to {}")
+	assert.Contains(t, body, "compact = item",
+		"compactListFields must assign the original item as the per-item fallback")
+
+	for _, key := range []string{
+		`"price":`, `"fare":`, `"currency":`,
+		`"rating":`,
+		`"locale":`, `"language":`,
+		`"code":`,
+	} {
+		assert.Contains(t, body, key,
+			"compactListFields allowlist must include %s so canonical-schema records keep high-signal scalars across business/travel/commerce/locale APIs", key)
+	}
+}
+
 // TestPipedJsonGateRespectsExplicitFormatFlags pins the contract: the
 // piped-output auto-JSON gate must defer to explicit --csv / --quiet /
 // --plain flags so piped consumers that asked for a non-JSON format
--- a/internal/generator/templates/helpers.go.tmpl
+++ b/internal/generator/templates/helpers.go.tmpl
@@ -777,12 +777,31 @@ func compactFields(data json.RawMessage) json.RawMessage {
 }
 
 // compactListFields keeps only high-gravity fields for array responses.
+// When an item carries none of these keys the original is preserved, so
+// `--agent` does not silently emit {} for APIs whose response shapes use
+// domain-specific field names (travel: airline/destination; commerce: vendor).
 func compactListFields(items []map[string]any) json.RawMessage {
 	keepFields := map[string]bool{
+		// Identity
 		"id": true, "name": true, "title": true, "identifier": true,
-		"status": true, "state": true, "type": true, "priority": true,
-		"url": true, "email": true, "key": true,
+		"code": true, "slug": true, "key": true,
+		// Categorization
+		"status": true, "state": true, "type": true, "kind": true, "priority": true,
+		// Communication
+		"url": true, "email": true,
+		// Monetary
+		"price": true, "amount": true, "cost": true, "fare": true,
+		"rate": true, "currency": true,
+		// Metrics
+		"rating": true, "score": true, "count": true,
+		// Locale / geo
+		"language": true, "locale": true, "country": true, "region": true,
+		"city": true, "domain": true,
+		// Temporal
 		"created_at": true, "updated_at": true, "createdAt": true, "updatedAt": true,
+		"date": true,
+		// Versioning
+		"version": true,
 	}
 
 	filtered := make([]map[string]any, 0, len(items))
@@ -792,6 +811,9 @@ func compactListFields(items []map[string]any) json.RawMessage {
 			if keepFields[k] {
 				compact[k] = v
 			}
+		}
+		if len(compact) == 0 {
+			compact = item
 		}
 		filtered = append(filtered, compact)
 	}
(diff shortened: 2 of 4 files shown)


tmchow:
Greptile is unavailable for this account (trial-account 50-review limit reached). Ran the fallback `/pr-review-toolkit:review-pr` skill instead — code-reviewer, pr-test-analyzer, and comment-analyzer all report no critical or correctness findings.

One stylistic note accepted for later: the comment-analyzer flagged that the inline category label `// Locale / geo` encompasses `domain`, which is sometimes web/email rather than geographic. Defensible by issue context (#1046 cites `domain` as a locale-adjacent field), and the categorization comment is an organizational aid that doesn't affect behavior, so not addressed in this PR.

PR #985, opened 2026-05-10
fix(cli): force UTF-8 stdio in verify-skill Python subprocess

tmchow:
## What

Windows consoles default to cp1252, which cannot encode the `✓`/`✘` glyphs the verify-skill Python script prints. The subprocess crashed with `UnicodeEncodeError` mid-print, and the shipcheck umbrella reported `verify-skill FAIL` even when every underlying check passed. The ship verdict was wrong for every Windows user.

## Fix (defense in depth)

1. **Python side** — `sys.stdout.reconfigure(encoding="utf-8")` and `sys.stderr.reconfigure(encoding="utf-8")` at the top of `main()` in [scripts/verify-skill/verify_skill.py](scripts/verify-skill/verify_skill.py) and the bundled copy at [internal/cli/verify_skill_bundled.py](internal/cli/verify_skill_bundled.py). Self-contained, no env dependency, covers direct `python3 verify_skill.py …` invocations. Guarded by `try/except (AttributeError, OSError)` so an older Python or a non-reconfigurable stdout falls through cleanly. The helper lives on `main()` so importers (`test_resolve_command_path.py`) aren't affected.
2. **Go side** — new `pythonUTF8Env()` helper in [internal/cli/verify_skill.go](internal/cli/verify_skill.go) forces `PYTHONIOENCODING=utf-8` and `PYTHONUTF8=1` on the subprocess env, dropping any conflicting parent values first. Robust even if a future edit removes the Python-side guard.

Either change alone closes the bug; both together is belt-and-suspenders against env stripping or script edits.

## Tests

- `TestPythonUTF8Env_AppendsEncodingVarsWhenAbsent`
- `TestPythonUTF8Env_OverridesUserSetting` — ensures a cp1252 inherited from the parent shell is dropped, not preserved
- `TestPythonUTF8Env_EmptyBase`
- `TestVerifySkillScriptInSync` (existing) confirms the bundled copy matches the canonical
- Existing `python3 scripts/verify-skill/test_resolve_command_path.py`: 14/14 pass (helper guard doesn't fire on import)

Full suite: `go test ./...` → 3474 passed.
Lint: `golangci-lint run ./...` → clean.
Golden: `scripts/golden.sh verify` → 16/16 pass.

## Scope boundary

- IN: verify-skill Python stdio + Go subprocess env propagation.
- OUT: other Python scripts in the scorer suite (each can get the same one-line fix independently if it prints non-ASCII).

Closes #976
Closes #819

PR #1086, opened 2026-05-11
fix(cli): skip Windows Python Store stub in verify-skill

### LINKED ISSUE #876: verify-skill Python invocation breaks on Windows in three places
> Filed by `/printing-press-retro` from a run of **TradingView** (scorecard 61/100, 100% verify pass after fixes).

## Summary

`shipcheck`'s `verify-skill` leg crashes on Windows in three independent ways before any check runs. Each blocks the next: (1) the `python3` interpreter resolves to the Microsoft Store launcher stub by default and exits with a "Python was not found" message; (2) the bundled Python script's `Path.read_text()` calls default to cp1252 and trip on UTF-8 SKILL.md content (em-dashes, arrows, curly quotes); (3) the script's `print()` output also defaults to cp1252 and crashes on the `✓` it emits in summary lines. Sub-bug 3 is already filed as #819 (P3). Sub-bugs 1 and 2 are not yet filed and surface before #819's fix would be reached.

## Where to look

- **Component:** scorer
- **Likely area:** `internal/cli/verify_skill.go` (the Go invocation: lines ~63-73 build the `pyArgs` and exec `python3`) and `internal/cli/verify_skill_bundled.py` (the embedded Python script: 10 `Path.read_text()` call sites, plus the stdout-print sites covered by #819).
- **Triggered when:** `printing-press shipcheck` or `printing-press verify-skill` on `runtime.GOOS == "windows"` against any printed CLI whose SKILL.md contains a non-ASCII character. Generated SKILL.md emits non-ASCII routinely from research narratives.

## What we observed

Three sequential `printing-press shipcheck` retries on Windows 11 + Python 3.14 each produced a different Python error:

**Sub-bug 1 — Microsoft Store stub:**
```
=== verify-skill ===
Python was not found; run without arguments to install from the Microsoft Store, or
disable this shortcut from Settings > Apps > Advanced app settings > App execution aliases.
```
`exec.LookPath("python3")` resolves to `C:\Users\<user>\AppData\Local\Microsoft\WindowsApps\python3.exe`, which is the Store launcher stub, not a real Python.

**Sub-bug 2 — file decode crash:**
```
File "C:\Users\TOOTHL~1\AppData\Local\Temp\verify-skill-1996074093.py", line 605, in extract_recipes
    text = skill.read_text()
File "C:\Python314\Lib\encodings\cp1252.py", line 23, in decode
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 121
```
SKILL.md contains an em-dash byte sequence; cp1252 has no mapping for `0x9d`.

**Sub-bug 3 — stdout encode crash (the part #819 covers):**
```
File "C:\Users\TOOTHL~1\AppData\Local\Temp\verify-skill-4072580677.py", line 1042, in main
    print(format_human(report))
File "C:\Python314\Lib\encodings\cp1252.py", line 19, in encode
UnicodeEncodeError: 'charmap' codec can't encode character '✓' in position 30
```

Patching all three produced a clean PASS:
```
=== tradingview-pp-cli ===
  ✓ All checks passed (flag-names, flag-commands, positional-args, unknown-command)
```

## Suspected root cause

Confident on each sub-bug independently:

1. **Interpreter selection** — `verify_skill.go` line 74 hardcodes `exec.Command("python3", ...)`. `LookPath("python3")` on Windows finds the WindowsApps stub before any real interpreter; the stub exits non-zero with the Microsoft Store error.
2. **File-read default encoding** — `verify_skill_bundled.py` calls `Path.read_text()` 10 times without `encoding="utf-8"`. Python on Windows defaults `read_text()` to the system locale codec (cp1252 on en-US), which lacks UTF-8 byte mappings for em-dashes / arrows / curly quotes that SKILL.md routinely contains.
3. **Stdout default encoding** — Python's `sys.stdout.encoding` defaults to the locale codec on Windows. The script prints `✓` which cp1252 cannot encode.

## Suggested direction

Three independent two-line fixes:

```go
// internal/cli/verify_skill.go — sub-bug 1 + 3 (env vars are #819's scope)
pyBin := "python3"
if path, err := exec.LookPath("python3"); err != nil || isWindowsStorePython(path) {
    if _, err := exec.LookPath("py"); err == nil {
        pyBin = "py"
    } else if path, err := exec.LookPath("python"); err == nil && !isWindowsStorePython(path) {
        pyBin = "python"
    }
}
py := exec.Command(pyBin, pyArgs...)
py.Env = append(os.Environ(), "PYTHONIOENCODING=utf-8", "PYTHONUTF8=1")  // #819

func isWindowsStorePython(path string) bool {
    if path == "" { return false }
    lower := strings.ToLower(path)
    return strings.Contains(lower, "windowsapps") && strings.Contains(lower, "python")
}
```

```python
# internal/cli/verify_skill_bundled.py — sub-bug 2
# Replace every .read_text() call with .read_text(encoding="utf-8")
# (10 occurrences as of HEAD)
```

`isWindowsStorePython` is needed because `LookPath` succeeds against the Store stub — the stub *is* on PATH, it just doesn't function as Python. Path-based detection is the only signal available before exec.

`PYTHONIOENCODING=utf-8` + `PYTHONUTF8=1` (defense in depth — both env vars matter on different Python versions) is what #819 already proposes for sub-bug 3. The interpreter-selection fix and the file-decode fix should ship together since the failure modes cascade in that order.

## Acceptance criteria

- positive: on Windows in CI, `printing-press shipcheck` produces a verify-skill PASS (or honest FAIL based on actual mismatches, not on a Python error). SKILL.md content with em-dashes round-trips through `read_text()` without crash. The script's `✓` summary prints without `UnicodeEncodeError`.
- negative: on macOS/Linux, `verify-skill` continues to use `python3` first, behaves identically — same exit codes, same finding set, same output formatting.

## Frequency

every Windows shipcheck.

## Complexity

medium (3 distinct sub-fixes across 2 files; the `LookPath` stub-detection logic is the trickiest piece, but it's roughly 5 lines of Go).

## Scope boundary

Don't restructure verify-skill's check set or add new findings. Purely about making the existing check set runnable on Windows.

## Dependencies

#819 covers sub-bug 3 in isolation. If #819 is merged first, sub-bugs 1 and 2 still remain — the cascade order means users won't reach sub-bug 3 until 1 and 2 are fixed. Recommend either subsuming #819 into this issue or shipping all three sub-fixes together.

#### F2: verify-skill Python invocation breaks on Windows in three places

- Symptom: three sequential shipcheck retries each produced a different Python error (Store stub message → cp1252 decode crash → cp1252 encode crash). Verify-skill's own checks never got to run.
- Evidence: stack traces from `verify-skill-*.py` temp scripts captured during Phase 4 of the run. Patching all three made the leg PASS.

## Related issues

- #819 — open issue (`related-area`): covers sub-bug 3 (stdout cp1252) only. Maintainer should decide whether to subsume #819 into this broader fix or ship them in sequence — sub-bug 3 is unreachable in the cascade until 1 and 2 are fixed.

## Artifacts

| Artifact | Link |
|----------|------|
| Retro document | https://files.catbox.moe/p4u8mn.md |
| Manuscripts (research + proofs) | https://files.catbox.moe/w1oxy2.zip |
| Generated CLI source code | https://files.catbox.moe/8jue01.zip |

---

tmchow:
## Intent

Windows verify-skill crashed before any check ran because `exec.LookPath("python3")` resolves to the Microsoft Store launcher stub at `.../WindowsApps/python3.exe`. The stub is on PATH but exits with "Python was not found; run without arguments to install from the Microsoft Store" instead of executing the embedded verifier.

Issue: Closes #876

## Approach

`runVerifySkillScript` now calls a new `resolvePython()` helper instead of hard-coding `"python3"`. The helper:

- Returns `"python3"` unchanged on non-Windows hosts (no extra `LookPath` cost).
- On Windows, tries `python3` first, rejecting it when the resolved path matches the Store stub (`isWindowsStorePython`). Falls back to the Python Launcher `py`, then `python` (also stub-checked), then `python3` as a last-resort no-op.

`isWindowsStorePython` checks for both `windowsapps` and `python` in the lower-cased path. Path-based detection is the only signal — `LookPath` succeeds against the stub because it is a real executable that exits non-zero on launch.

Sub-bugs 2 (UTF-8 file decode) and 3 (UTF-8 stdout encode) from issue #876 were already resolved by #985 (`read_utf8` helper plus `pythonUTF8Env` env vars and `sys.stdout.reconfigure`); this PR only covers the remaining interpreter-selection cascade.

## Scope

Primary area: `internal/cli/verify_skill.go`. Machine change only — no generator templates, catalog, MCP surface, or printed-CLI output affected.

Why this belongs in this repo: this is the `printing-press verify-skill` runtime in the Press binary itself, not a generated CLI. Failure mode reproduces on every Windows `shipcheck` against any printed CLI whose SKILL.md contains non-ASCII content.

## Risk

Low. Non-Windows hosts hit a single `runtime.GOOS` check and the prior code path. On Windows, the worst case (no real Python anywhere, only the Store stub) reaches the same failing exec the unfixed binary already produced. The added helpers are pure functions with no shared state.

## Output Contract

N/A — no generator output, catalog, manifest, MCP schema, or scorecard text changes. `scripts/golden.sh verify` passes unchanged.

## Verification

- `go fmt ./...`
- `go vet ./...`
- `go test ./...` — 3687 passed in 34 packages
- `golangci-lint run ./...`
- `scripts/golden.sh verify` — 16 cases pass
- Unit tests added: `TestIsWindowsStorePython` (9 table cases covering stub paths, mixed case, forward slashes, real installs, `py` launcher, Unix path, unrelated WindowsApps binary, empty path).

Windows runtime behavior of `resolvePython` is not directly covered by Linux/macOS Go tests — the four fallback paths can only fire when `runtime.GOOS == "windows"`. Restructuring the helper to inject `runtime.GOOS` + `exec.LookPath` for testability was discussed during review and accepted-for-later: the helper is six lines, the path-detection logic (`isWindowsStorePython`) is fully unit-tested, and the testing reviewer assessed the residual risk as acceptable given the simplicity.

### DIFF
--- a/internal/cli/verify_skill.go
+++ b/internal/cli/verify_skill.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/mvanhorn/cli-printing-press/v4/internal/generator"
@@ -24,6 +25,12 @@ import (
 var verifySkillScript string
 
 const canonicalSectionsCheckName = "canonical-sections"
+
+const (
+	pythonPython3 = "python3"
+	pythonPy      = "py"
+	pythonPython  = "python"
+)
 
 type verifySkillRunResult struct {
 	Stdout   string
@@ -71,7 +78,7 @@ func runVerifySkillScript(dir string, only []string, asJSON bool, strict bool) (
 		pyArgs = append(pyArgs, "--strict")
 	}
 
-	py := exec.Command("python3", pyArgs...)
+	py := exec.Command(resolvePython(), pyArgs...)
 	py.Stdin = os.Stdin
 	py.Env = pythonUTF8Env(os.Environ())
 	var stdout, stderr bytes.Buffer
@@ -333,6 +340,37 @@ func emitMergedJSON(pyResult verifySkillRunResult, runCanonical, hasCanonicalFin
 	}
 	fmt.Println(string(out))
 	return nil
+}
+
+// resolvePython picks the interpreter name to pass to exec.Command for the
+// embedded verify-skill script. The Windows fallback chain exists because
+// exec.LookPath("python3") on Windows often resolves to the Microsoft Store
+// launcher stub, which is on PATH but exits with a "Python was not found"
+// Store-redirect message instead of running the script.
+func resolvePython() string {
+	if runtime.GOOS != "windows" {
+		return pythonPython3
+	}
+	if path, err := exec.LookPath(pythonPython3); err == nil && !isWindowsStorePython(path) {
+		return pythonPython3
+	}
+	if _, err := exec.LookPath(pythonPy); err == nil {
+		return pythonPy
+	}
+	if path, err := exec.LookPath(pythonPython); err == nil && !isWindowsStorePython(path) {
+		return pythonPython
+	}
+	return pythonPython3
+}
+
+// LookPath succeeds against the Microsoft Store Python launcher stub at
+// .../WindowsApps/python*.exe, so the only pre-exec signal is the path itself.
+func isWindowsStorePython(path string) bool {
+	if path == "" {
+		return false
+	}
+	lower := strings.ToLower(path)
+	return strings.Contains(lower, "windowsapps") && strings.Contains(lower, "python")
 }
 
 // pythonUTF8Env returns base with PYTHONIOENCODING and PYTHONUTF8 forced to
--- a/internal/cli/verify_skill_internal_test.go
+++ b/internal/cli/verify_skill_internal_test.go
@@ -64,3 +64,32 @@ func TestPythonUTF8Env_EmptyBase(t *testing.T) {
 		t.Errorf("expected both UTF-8 entries, got %v", got)
 	}
 }
+
+func TestIsWindowsStorePython(t *testing.T) {
+	t.Parallel()
+
+	cases := []struct {
+		name string
+		path string
+		want bool
+	}{
+		{"store stub python3", `C:\Users\alice\AppData\Local\Microsoft\WindowsApps\python3.exe`, true},
+		{"store stub python", `C:\Users\alice\AppData\Local\Microsoft\WindowsApps\python.exe`, true},
+		{"store stub mixed case", `C:\Users\alice\AppData\Local\Microsoft\WINDOWSAPPS\Python3.exe`, true},
+		{"store stub forward slashes", `C:/Users/alice/AppData/Local/Microsoft/WindowsApps/python3.exe`, true},
+		{"real python install", `C:\Python314\python.exe`, false},
+		{"py launcher", `C:\Windows\py.exe`, false},
+		{"unix path", "/usr/bin/python3", false},
+		{"unrelated windowsapps binary", `C:\Users\alice\AppData\Local\Microsoft\WindowsApps\winget.exe`, false},
+		{"empty path", "", false},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			if got := isWindowsStorePython(tc.path); got != tc.want {
+				t.Errorf("isWindowsStorePython(%q) = %v, want %v", tc.path, got, tc.want)
+			}
+		})
+	}
+}
--- a/internal/cli/verify_skill_internal_test.go
+++ b/internal/cli/verify_skill_internal_test.go
@@ -75,6 +75,7 @@ func TestIsWindowsStorePython(t *testing.T) {
 	}{
 		{"store stub python3", `C:\Users\alice\AppData\Local\Microsoft\WindowsApps\python3.exe`, true},
 		{"store stub python", `C:\Users\alice\AppData\Local\Microsoft\WindowsApps\python.exe`, true},
+		{"store stub versioned python3.13", `C:\Users\alice\AppData\Local\Microsoft\WindowsApps\python3.13.exe`, true},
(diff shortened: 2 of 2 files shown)


tmchow:
Fallback review via `/pr-review-toolkit:review-pr` (Greptile is at trial-limit for this account). Findings, with my disposition:

**No Critical or correctness issues blocking merge.**

Important — accepted-for-later (out of scope for sub-bug 1 of #876, which is purely about interpreter cascade order):
- `resolvePython` returns `python3` as a no-op when every Windows probe fails. On a degenerate system with only the Store stub and no real Python or `py.exe`, the user re-encounters the Store-redirect message instead of an actionable "no usable Python interpreter" diagnostic. Improving this is additive — it touches the function signature and the caller's error path — and the issue's scope boundary says "don't restructure verify-skill's check set." Worth a follow-up.
- `exec.LookPath` permission-denied is treated identically to not-found across the fallback chain. A permission/IO error on a real `python3.exe` install is silently swallowed when a working `py.exe` is found. Same scope rationale; follow-up.

Suggestion — applied:
- Added a `python3.13.exe` test case to pin the substring match against a future tightening to strict basename detection (commit `test(cli): pin versioned Microsoft Store stub case`).

Suggestion — declined:
- DI refactor of `resolvePython` for unit-testable fallback ordering. Function is 6 lines; the only non-trivial predicate (`isWindowsStorePython`) is fully covered. The testing reviewer in the prior ce-code-review pass also classified this as acceptable residual.
- Further wording polish on the two new comments (a prior reviewer pass already trimmed the WHAT lead-in on `isWindowsStorePython`).

Issue #1060, opened 2026-05-11
[P2] WU-1: Tier 2 sync.walker spec parameter for hierarchical APIs


justinwfu:
**Priority:** P2 · **Component:** generator

## Goal

Generate hierarchical-walk sync code from a spec declaration so APIs like Yahoo Fantasy, Reddit pre-2024, YouTube Data v3, and MLB Stats no longer require a hand-written 990-line walker per CLI.

## Context

PR #1009 (commit `5fabad63`, merged 2026-05-11, closing #1006) landed **Tier 1**: when a resource path retains `{key}` placeholders after spec-driven substitution, sync emits a structured `sync_warning` event and continues instead of aborting the entire run on HTTP 500. The PR body explicitly defers **Tier 2** — a generator-side `sync.walker` spec parameter for hierarchical orchestration — as "a separate retro after more catalog evidence accumulates." Catalog evidence is now concrete (four APIs named with concrete path shapes), so this issue picks up that deferred work.

The yahoo-fantasy printed CLI currently carries 990 lines of hand-written `yahoo_walker.go` (under `internal/cli/yahoo_walker.go` in the cli-source zip) that walks user → games → leagues → teams → rosters. The file's 22-line header documents itself as the explicit Tier 2 placeholder: *"The generator-side equivalent ('sync.walker' spec parameter) is the retro's Tier 2 fix and is intentionally out of scope here."*

## Target

- `internal/generator/templates/sync.go.tmpl` — walker code-gen branch
- `internal/spec/` — parser for the `sync.walker` internal-YAML field
- `internal/openapi/` — parser for the `x-pp-sync-walker` operation extension

## Acceptance criteria

**Positive test:** a fixture spec declaring a 2-level walker (parent resource + child path with `{parent_key}` placeholder + key-extraction rule from the parent's response) produces a generated `sync.go` that:
- fetches the parent resource via the existing flat-list path
- extracts the parent keys per the spec's key-extraction rule
- iterates the child fetches under the existing concurrency/retry/cursor machinery
- routes leaves into typed store tables via the existing Upsert paths

**Negative test:** existing flat-list specs under `testdata/golden/` continue to emit byte-identical sync code. The walker branch is gated on presence of `sync.walker` (internal YAML) or `x-pp-sync-walker` (OpenAPI operation extension).

## Scope boundary

[...]

PR #1099, opened 2026-05-11
fix(cli): gate OAuth refresh-token params on x-oauth-refresh-token-mechanism

### LINKED ISSUE #916: OAuth template: refresh-token semantics are Google-only by default
## Problem

The OAuth template emits `access_type=offline` as an authorization-URL query param across every CLI it generates. This is **Google-specific**: only Google's OAuth server reads that param. Other providers (WHOOP, X/Twitter, GitHub, Spotify, …) ignore it, and most of them require a different mechanism — a **scope value** like `offline`, `offline.access`, or `offline_access` — to be returned a refresh_token from the token endpoint.

When the template gets the mechanism wrong, the symptom is invisible until the access-token TTL expires:

- token endpoint returns no refresh_token
- the CLI's transparent refresh path has nothing to use
- every call after the access-token TTL 401s
- recovery requires an interactive re-auth

## Reported externally

mvanhorn/printing-press-library#391 (whoop)

## Library-side patch

mvanhorn/printing-press-library#395 patches the published whoop copy: drop `access_type=offline`, append `offline` to the scope list.

## Audit of currently-published CLIs

Five library CLIs use the `access_type=offline` template path:

| CLI | Provider | Status |
|---|---|---|
| whoop | WHOOP | broken — needed `offline` scope value (patched in #395) |
| x-twitter | X/Twitter | harmless — `offline.access` is also in the scope list, so refresh tokens still issue |
| cloud-run-admin | Google | correct |
| google-ads | Google | correct |
| google-photos | Google | correct |

Whoop happened to expose the latent bug because its scope list lacked the magic value. Any future non-Google OAuth CLI will hit the same issue unless the template changes.

## Suggested fix

Move the refresh-token mechanism out of the template's hardcoded defaults and into a spec-annotation:

\`\`\`yaml
# OpenAPI spec
x-oauth-refresh-token-mechanism: scope:offline
# or
x-oauth-refresh-token-mechanism: scope:offline.access
# or
x-oauth-refresh-token-mechanism: query:access_type=offline   # Google
\`\`\`

The template would consume the annotation and emit the right shape:
- `scope:<value>` → append `<value>` to the scope list, no query param
- `query:k=v` → set `params.Set(k, v)`, no scope changes

Default when annotation is absent: don't emit either. Forcing an explicit declaration is better than a Google-shaped silent default that silently breaks other providers.

A simpler alternative: ship a small built-in mapping from well-known token URLs to mechanisms (api.prod.whoop.com → scope:offline, api.x.com → scope:offline.access, accounts.google.com → query:access_type=offline). Less explicit but covers the common-case providers without spec annotations.

## Severity

Low for any CLI whose user runs commands manually (they just re-auth when prompted). High for any CLI an agent or cron job depends on for unattended sync — those break exactly 1h after first auth and there's no in-band recovery.

tmchow:
## Intent

The OAuth template emits `access_type=offline` and `prompt=consent` as authorization-URL query params on every `authorization_code` CLI it generates. Both are Google-specific; non-Google providers ignore them and require a magic *scope* value instead (`offline`, `offline.access`, `offline_access`). The failure mode is invisible: the token endpoint silently omits `refresh_token`, the transparent refresh path has nothing to use, and every call after the access-token TTL 401s. WHOOP hit this in production -- patched downstream in [mvanhorn/printing-press-library#395](https://github.com/mvanhorn/printing-press-library/pull/395), but every future non-Google OAuth CLI inherits the same shape unless the generator changes.

Issue: Closes #916

## Approach

Adds an `x-oauth-refresh-token-mechanism` security-scheme extension parsed into `AuthConfig.RefreshTokenMechanism`. A single `ParseRefreshTokenMechanism()` method returns a typed `ParsedRefreshTokenMechanism` struct; the template pins the parse to a local variable and branches on `Kind`:

- `scope:<value>` -> `scopes = append(scopes, <value>)`
- `query:<k>=<v>` -> `params.Set(<k>, <v>)`
- empty / malformed -> emit neither

Default is silent. A Google-shaped default that silently breaks other providers is worse than no default. Spec authors declare the mechanism their API actually needs; unannotated specs get a clean authorization URL.

Reserved authorization-URL parameter names (`client_id`, `redirect_uri`, `response_type`, `state`, `scope`) are rejected from the `query` form to prevent a spec author from silently clobbering the generator's CSRF state token or core OAuth params.

I considered a built-in token-URL-to-mechanism map (the alternative the issue called out), but the spec annotation is more honest -- it puts the choice in the version-controlled spec rather than in a generator heuristic that has to guess which API a spec is describing.

## Scope

Primary area: `cli` (generator template + spec extension + parser plumb).

Why this belongs in this repo: machine change per AGENTS.md. The same bug reproduces on every printed CLI generated from an OAuth2 `authorization_code` spec; fixing it in one CLI's source would not generalize and would re-break on next regen.

## Risk

**Behavior change on regeneration for Google CLIs.** The audit in #916 covers five published CLIs using the existing template path. Three are Google (`cloud-run-admin`, `google-ads`, `google-photos`) and currently rely on the hardcoded `access_type=offline` + `prompt=consent`. After this change, those three CLIs regenerated without declaring `x-oauth-refresh-token-mechanism: query:access_type=offline` lose:

1. **`access_type=offline`** -- Google still issues a refresh_token on first consent (the user granted offline access during the OAuth dance), but subsequent re-consents may not.
2. **`prompt=consent`** -- forced re-consent on subsequent logins. Without it, Google can skip the consent screen and skip re-issuing a refresh_token.

The single-mechanism shape cannot express Google's two-param recipe. Documented in `docs/SPEC-EXTENSIONS.md` as a known constraint; multi-mechanism syntax is out of scope here. Existing published Google CLIs are not regenerated automatically; the next maintainer-driven regen needs to add the annotation.

For the other two CLIs in the audit:

- `whoop`: already patched in the library (#395). A regen with this PR's template + a `scope:offline` annotation in the WHOOP spec lands the fix here.
- `x-twitter`: harmless today (`offline.access` is already in the scope list). A regen without an annotation drops `access_type=offline` (X ignored it anyway).

Other risk surfaces:

- **Existing catalog entries** without a `x-oauth-refresh-token-mechanism` annotation: behavior unchanged on the parser side; only the `authorization_code` template emission changes.
- **OAuth2 `client_credentials` grant**: template path is separate (`auth_oauth_cc.go.tmpl`). Verified by `client_credentials grant ignores mechanism` test that the new field has no effect there.
- **Parser API surface**: `ParseRefreshTokenMechanism` is additive; existing `Parse` / `ParseLenient` / `ParseFile*` signatures unchanged.
- **MCP surface / scorer / publish flow**: no changes.

## Output Contract

- `internal/generator/templates/auth.go.tmpl` no longer emits the hardcoded `"access_type": {"offline"}` and `"prompt": {"consent"}` entries in the authorization-URL `url.Values` literal. They are replaced with conditional `params.Set` / `scopes = append` blocks gated on `$refreshMech.Kind`.
- Generated CLIs for OAuth2 `authorization_code` specs without `x-oauth-refresh-token-mechanism`: cleaner authorization URL (no Google-specific params).
- Generated CLIs for OAuth2 `authorization_code` specs *with* the annotation: emit exactly the requested mechanism.
- No golden fixture currently covers the `authorization_code` grant template path. Pre-existing gap; the new unit tests in `TestGenerateOAuth2RefreshTokenMechanism` exercise both branches plus the default. Adding a dedicated golden case for this grant is left as a follow-up.
- All 16 existing golden cases pass byte-identically (none exercise authorization_code).

## Verification

Commands actually run on macOS / go 1.25.4:

- `go fmt ./...` -- clean
- `go vet ./...` -- clean
- `go build ./...` -- passes
- `go test ./...` -- 3749 passed in 34 packages, including 3 new test suites:
  - `TestParseRefreshTokenMechanism` (spec parsing, 16 cases incl. reserved-key rejection)
  - `TestOpenAPIOAuthRefreshTokenMechanism` (extension round-trip)
  - `TestGenerateOAuth2RefreshTokenMechanism` (template emission, plus client_credentials grant isolation)
- `scripts/golden.sh verify` -- 16 cases pass
- `golangci-lint run ./...` -- 0 issues

## AI / Automation Disclosure

### DIFF
--- a/docs/SPEC-EXTENSIONS.md
+++ b/docs/SPEC-EXTENSIONS.md
@@ -31,6 +31,7 @@ in the same change as any new `Extensions["x-*"]` lookup in that file.
 | `x-auth-description` | `components.securitySchemes.<name>` | `APISpec.Auth.Description` | No |
 | `x-auth-cookie-domain` | `components.securitySchemes.<name>` | `APISpec.Auth.CookieDomain` | No |
 | `x-auth-cookies` | `components.securitySchemes.<name>` | `APISpec.Auth.Cookies` | No |
+| `x-oauth-refresh-token-mechanism` | `components.securitySchemes.<name>` | `APISpec.Auth.RefreshTokenMechanism` | No |
 | `x-resource-id` | path item | `Endpoint.IDField` | No |
 | `x-critical` | path item | `Endpoint.Critical` | No |
 | `x-tier` | path item or operation | `Endpoint.Tier` | No |
@@ -494,6 +495,58 @@ components:
       x-auth-cookies:
         - session_id
         - csrf_token
+```
+
+### `x-oauth-refresh-token-mechanism`
+
+Declares how the authorization endpoint should be asked to issue a refresh
+token. Providers diverge: Google reads `access_type=offline` as a query
+parameter, while WHOOP, X/Twitter, and others read a magic scope value
+(`offline`, `offline.access`, `offline_access`) instead. The generator emits
+neither by default because a Google-shaped silent default silently breaks
+other providers (broken refresh path is invisible until access-token TTL
+expires).
+
+Parsed field: `APISpec.Auth.RefreshTokenMechanism`
+
+Rules:
+
+- Optional. Only consumed by the authorization_code grant template; ignored
+  for other grants and non-OAuth2 auth.
+- Must be a string. Leading and trailing whitespace on the whole value is
+  trimmed.
+- Two exact-match prefixes are accepted:
+  - `scope:<value>` appends `<value>` to the scope list. No query param is
+    added.
+  - `query:<key>=<value>` sets the query parameter exactly once. No scope
+    change.
+- Malformed values (empty key, empty value, missing `=` for `query`, unknown
+  prefix, uppercase prefix) are ignored and produce no emission.
+- For `query:<key>=<value>`, the reserved authorization-URL parameter names
+  `client_id`, `redirect_uri`, `response_type`, `state`, and `scope` are
+  rejected. Permitting them would let a spec author silently overwrite the
+  generator's CSRF state token or core OAuth params.
+- Note: the single-mechanism shape cannot express Google's two-param recipe
+  (`access_type=offline` + `prompt=consent`). The first param is sufficient
+  for refresh-token issuance on initial consent; the second forces re-consent
+  on subsequent logins to keep the refresh-token contract alive. Specs that
+  need both should declare one via this extension and add the other through a
+  future multi-mechanism syntax (out of scope here).
+
+Example:
+
+```yaml
+components:
+  securitySchemes:
+    OAuth2:
+      type: oauth2
+      x-oauth-refresh-token-mechanism: scope:offline
+      flows:
+        authorizationCode:
+          authorizationUrl: https://api.example.com/oauth/authorize
+          tokenUrl: https://api.example.com/oauth/token
+          scopes:
+            read: Read access
 ```
 
 ## Path Item Extensions
--- a/internal/generator/generator_test.go
+++ b/internal/generator/generator_test.go
@@ -657,6 +657,119 @@ func TestGenerateOAuth2AuthTemplateConditionally(t *testing.T) {
 	})
 }
 
+func TestGenerateOAuth2RefreshTokenMechanism(t *testing.T) {
+	t.Parallel()
+
+	baseSpec := func() *spec.APISpec {
+		return &spec.APISpec{
+			Name:    "oauthapi",
+			Version: "0.1.0",
+			BaseURL: "https://api.example.com",
+			Auth: spec.AuthConfig{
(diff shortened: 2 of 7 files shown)


tmchow:
Rebased onto current `main`. The conflict in `internal/openapi/parser.go` was the `extensionAuthEnvVars` constants block — main added `extensionSyncWalker` (from #1060) while this PR added `extensionOAuthRefreshTokenMech`, two non-overlapping additions to the same declaration block. Resolved by union; both constants now present.

All verification gates green on the rebased tree:
- `go build ./...` — passes
- `go vet ./...` — clean
- `go fmt ./...` — clean
- `golangci-lint run ./...` — 0 issues
- `go test ./...` — 3784 passed in 34 packages
- `scripts/golden.sh verify` — 17 cases pass (including the new `generate-sync-walker-api` from main)

PR state is now `mergeable: MERGEABLE`.

Issue #927, opened 2026-05-10
generator/profiler: infer pagination params from plain params; preserve cursor=0 in paginatedGet


Bignella:
> Surfaced by the OneDev retro (May 2026 run). The full retro doc is in the related issue and locally archived.

## Goal

When a spec author declares pagination params (`offset`, `count`, `limit`, `page`, `per_page`, `cursor`, etc.) as plain `params:` entries — instead of an explicit `pagination:` block — the profiler should detect them and emit the correct defaults instead of falling back to the literal `after`/`limit` defaults. The generated `paginatedGet` helper should not strip `"0"` from the cursor param, so APIs that require explicit `offset=0` on first page work without manual patching.

## What happened

OneDev list endpoints declare `offset` (int) and `count` (int) as plain `params:` entries on every list endpoint. No endpoint had a `pagination:` block. The profiler (`internal/profiler/profiler.go`) only counts `cursorParams[name]++` and `pageSizeParams[name]++` when `endpoint.Pagination != nil` (lines 370–377). With zero matches, `mostCommon(...)` returned the literal default fallback values: `cursor → "after"`, `limit → "limit"`. The generated `sync.go` then sent `?after=…&limit=…`, OneDev rejected with HTTP 406:

```
GET /agents returned HTTP 406: Unexpected query params: [limit]
```

After patching `determinePaginationDefaults()` to return `offset`/`count`, OneDev accepted the param names but failed with:

```
GET /agents returned HTTP 406: Missing query params: [offset]
```

— OneDev requires `?offset=0` to be **explicitly present** on first page. The generic `paginatedGet` helper in `helpers.go.tmpl` strips `"0"` and `"false"` string params (sensible for filter values), so the cursor was never sent. After exempting the cursor param from zero-stripping AND defaulting it to `"0"` when missing, sync succeeded for 11/12 resources, 18 records into local SQLite.

## Root cause

**Profiler:** the candidate name set already exists in the same file (line 656: `"limit", "per_page", "page_size", "count", "first", "max_results"` and line 657: `"after", "cursor", "page_token", "offset", "page", "before", "starting_after"`) — used by `hasRequiredScopeParams` to recognize pagination params for an unrelated check. The profiler simply doesn't reuse it when populating `cursorParams` / `pageSizeParams` from plain `params:` declarations.

[...]

PR #1115, opened 2026-05-11
fix(cli): infer pagination defaults from plain params; preserve cursor=0 in paginatedGet

tmchow:
## Intent

Issue: Refs #927

Internal YAML specs that declare pagination via plain query params (offset, count, limit, etc.) without an explicit `pagination:` block emit broken sync code. The profiler only counts `cursorParams` / `pageSizeParams` when `endpoint.Pagination != nil`, so the global `PaginationProfile` falls back to the literal `after` / `limit` defaults. Generated `sync.go` then requests `?after=…&limit=…` against APIs that expect `?offset=…&count=…`, producing HTTP 406 errors against OneDev, Clerk, and similar APIs.

Separately, `paginatedGet`'s clean loop strips the literal `"0"` for every param. When the cursor flag is an int defaulting to 0, it formats to `"0"`, gets stripped, and offset-paginated APIs reject the request as missing offset.

## Approach

Two narrow changes:

1. **Profiler inference fallback.** When an endpoint has no explicit `pagination:` block, walk `endpoint.Params` and increment `cursorParams` / `pageSizeParams` for non-positional, non-path params whose lowercase name matches the existing candidate sets. Explicit `pagination:` blocks still win because the `Pagination != nil` branch runs first. Lifted the candidate sets to package-level vars so `hasRequiredScopeParams` shares them; behavior of that function is unchanged.

2. **Cursor exempt from zero-strip.** `paginatedGet` keeps the cursor param when its value is `"0"` while still dropping non-cursor `"0"` / `"false"` filter defaults.

## Scope

Primary area: `internal/profiler/profiler.go`, `internal/generator/templates/helpers.go.tmpl`

Why this belongs in this repo: machine-level fix — internal-YAML-authored specs are the dominant style and currently emit silently-wrong pagination param names. Affects every future printed CLI generated from such a spec.

## Risk

- Generated `sync.go` for specs **with** explicit `pagination:` blocks: unchanged. The `Pagination != nil` branch still wins.
- Generated `sync.go` for specs **without** explicit `pagination:` blocks: cursor/limit param names are now inferred. Visible in the goldens (helpers.go).
- Specs with no recognizable pagination params: still default to `after` / `limit` (covered by `TestProfilePagination_NoPaginationParamsKeepsDefaults`).
- `paginatedGet`: cursor=`"0"` now survives the clean loop. Non-cursor filters that format to `"0"` / `"false"` are still stripped (pinned by `TestPaginatedGetExemptsCursorParamFromZeroStripping`).

Known follow-ups (not addressed here):
- Inference walks all endpoints regardless of HTTP method — matches existing explicit-counter behavior; would require a broader change to gate both paths.
- `sync.go.tmpl`'s page loop only sets `params[cursorParam]` when `cursor != ""`. APIs that require an explicit `offset=0` on first page need a separate fix in that loop. Out of scope for this PR.

## Output Contract

Generated `paginatedGet` body changes (4 added lines, 2 modified). Two `testdata/golden/expected/.../internal/cli/helpers.go` fixtures updated via `scripts/golden.sh update` to reflect the new template output.

## Verification

- `go fmt ./...`
- `go build ./...`
- `go vet ./...`
- `go test ./...` — 3799 passed
- `scripts/golden.sh verify` — 17 cases pass after update
- `git diff --check` — clean

### DIFF
--- a/internal/generator/generator_test.go
+++ b/internal/generator/generator_test.go
@@ -3589,6 +3589,22 @@ func TestCompactListFieldsPreservesUnknownShapes(t *testing.T) {
 	}
 }
 
+// The cursor param's "0" must survive paginatedGet's zero-value strip:
+// offset-paginated APIs require offset=0 on the first page request.
+func TestPaginatedGetExemptsCursorParamFromZeroStripping(t *testing.T) {
+	t.Parallel()
+
+	path := filepath.Join("templates", "helpers.go.tmpl")
+	data, err := os.ReadFile(path)
+	require.NoError(t, err, "template must exist: %s", path)
+	body := string(data)
+
+	assert.Contains(t, body, "if k == cursorParam || (v != \"0\" && v != \"false\") {",
+		"paginatedGet must keep the cursor param when its value is \"0\" so offset-paginated APIs receive offset=0 on the first page")
+	assert.NotContains(t, body, "if v != \"\" && v != \"0\" && v != \"false\" {",
+		"the unconditional zero/false strip is wrong for the cursor param — every k must be checked against cursorParam first")
+}
+
 // TestPipedJsonGateRespectsExplicitFormatFlags pins the contract: the
 // piped-output auto-JSON gate must defer to explicit --csv / --quiet /
 // --plain flags so piped consumers that asked for a non-JSON format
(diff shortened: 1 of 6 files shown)


tmchow:
Tightened the paginatedGet template test per Greptile's feedback: assertions are now scoped to the clean-loop block and pin two structural properties (cursorParam reference present; old unconditional zero-strip absent) instead of the exact source expression. Whitespace/reformulation refactors that preserve semantics now pass; logic regressions still fail.

@greptileai

PR #1128, opened 2026-05-12
fix(cli): unwrap RPC envelope arrays in --select for item-level paths

klatt42:
## Intent

Make `--select` ergonomic on RPC-style POST endpoints that wrap items in a common envelope (`results` / `data` / `items` / `hits`) alongside heavy companion fields like full context blobs, pre-fetched prose, or `raw_response`.

Today, an agent calling `--select id,score,source_path` against a `/api/query`-style endpoint gets `{}` back because none of the selected paths match a top-level key (the items live one level down). The fix-by-RTM is dotted paths (`--select results.id,results.score,results.source_path`), but agents reasonably write the simpler form and silently get nothing useful.

Issue: N/A (downstream contribution from a Printing Press user)

## Approach

`filterFields` now checks for a recognized array-envelope before falling back to `filterFieldsRec`. If `data` is an object whose top-level keys don't match any selected path head, and a common envelope key (`results`/`data`/`items`/`hits`) holds an array, descend into that array and apply the filter element-wise. The heavy envelope companion gets dropped — that's where the token savings live.

Dotted paths (`--select results.id`) keep working unchanged, so the escape hatch is preserved when an agent needs the envelope wrap.

The new behavior is guarded by `selectEnvelopeArray`, which returns `(nil, false)` and yields the existing code path when:

- A selected path head matches any top-level key (no need to unwrap)
- No recognized envelope key is present, or no envelope key holds an array

So this is a strict superset of the current behavior — existing `--select` callers see no change.

## Scope

Primary area: `cli` (generator template + emitted printed-CLI behavior)

Why this belongs in this repo: The symptom appeared in a printed CLI (rok-brain-cli against Brain's `/api/query`), but the root cause is the generic `filterFields` helper emitted into every printed CLI. Per AGENTS.md's "default to machine changes" rule, the fix lives in `helpers.go.tmpl`. Generic across any RPC-shape API; no domain-specific naming.

## Risk

Low risk for existing CLIs:

- Strict superset: when no envelope key matches or any path matches a top-level key, behavior is unchanged.
- The four recognized envelope keys (`results`/`data`/`items`/`hits`) are conventional response-list keys; if an API happens to use one of these for an array-valued field that ISN'T the response payload, AND none of the selected paths match a different top-level key, the unwrap would fire. In practice this is the same shape as a real envelope, so the descent is the right call.

Generator-side risk:

- New file emitted (`internal/cli/helpers_test.go`) shifts file counts; bumped `expectedFiles` and added to `mustInclude` in `generator_test.go`.
- Two golden fixtures gain a new file, two helpers.go fixtures show the diff, and `dogfood.json` shifts `dead_functions.total: 56 → 57` (the new `selectEnvelopeArray` function). All captured via `scripts/golden.sh update`.

## Output Contract

Template change to `helpers.go.tmpl` — every regenerated CLI will get the new `selectEnvelopeArray` helper and an extended `filterFields` doc comment. New always-emitted file `internal/cli/helpers_test.go` ships unit tests alongside the helper in every printed CLI.

Golden fixtures updated for both `generate-golden-api` and `generate-golden-api-rich-auth`:
- `helpers.go` gets `+49 lines` (new function + wiring + doc comment)
- New `helpers_test.go` artifact (~140 lines, 5 table-driven test cases)
- `dogfood.json` `dead_functions.total: 56 → 57`

## Verification

- `go test ./...` — full suite green (29 packages, 0 fails)
- `go vet ./...` — clean
- `go fmt ./...` — clean
- `scripts/golden.sh verify` — 17 cases pass after `scripts/golden.sh update` (run with `GOLDEN_ALLOW_DIRTY=1` since code + goldens land in one atomic commit)
- New structural test `TestFilterFieldsEnvelopeUnwrap_EmittedInHelpers` passes
- New behavior tests in `helpers_test.go.tmpl` cover: envelope unwrap with heavy field dropped, dotted-path escape hatch, top-level-match guard, no-recognized-envelope fallback, all four envelope keys
- `golangci-lint`: not run locally (not installed); pre-push and CI will catch any issues. `go vet` is clean.

## AI / Automation Disclosure

## Discovery Context

This dropped out of a token-spend A/B test on a generated CLI (Brain's RAG API). Pre-fix, the prompt "find 3 docs about RRF" against Brain's `/api/query` had three options:
1. `curl` + JSON parse → 5.6 KB response into the agent's context (includes the `context` field — ~3 KB of pre-fetched markdown that's the dominant token sink).
2. CLI without `--select` → same payload, same token cost.
3. CLI with `--select id,score,source_path` → `{}`. Useless.

Post-fix, option 3 returns `[{id,score,source_path},...]` — 469 bytes. **11.9× payload reduction on the dominant RAG-query shape.**

Happy to split this differently if you'd prefer (e.g. land the template change first and the test template second), or scope down the goldens churn. Let me know what's most useful.


tmchow:
@klatt42 address code review finding

klatt42:
@tmchow addressed both Greptile P2 findings in 9b7b5c87:

**P2-a — weak fallback assertion**: `TestFilterFields_NoEnvelopeKey_FallsBackToObjectFilter` now asserts the output is exactly `{}` instead of just "starts with `{`". The new form fails loudly in both directions — if the unwrap incorrectly fires on the non-envelope `records` key it would return the unwrapped array, and if the filter is skipped entirely it would return the original object. Either bug now surfaces immediately.

**P2-b — case-sensitivity asymmetry**: `selectEnvelopeArray`'s envelope lookup is now case-insensitive (iterates `obj` and lowercases keys), matching the case-insensitive guard above and the broader `filterFields` convention (which lowercases path segments via `matchSelectSegment`). `Results`, `RESULTS`, `Hits` now trigger the unwrap the same way `results` does. Test coverage extends `DataEnvelope_UnwrapsAndNarrows` to exercise all four envelope keys in both canonical and mixed/uppercase forms (8 subtests).

Verified locally:
- `go test ./...` — full upstream suite green (29 packages)
- `scripts/golden.sh verify` — 17 cases pass after `GOLDEN_ALLOW_DIRTY=1 scripts/golden.sh update`
- Generated a fresh CLI from the updated template and ran `go test ./internal/cli` — all 12 emitted test cases (4 behavior tests + 8 case-variant subtests) pass

CI is still waiting on workflow approval for the action_required runs.

klatt42:
Rebased onto current upstream/main (post-#1194 merge + 4.5.1 release). Conflicts resolved:

- `internal/generator/generator.go` — singleFiles map now contains both `helpers_test.go.tmpl` (this PR) and `root_test.go.tmpl` (#1194).
- `internal/generator/generator_test.go` — `expectedFiles` bumped to `60 / 65 / 62` (cumulative +2 from main's original 58/63/60 to account for both PRs' emitted test files). `mustInclude` now contains both `internal/cli/helpers_test.go` and `internal/cli/root_test.go`.
- `testdata/golden/expected/generate-golden-api/dogfood.json` — `dead_functions.total` resolved to 59 (cumulative +2: `isCobraUsageError` from #1194 + `selectEnvelopeArray` from this PR), then re-verified via `scripts/golden.sh update` + `verify`.

Verified locally:
- `go test ./...` — full suite green (29 packages)
- `scripts/golden.sh verify` — 17 cases pass

Branch is now at `a9d42bb0` + `8e179d10` on `klatt42:cli/filter-envelope-unwrap`. Ready for CI workflow approval.

klatt42:
Closing in favor of @tmchow's #1379, which shipped the same envelope-unwrap behavior with a strictly better approach.

**Why theirs is better:**
- **Generic vs allowlist**: my PR hardcoded the envelope-key set (`results` / `data` / `items` / `hits`) via a separate `selectEnvelopeArray` helper. Theirs uses "any sibling array" detection, so it handles `{"projects":[...]}`, `{"object":"list","results":[...]}`, `{"total_count":N,"items":[...]}`, and any other shape without enumeration.
- **Preserves envelope metadata**: theirs keeps non-array siblings (counts, pagination cursors) visible — useful info that mine would have stripped.
- **Lives in the existing recursion** (`filterFieldsRec`) rather than adding a new helper; smaller surface, less code to maintain.

The discovery context (Brain's RAG `/api/query` returning `{}` for item-level selects) still stands as the motivating case — both approaches solve it. Glad the generic version landed; will benefit any list-envelope shape, not just RAG.

Closing this PR + deleting the branch.

PR #1167, opened 2026-05-12
fix(cli): preserve novel-command keys in --compact partial-strip path

### LINKED ISSUE #1127: generator: compactListFields allow-list strips hand-written novel-command output keys
## Summary

The framework's `compactListFields` (and the broader `--agent` projection path) uses a hard-coded allow-list of "known" keys (id, name, status, timestamps). Hand-written commands that emit different keys silently get their output projected to `[{}]` empty rows.

The polish skill discovered this on the unify run for all 9 novel transcendence commands (`search`, `sql`, `coverage`, `audit-scores`, `schema diff`, `vet`, `import-csv`, `trace`, `watch`). `--agent` (which implies `--compact`) returned empty rows for every one of them. Polish's per-CLI fix expanded the allow-list to cover unify-specific keys (`object_name`, `match_key`, `snippet`, `exists`, `record_id`, etc.) and added a non-empty-row fallback. The same fix should ship in the framework template so polish doesn't have to apply it per CLI.

## Scope

Recurs for every printed CLI that ships novel commands not derived from the spec's resource fields. Phase 1.5 transcendence is required for ship, so this hits effectively every CLI that survives shipcheck. Catalog candidates: `recipe-goat`, `weather-goat`, `flightgoat`, `company-goat`, and any CLI with hand-written transcendence commands.

## Proposed fix

Two paths:

1. **Data-driven keys (preferred):** `compactListFields` keeps keys present in 80%+ of input rows rather than checking against a hard-coded allow-list. The semantic intent of `--compact` (return only "key fields") is preserved by frequency, not by name lookup.
2. **Additive allow-list (minimum):** Keep the allow-list but add any keys observed in the input rows that have non-zero values. Plus a non-empty-row fallback: if the projected row would be `{}`, return the original row instead.

Polish's empirical fix in unify implemented both halves of (2); shipping that as the template baseline is the floor.

## Acceptance

- **Positive:** Generate a CLI with a hand-written command emitting `{"foo": "x", "bar": "y"}`. Run `<cli> <cmd> --compact` and `<cli> <cmd> --agent` and assert the output contains the keys (or at minimum, that `-z$(<cli> <cmd> --agent | jq '.[0]')` is empty — meaning the row is non-empty).
- **Negative:** Endpoint-mirror commands continue to produce the existing projection (no regression on the spec-derived flow).

**Artifacts:**
- 📄 [Retro document](https://files.catbox.moe/x8gizy.md) — full triage rationale, drops, skips, what went right
- 📦 [Manuscripts](https://files.catbox.moe/syft99.zip) (37K) — research brief, absorb manifest, shipcheck/dogfood/polish proofs
- 💻 [CLI source](https://files.catbox.moe/clyzog.zip) (11M, no binary/vendor) — generated Go code

> Filed by `/printing-press-retro` from a run of **unify-pp-cli** (scorecard 75/100, verify 100%, full live dogfood 102/102 against the user's real workspace).

tmchow:
## Summary

Closes #1127.

`compactListFields` had a partial-strip bug: if a row contained even one static allow-list key (e.g. `id`), the rest of the row's keys were stripped, even when they carried real data the agent needed. Example from issue #1127:

```
input:  {"id":"x","object_name":"users","match_key":"uuid","snippet":"..."}
before: {"id":"x"}            // object_name, match_key, snippet silently dropped
after:  {"id":"x","object_name":"users","match_key":"uuid","snippet":"..."}
```

The fix adds a data-driven extension on top of the existing static allow-list: any scalar key present in 80%+ of input rows is kept too. The verbose blocklist (`description`, `body`, `content`, `comments`, `attachments`, `html`, `markdown`) is excluded from the extension regardless of frequency, so the compact intent — short identifying values for agent consumption, not full prose — is preserved. PR #1078's per-item fallback still fires when no key matches at all.

Side cleanups in the same diff:
- Pulled the verbose blocklist into a package-level `compactVerboseFields` shared by `compactListFields` and `compactObjectFields` (was duplicated as `verboseFields` / `stripFields`).
- `printAutoTable`'s inline scalar type-switch now reuses the new `isCompactScalar` predicate.

## Scope

Machine change — `internal/generator/templates/helpers.go.tmpl`. Rendered into every printed CLI's `internal/cli/helpers.go`. MCP endpoint mirrors bypass `compactFields` (verified at `internal/generator/templates/mcp_tools.go.tmpl` per PR #1078), so this affects the CLI surface for both human users and agents driving the cobratree walker. Endpoint-mirror typed MCP tools are unchanged.

## Behavior table (per row, single-row response unless noted)

| Input | Before | After |
|-------|--------|-------|
| `{"id":"x","object_name":"users","match_key":"uuid"}` | `{"id":"x"}` (partial strip) | full row (frequency=100% on scalar keys) |
| `{"airline":"JL","fare":12345,"currency":"USD","destination":"NRT"}` | `{"currency":"USD","fare":12345}` (allow-list only) | full row (all scalar, all 100%) |
| `{"id":7,"name":"W","description":"long blurb","price":9.99,"rating":4.5}` | `{"id":7,"name":"W","price":9.99,"rating":4.5}` | unchanged — `description` still stripped by verbose blocklist |
| `{"some_obscure_field":"v","another":42}` | original (fallback fires) | original (now via 100% frequency, same shape) |
| List with mostly-nested rows | nested objects stripped via allow-list | unchanged — `isCompactScalar` excludes maps/slices from the extension |

## Test plan

- [x] `go fmt ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./...` — 3801 passing across 34 packages
- [x] `golangci-lint run ./...` — no issues
- [x] `scripts/golden.sh verify` — 17 cases pass after regenerating goldens for the template change
- [x] `TestCompactListFieldsPreservesUnknownShapes` extended with assertions for the new symbols (`keyCounts`, `isCompactScalar`, `compactVerboseFields`, `threshold`)

## Related work

- Builds on #1078, which added the per-item fallback for "row has zero allow-list keys."
- The class of bug (narrow classifier with asymmetric failure cost — false-drop > false-keep) is documented in `docs/solutions/design-patterns/avoid-classification-when-failure-is-asymmetric-2026-05-06.md`. The frequency-rule fix matches the documented "structural detection over enumerative allow-list expansion" pattern.

### DIFF
--- a/internal/generator/generator_test.go
+++ b/internal/generator/generator_test.go
@@ -3556,11 +3556,17 @@ func TestGeneratedOutput_MutatingCommandsHaveEnvelope(t *testing.T) {
 	assert.Contains(t, content, `envelope["success"] = false`)
 }
 
-// TestCompactListFieldsPreservesUnknownShapes pins the two contracts that
-// keep `--agent` / `--compact` from silently emitting {} for records that
-// don't match the canonical id/name/title allowlist: a per-item fallback
-// to the original item, and a wider scalar allowlist that covers
-// monetary, metric, locale, and identity-adjacent keys.
+// TestCompactListFieldsPreservesUnknownShapes pins the contracts that keep
+// `--agent` / `--compact` from silently emitting {} or partial-strip rows
+// for records that don't match the canonical id/name/title allowlist:
+//
+//  1. The static allow-list covers canonical scalars (price/fare/locale/code/...).
+//  2. A data-driven extension keeps any scalar key present in 80%+ of rows,
+//     so hand-written novel commands whose output keys aren't on the
+//     allow-list (object_name, match_key, snippet) survive projection.
+//  3. Verbose fields (description, body, ...) are excluded from the
+//     extension regardless of frequency.
+//  4. A per-item fallback preserves the original item when no key matches.
 //
 // Pinned at the template level because the helper renders into every
 // printed CLI's helpers.go and a per-fixture assertion would miss future
@@ -3574,7 +3580,7 @@ func TestCompactListFieldsPreservesUnknownShapes(t *testing.T) {
 	body := string(data)
 
 	assert.Contains(t, body, "if len(compact) == 0 {",
-		"compactListFields must preserve the original item when no allowlist keys match — otherwise records with domain-specific field names get reduced to {}")
+		"compactListFields must preserve the original item when no keep keys match — otherwise records with domain-specific field names get reduced to {}")
 	assert.Contains(t, body, "compact = item",
 		"compactListFields must assign the original item as the per-item fallback")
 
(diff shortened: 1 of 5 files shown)


tmchow:
Re Greptile's two follow-up P2s on the 5/5 re-review:

1. **Filter loop doesn't re-check `isCompactScalar`** — accepted-for-later. The data-driven extension only adds keys that appear as scalars in ≥80% of rows, so the value will be scalar in the typical case. The corner case Greptile names (a key that's scalar in 80% of rows but complex in the minority) emits the complex value rather than dropping it; preserving data here matches the issue's bias against silent partial-strip. Worth revisiting if it surfaces a real complaint.
2. **Tests pin symbols, not runtime behavior** — accepted-for-later. The rendered helper is locked by golden fixtures and the symbol-presence test; a behavioral test would require either copying the function into a Go test file (drift risk) or extracting it from the template into a sibling package. Tracking as a separate follow-up rather than expanding this PR's scope.

PR #1186, opened 2026-05-12
fix(cli): refuse to ship CLIs with placeholder base URL

### LINKED ISSUE #1012: Detect placeholder base URL and refuse to ship
## Goal
Generation fails fast with a clear error when the resolved base URL is a known placeholder (`api.example.com`, `example.com`, `example.org`, `localhost`, `localhost:*`, or unresolved relative path).

## Problem
The OpenAPI 3 default-`servers` fallback (`[{ url: "/" }]` → resolved against the spec's loaded URL) can resolve to the literal example.com placeholder when the spec omits `servers:`. PP silently writes that placeholder to `config.toml` and ships. `doctor` then fails immediately with `lookup api.example.com: no such host`. The shipped CLI has no working base URL out of the box.

## Evidence
- Run: `selling-partner-api-pp-cli` (Catalog Items spec, May 2026)
- Spec target: `/tmp/selling-partner-api-models/models` (any spec from `amzn/selling-partner-api-models`)
- Printed config: `base_url: https://api.example.com`
- Doctor output:
  ```
  OK Config: ok
  OK Auth: not required
  FAIL API: unreachable: GET /: Get "https://api.example.com/": dial tcp: lookup api.example.com: no such host
  ```

## Step B (cross-API evidence)
Honestly weak — only one API named with hard evidence (SP-API Catalog Items). All 40 SP-API specs in the same repo would hit the same bug (they all omit standard `servers:`). Outside SP-API: speculative. Filing as P2 reflects this.

## Step C (counter-check)
The fix is a placeholder-domain guard. example.com / example.org are reserved by RFC 2606 for documentation — no legitimate spec ships pointing there. localhost / 127.0.0.1 / unresolved-relative also have no legitimate ship-state use. Safe.

## Durable fix
In `internal/openapi/` (wherever the parser resolves `servers[].url` into `base_url`):
1. Check whether the resolved URL is in `{api.example.com, example.com, example.org, localhost, 127.0.0.1, unresolved-relative-path}`.
2. If yes, refuse to ship. Emit a clear error naming the spec file and instructing the user to specify `--base-url` or fix the spec.

## Acceptance criteria
- **Positive:** feeding a spec whose `servers:` is missing or resolves to `api.example.com` halts generation with a message naming the spec file and instructing the user to specify `--base-url` or fix the spec.
- **Negative:** feeding a spec with `servers: [{ url: "https://api.real.com" }]` proceeds normally.

## Scope boundary
Does not include auto-detecting the real base URL from README or vendor extensions — that's a follow-up.

## Complexity
small

## Related context
- Sibling retro WUs in the same run: WU-2 (vendor-extension auth detection), WU-3 (multi-spec directory navigation). Same SP-API run surfaced all three; this is the smallest/most-isolated of the three and could land independently.

tmchow:
## Summary

The OpenAPI, GraphQL, and docspec parsers all fall back to `https://api.example.com` when the source declares no real base URL. Previously the generator wrote that placeholder into `config.toml` and shipped a CLI whose `doctor` would DNS-fail on every call.

This change:

- Adds a shared `spec.PlaceholderBaseURL` constant so all three parsers reference one sentinel.
- Tracks the fallback via a new non-serialized `BaseURLIsPlaceholder` flag on `spec.APISpec`. All three parsers mark the flag when they fall back.
- The generate command refuses to proceed when the flag is set, in both the OpenAPI/GraphQL spec-parsing loop and the `--docs` path.
- The LLM doc-scrape path re-detects the placeholder after `spec.ParseBytes` since `yaml:"-"` strips the flag on the marshal/unmarshal round-trip — without this, an LLM that echoes the prompt's example `base_url` would bypass the refusal.

## Test plan

- [x] New parser tests (openapi, graphql, docspec) verify the flag is set on fallback and unset when a real host is resolved.
- [x] New `TestGenerateCmdRefusesPlaceholderBaseURL` asserts end-to-end refusal via the `generate` command, including that no output directory is written.
- [x] `go test ./...` (3820 tests pass).
- [x] `scripts/golden.sh verify` (17 cases pass).
- [x] `go vet ./...` and `golangci-lint run ./...` clean.

Closes #1012

### DIFF
--- a/internal/cli/generate_test.go
+++ b/internal/cli/generate_test.go
@@ -1579,3 +1579,42 @@ func runGoCommandForCLITest(t *testing.T, dir string, args ...string) {
 	out, err := cmd.CombinedOutput()
 	require.NoError(t, err, string(out))
 }
+
+// TestGenerateCmdRefusesPlaceholderBaseURL pins the contract that the
+// generate command halts when the spec declares neither `servers:` nor
+// any per-operation server. The parser falls back to a fake hostname so
+// the spec keeps parsing, but the generate command must refuse to write
+// a CLI whose `doctor` would DNS-fail on every call.
+func TestGenerateCmdRefusesPlaceholderBaseURL(t *testing.T) {
+	t.Parallel()
+
+	dir := t.TempDir()
+	specPath := filepath.Join(dir, "spec.yaml")
+	outputDir := filepath.Join(dir, "noserversapp")
+	require.NoError(t, os.WriteFile(specPath, []byte(`openapi: "3.0.3"
+info:
+  title: No Servers App
+  version: "1.0"
+paths:
+  /things:
+    get:
+      operationId: listThings
+      responses:
+        '200':
+          description: OK
+`), 0o644))
+
+	cmd := newGenerateCmd()
+	cmd.SetArgs([]string{
+		"--spec", specPath,
+		"--output", outputDir,
+		"--validate=false",
+		"--force",
+	})
+
+	err := cmd.Execute()
+	require.Error(t, err, "expected generate to refuse a spec without servers")
+	assert.Contains(t, err.Error(), specPath, "error must name the offending spec file")
+	assert.Contains(t, err.Error(), "no `servers:`", "error must explain that the spec declares no servers")
+	assert.NoDirExists(t, outputDir, "refusal must fire before any output is written")
+}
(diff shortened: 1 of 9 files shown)


tmchow:
Greptile findings addressed:

- **Hardcoded placeholder string in tests** — fixed in the follow-up commit. Tests now reference `spec.PlaceholderBaseURL` directly.
- **No end-to-end test for the `--docs` refusal path** — accepted for follow-up. The unit-level coverage (parser test + the post-`spec.ParseBytes` re-detection logic) is in place; end-to-end coverage would need HTTP/LLM mocking infrastructure not currently in the cli-package tests. Tracking as a separate gap.

PR #1193, opened 2026-05-12
fix(cli): guard nil Phases in LoadState to prevent lock promote panic

### LINKED ISSUE #1189: lock promote panics (assignment to entry in nil map) on second run — LoadState migration loop hits a nil Phases map
## Summary
`printing-press lock promote` panics with `panic: assignment to entry in nil map` when run a **second** time for the same CLI.

```
panic: assignment to entry in nil map
goroutine 1 [running]:
.../internal/pipeline.LoadState({...})  internal/pipeline/state.go:510
.../internal/pipeline.NewMinimalState(...)  internal/pipeline/state.go:366
.../internal/cli.newLockPromoteCmd.func1(...)  internal/cli/lock.go:238
```

## Root cause
- The first `lock promote` rewrites `state.json` with `working_dir` pointed at the **library** path (not the original working dir) and `phases: null`.
- A subsequent `lock promote --dir <original-workdir>` no longer matches via `FindStateByWorkingDir`, so `lock.go:238` falls back to `pipeline.NewMinimalState`, which calls `LoadState(apiName)`.
- `LoadState` unmarshals the on-disk `state.json` → `Version: 0`, `Phases: nil` (from `"phases": null`). Because `0 < currentStateVersion (3)`, the version-migration loop around `state.go:505-525` does `s.Phases[name] = PhaseState{...}` for each `name` in `PhaseOrder` — **assigning into a nil map → panic**.
- `NewMinimalState` itself initializes `Phases` to a non-nil map, but the `LoadState` it then calls runs the migration loop on the *loaded* (nil-map) state.

## Reproduction
1. Write a `state.json` containing `{"version":0,"phases":null, ...}` (this is what a prior `lock promote` leaves on disk).
2. `printing-press lock promote --cli <name>-pp-cli --dir <workdir>` — succeeds.
3. Run the same `lock promote` again — panics.

It also triggers on the main `/printing-press` skill's Phase 6 hold-path, which re-invokes `lock promote` after a "Polish to retry" pass, and on the natural flow of shipping a post-promote fix into the library.

## Proposed fix (small)
In `internal/pipeline/state.go`, after the unmarshal in `LoadState` and **before** the version-migration loop:
```go
if s.Phases == nil {
    s.Phases = make(map[string]PhaseState)
}
```
Belt-and-suspenders: also have `pipeline.PromoteWorkingCLI` / `lock promote`'s state-rewrite path persist `version: currentStateVersion` and `phases: {}` rather than `phases: null`.

## Acceptance
- A `state.json` with `{"version":0,"phases":null,...}` loads without panic; `lock promote --dir <dir>` run twice both succeed.
- A current-version `state.json` with a populated `phases` map still promotes correctly; the migration loop stays a no-op and doesn't clobber existing phase state.

## Context
Surfaced during a `/printing-press` run for `bing-webmaster-tools`. Worked around by copying the changed files into the library dir and rebuilding the binary directly (loses the manifest / CurrentRunPointer update `lock promote` does). No prior retro raised this. Full retro: `~/printing-press/manuscripts/bing-webmaster-tools/20260511-210547/proofs/20260512-195341-retro-bing-webmaster-tools-pp-cli.md`.

Filed by `/printing-press-retro`. Component note: the fix lands in `internal/pipeline/state.go`; `comp:generator` is the closest label in the fixed set (no `pipeline`/`lock` slug exists).

tmchow:
## Intent

`printing-press lock promote` panics with `panic: assignment to entry in nil map` on the second run for the same CLI. The first promote rewrites `state.json` with `working_dir` pointed at the library path and `phases: null`; a subsequent `lock promote --dir <orig-workdir>` no longer matches via `FindStateByWorkingDir`, falls back to `NewMinimalState`, and triggers `LoadState`. `LoadState` unmarshals `phases: null` into a nil map and the version-migration loop then assigns into it.

Issue: Closes #1189

## Approach

Initialize `s.Phases` to a non-nil map at the top of `LoadState` before the version-migration loop runs. Belt-and-suspenders: initialize it in `save()` too so any code path that constructs `PipelineState{}` directly never serializes `phases: null` back to disk.

Both guards are small and idempotent. The migration loop is unchanged, so existing v1->v3 backfill semantics are preserved.

## Scope

Primary area: `internal/pipeline/state.go` — pipeline state load/save.

Why this belongs in this repo: machine change. The panic occurs in the generator/binary's pipeline package, blocks `lock promote` re-runs, and surfaced from a real run for `bing-webmaster-tools`. It is not API-specific.

## Risk

- The version-migration loop semantics are unchanged. The new guard initializes the map only when it is nil after unmarshal — when `phases` is an object (the common case), the guard is a no-op and the loop continues to backfill missing phases for `Version < currentStateVersion`.
- The `save()` guard initializes Phases only when it is nil on the in-memory state — the common construction paths (`NewState`, `NewStateWithRun`, `NewMinimalState`) already populate Phases, so the guard is a no-op there.
- No on-disk format change. JSON shape for `phases` now serializes as `{}` instead of `null` when the in-memory map was nil, but existing consumers either fall through the migration loop or iterate via the populated map.
- AGENTS.md notes `docs/PIPELINE.md` should be updated when `internal/pipeline/state.go` changes. This fix is a defensive guard with no contract impact — phase names, ordering, transitions, and the migration story are unchanged. I did not update PIPELINE.md; happy to add a note if a reviewer disagrees.

## Output Contract

N/A — no template, generated file, manifest, or scorecard surface changed.

## Verification

- `go test ./...` — 3859 pass, including new `TestLoadStateHandlesNullPhases` (subtests for `version: 0` and `version: currentStateVersion`) and `TestSavePersistsEmptyPhasesAsObject`.
- `go vet ./...` — clean.
- `golangci-lint run ./...` — clean.
- `scripts/golden.sh verify` — all 17 cases pass.
- `git diff --check` — clean.

## AI / Automation Disclosure

### DIFF
--- a/docs/PIPELINE.md
+++ b/docs/PIPELINE.md
@@ -39,6 +39,8 @@ Every phase has two orthogonal status fields:
 
 `Status` tracks the execution state of the phase. `PlanStatus` tracks the state of the phase's plan file: a freshly-initialized phase has a `seed` plan; once expanded by the planner it becomes `expanded`; once the phase finishes it becomes `completed`. Both are defined in `internal/pipeline/state.go`.
 
+`LoadState` and `save()` defensively initialize the on-disk `phases` map to an empty object when it is unmarshaled as nil. Phase names, ordering, transitions, and the version-migration semantics described in this document are unaffected by that guard.
+
 ## Phases
 
 Each phase below lists four fields: what it consumes, what it produces, what gates it must pass, and which artifacts it leaves on disk.
--- a/internal/pipeline/state.go
+++ b/internal/pipeline/state.go
@@ -416,6 +416,9 @@ func NewStateWithRun(apiName, outputDir, runID, scope string) *PipelineState {
 }
 
 func (s *PipelineState) save(updateCurrentPointer bool) error {
+	if s.Phases == nil {
+		s.Phases = make(map[string]PhaseState)
+	}
 	if s.Scope == "" {
 		s.Scope = WorkspaceScope()
 	}
@@ -480,6 +483,11 @@ func LoadState(apiName string) (*PipelineState, error) {
 	}
 
 	needsSave := false
+	// Must run before the version-migration loop below, which assigns into s.Phases.
+	if s.Phases == nil {
+		s.Phases = make(map[string]PhaseState)
+		needsSave = true
+	}
 	if s.RunID == "" {
 		runID, err := newRunID(time.Now())
 		if err != nil {
--- a/internal/pipeline/state_test.go
+++ b/internal/pipeline/state_test.go
@@ -2,6 +2,7 @@ package pipeline
 
 import (
 	"encoding/json"
+	"fmt"
 	"os"
 	"path/filepath"
 	"testing"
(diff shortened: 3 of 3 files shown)


tmchow:
@mergify queue

tmchow:
@mergifyio rebase

PR #1194, opened 2026-05-12
fix(cli): map Cobra/pflag pre-RunE usage errors to exit code 2

klatt42:
## Intent

Honor the POSIX exit-code convention already documented in `helpers.go`: usage errors exit `2`, not `1`. Cobra and pflag generate usage errors (unknown flag, unknown command, missing required flag, missing flag value, invalid argument) inside `rootCmd.Execute()` before any user `RunE` runs, so those errors never flow through `usageErr()` — `ExitCode()` falls through to the default and emits `1`, silently breaking the contract every printed CLI claims.

Issue: N/A (downstream finding from the rok-brain-cli pilot)

## Approach

`isCobraUsageError(err)` detects the six pre-RunE error shapes by message prefix/substring (Cobra and pflag don't export typed sentinels for these — same approach the existing "unknown flag" hint path already uses).

`Execute()` routes matching errors through `usageErr` before returning. The existing unknown-flag hint-suggestion path was switched from `return fmt.Errorf(...)` to `err = fmt.Errorf(...)` so the rewrapped error falls through to the new usage wrap rather than short-circuiting it; the emitted test pins that switch and confirms hint-rewrapped errors still classify as usage errors after the rewrap.

Strict superset: non-usage errors and `nil` returns are unchanged.

## Scope

Primary area: `cli` (generator template — every printed CLI is affected)

Why this belongs in this repo: the bug is in `root.go.tmpl`'s `Execute()` + `ExitCode()` contract, which is generator-emitted boilerplate shared by every printed CLI. Per AGENTS.md's machine-first rule, the fix lives in the template.

## Risk

Low. Strict superset:

- `nil` and non-matching errors are unchanged in shape and exit code.
- The hint-suggestion path's user-visible output (the `hint: did you mean --foo?` line) is unchanged.
- Detection patterns match Cobra v1.x + pflag v1.x as of 2026-05; if a future Cobra release changes wording, the pattern fails closed (falls through to exit 1, same as today) rather than mis-classifying a real error.

## Output Contract

Template change to `root.go.tmpl` — every regenerated CLI gets `isCobraUsageError` and the usage wrap in `Execute()`. New always-emitted file `internal/cli/root_test.go` ships unit tests alongside `root.go` in every printed CLI.

Golden fixtures updated for both `generate-golden-api` and `generate-golden-api-rich-auth`:
- `root.go` gains the helper + wire-in (+50 lines)
- New `root_test.go` artifact (~70 lines, 11 test cases)

## Verification

- `go test ./...` — full upstream suite green (29 packages, 0 fails)
- `go vet ./...` — clean
- `go fmt ./...` — clean
- `scripts/golden.sh verify` — 17 cases pass after `GOLDEN_ALLOW_DIRTY=1 scripts/golden.sh update`
- New structural test `TestUsageErrorExitCode_EmittedInRoot` passes
- New behavior tests in `root_test.go.tmpl`: 9 subtests for `isCobraUsageError` shape detection, 1 for hint-rewrap pass-through, 1 for end-to-end `ExitCode(usageErr(...)) == 2`
- **End-to-end verified on a freshly generated CLI**:
  - `./loops-pp-cli --foobar` → exit `2` (was `1`)
  - `./loops-pp-cli notarealsubcommand` → exit `2` (was `1`)
  - `./loops-pp-cli --help` → exit `0` (unchanged)
- `golangci-lint`: not run locally (not installed); pre-push and CI will catch any issues. `go vet` is clean.

## File-count overlap with #1128

This PR bumps `expectedFiles` `+1` vs `main` (58→59 stytch, 63→64 clerk, 60→61 loops) and adds `internal/cli/root_test.go` to `mustInclude`. PR #1128 (currently open) makes the same `+1` bump for `internal/cli/helpers_test.go`. Whichever lands first leaves the second to trivially rebase to `+2`. I'll handle the rebase on whichever PR is second to merge.

## AI / Automation Disclosure

## Discovery Context

PILOT_NOTES from the rok-brain-cli pilot called this out as "plan expected 2 per Cobra conventions" (sharp edge #4). The plan was right — the convention is in `helpers.go`'s `usageErr` already, just bypassed by Cobra's own error path. Quick patch with measurable user-visible impact: bash test harnesses, CI scripts, and agent error-classification logic that key on the conventional exit-2 will now actually fire on bad flags.


klatt42:
Addressed Greptile P2 in e91ef496. The three substring patterns for `required flag`, `flag needs an argument`, and `invalid argument` are now anchored to the literal punctuation Cobra/pflag emit:

- `required flag(s) "foo" not set` → `HasPrefix("required flag(s) \"")`
- `flag needs an argument: --foo` → `HasPrefix("flag needs an argument:")`
- `invalid argument "x" for "--y" flag:` → `HasPrefix("invalid argument \"")`

Existing positive test cases (using real Cobra/pflag error strings) continue to pass. Three new negative cases lock the tightness — application errors that happen to mention "required flag" / "flag needs an argument" / "invalid argument" as prose now correctly classify as non-usage errors and retain their original exit code.

Verified locally:
- `go test ./internal/generator` — green
- Fresh CLI generation → 12 subtests pass (9 original + 3 new negative cases)
- `scripts/golden.sh verify` — 17 cases pass after `GOLDEN_ALLOW_DIRTY=1 scripts/golden.sh update`

PR #1204, opened 2026-05-12
fix(cli): support $-prefixed pagination params for Socrata-style APIs

youdidwhat-towho:
## Problem

Press's generator emits paginated-GET handlers that put each Param's `Name` directly into the URL query string. Socrata-backed APIs (NYC OpenData and ~70 other major US municipal portals) require a literal `$` prefix on pagination + SoQL params:

- `$limit`, `$offset` (pagination)
- `$where`, `$select`, `$order`, `$group` (SoQL passthrough)

Generated CLIs against Socrata returned `HTTP 400 Unrecognized arguments [limit]` on every live call.

## Fix (Option 2 — per-param `url_name` override)

Adds an optional `url_name` field to `spec.Param` that overrides the URL query-key emission while leaving the CLI flag name, Go identifier, and JSON body emission unchanged.

```yaml
params:
  - name: limit
    url_name: "$limit"   # NEW — optional
    type: integer
```

`--limit` stays the CLI flag, `flagLimit` stays the Go identifier, but the URL emits `?$limit=N`.

## Why this approach (vs auto-detect)

Auto-detecting `*.socrata.com` / `*.cityofnewyork.us` would couple the generator to URL patterns, which doesn't scale to other custom-prefix APIs. A spec-level opt-in is non-breaking and explicit — existing specs work unchanged, new Socrata specs declare the prefix where they declare every other param attribute.

## What's NOT touched

- Path substitution, JSON body keys, MCP tool bindings still use `Name`. `url_name` affects URL query keys only.
- Pagination's `limit_param` / `cursor_param` already accepted user-supplied strings, so spec-level pagination config already worked. This fixes the remaining gap for non-pagination SoQL params (`$where`, `$select`, etc.) where the user declares them as ordinary params.
- ArcGIS, REST APIs, GraphQL — anything without `url_name` declared keeps emitting plain `Name`.

## Tests

New file `internal/generator/url_name_test.go`:

- `TestParamWireNameUnit` — `spec.Param.WireName()` returns URLName when set, Name when empty
- `TestParamURLNameOverridesWireKey` — generated handler emits `params["$limit"]` for a param with `url_name: "$limit"`, plain `params["borough"]` for a sibling param without override, and CLI flag identifiers stay clean (no `$` leak into Go)
- `TestParamWithoutURLNameUnchanged` — regression guard: specs without `url_name` keep plain `params["limit"]`, `params["owner"]` emission

Full `internal/spec/` + `internal/generator/` test suite green:

```
ok  internal/spec      0.555s
ok  internal/generator 82.551s
```

## Live verification

**Before** (working around the bug by stripping `limit` from the spec):
```bash
nyc-liens-pp-cli records query --borough 3 --water-debt-only YES
# HTTP 400 if any of limit / where / select supplied
```

**After** (with `url_name` declared):
```bash
nyc-liens-pp-cli records query --borough 3 --water-debt-only YES --limit 2
# Returns 2 valid Brooklyn water-only lien records, structured JSON
```

**Regression check** — `bexar-cad-pp-cli` (ArcGIS, no `url_name` anywhere):
```bash
bexar-cad-pp-cli parcels query --where "Owner LIKE '%PULTE%'" --result-record-count 1
# Returns live PULTE HOMES TX data, unchanged
```

Inspected generated handler for bexar-cad-test confirms plain emission preserved:
```go
params["f"] = fmt.Sprintf("%v", flagF)
params["outFields"] = fmt.Sprintf("%v", flagOutFields)
params["geometry"] = fmt.Sprintf("%v", flagGeometry)
```

And for nyc-liens-test (mixed plain + url_name):
```go
"borough":         fmt.Sprintf("%v", flagBorough),         // plain
"water_debt_only": fmt.Sprintf("%v", flagWaterDebtOnly),   // plain
"$where":          fmt.Sprintf("%v", flagWhere),           // url_name override
"$select":         fmt.Sprintf("%v", flagSelect),          // url_name override
"$limit":          fmt.Sprintf("%v", flagLimit),           // url_name override
```

## Diff size

6 files, +166 / -16. Five files modified, one new test file.

## Follow-up — NOT addressed in this PR

Press's `id_field` extraction fails on Socrata because Socrata only returns its `:id` row identifier when explicitly `$select`-ed. Generated sync runs produce `all_items_failed_id_extraction` warnings on Socrata datasets. The clean fix is to detect `id_field: ":id"` in the spec (or any field starting with `:`) and auto-inject `$select=:id,*` (or the union of explicit `$select` + `:id`) into the URL. Filing as a separate PR to keep this one scoped.

## Unlocks

This change unblocks every Socrata-backed CLI Factory target:
- NYC: ACRIS, PLUTO, DOF Property Charges Balance, HPD Violations, DOB Violations, Tax Lien Sale Lists, etc. (600+ datasets)
- Cook County / Chicago, Boston, SF, Seattle, Austin, Denver — every major US city on Socrata

### DIFF
--- a/internal/generator/flag_collision.go
+++ b/internal/generator/flag_collision.go
@@ -304,13 +304,21 @@ func validatePublicFlagEntry(resKey, epName string, entry publicFlagEntry, reser
 // identifiers (via camel) or cobra flag names (via flagName). It is
 // IdentName when populated by the dedup pass and Name otherwise. The
 // resulting string must never be used for wire-side serialization;
-// callers writing URL params, JSON keys, or path substitutions read
-// Name directly.
+// callers writing URL params read paramWireName, while JSON keys, or
+// path substitutions read Name directly.
 func paramIdent(p spec.Param) string {
 	if p.IdentName != "" {
 		return p.IdentName
 	}
 	return p.Name
+}
+
+// paramWireName returns the URL query-key for this param. URLName overrides
+// when set (e.g., "$limit" for Socrata APIs); otherwise Name. Used by
+// generator templates for URL emission only — not for JSON body keys or
+// path substitution.
+func paramWireName(p spec.Param) string {
+	return p.WireName()
 }
 
 func publicFlagName(p spec.Param) string {
--- a/internal/generator/generator.go
+++ b/internal/generator/generator.go
@@ -248,6 +248,7 @@ func New(s *spec.APISpec, outputDir string) *Generator {
 		"mcpParamDesc":           g.mcpParamDescription,
 		"flagName":               flagName,
 		"paramIdent":             paramIdent,
+		"paramWireName":          paramWireName,
 		"typeFieldIdent":         typeFieldIdent,
 		"safeTypeName":           safeTypeName,
 		"hasNonScalarType": func(types map[string]spec.TypeDef) bool {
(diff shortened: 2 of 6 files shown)


tmchow:
@youdidwhat-towho resolve conflicts

PR #1226, opened 2026-05-12
fix(ci): fold skill docs into lint gate

tmchow:
## Summary
Mergify no longer depends on a separate `skill-docs` check that can be cancelled before it starts. Skill-doc validation still blocks merges through the existing `go-lint` gate when skill docs or the validator change, while ordinary PRs have one fewer required runner job.

## Verification
- `ruby -e 'require "yaml"; ARGV.each { |path| YAML.safe_load_file(path, aliases: false); puts "ok #{path}" }' .github/workflows/lint.yml .mergify.yml`
- `bash .github/scripts/validate-skill-docs.sh`
- `actionlint .github/workflows/lint.yml`

---

GPT-5

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -20,7 +20,7 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Check whether Go lint is needed
+      - name: Check whether lint validations are needed
         id: changes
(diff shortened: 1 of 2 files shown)


tmchow:
@mergify queue

PR #1244, opened 2026-05-12
fix(cli): gate refreshAccessToken emission on Auth.TokenURL non-empty

### LINKED ISSUE #1200: generator: gate refreshAccessToken emission on TokenURL non-empty (bearer-only CLIs ship dead OAuth refresh plumbing)
**Priority:** P2 · **Component:** generator

## Goal

Stop emitting dead `refreshAccessToken()` + the call site that invokes it for bearer-token CLIs that lack a real `Auth.TokenURL`. The function compiles and silently no-ops on every call, mislead­ing users into thinking refresh is wired when it isn't.

## What happened

While generating `google-search-console` (a bearer-token API), the printed CLI shipped with:

```go
// internal/client/client.go (generated)
func (c *Client) refreshAccessToken() error {
    if c.Config == nil {
        return nil
    }
    if c.Config.RefreshToken == "" {
        return nil
    }

    tokenURL := ""   // <-- {{.Auth.TokenURL}} renders empty for bearer-only specs
    if tokenURL == "" {
        return nil
    }
    // ...rest never executes
}
```

`auth status` cheerfully prints `has_refresh_token: ...` and the auth-header preflight calls `c.refreshAccessToken()` on every request whose access token has expired — except the function silently returns `nil` because tokenURL is empty. Users hit a 1-hour OAuth token-refresh wall with no error.

This isn't a `google-search-console` quirk. Survey of 10 published bearer-token CLIs:

| CLI | tokenURL value |
|---|---|
| `marketing/dub` | `""` |
| `marketing/customer-io` | `""` |
| `marketing/producthunt` | `""` |
| `marketing/google-search-console` | `""` (this CLI) |
| `cloud/digitalocean` | `""` |
| `cloud/render` | `""` |
| `developer-tools/firecrawl` | `""` |
| `developer-tools/trigger-dev` | `""` |
| `monitoring/sentry` | `""` |
| `media-and-entertainment/podscan` | `""` |
| `commerce/fedex` | `strings.TrimRight(c.BaseURL, "/") + "/oauth/token"` (only one — uses client_credentials) |

**9 of 10 ship with `tokenURL := ""` baked in.** None of those 9 hand-removed the dead plumbing.

## Root cause

`internal/generator/templates/client.go.tmpl`:

- Line ~1039 — call site, gated on `{{- if .HasAuthCommand}}`:
  ```
  {{- else if .HasAuthCommand}}
      if c.Config.AccessToken != "" && !c.Config.TokenExpiry.IsZero() && time.Now().After(c.Config.TokenExpiry) && c.Config.RefreshToken != "" {
          if err := c.refreshAccessToken(); err != nil {
              return "", err
          }
      }
  {{- end}}
  ```
- Line ~1122 — method emission, gated on `{{- if .HasAuthCommand}}`:
  ```
  {{- if .HasAuthCommand}}
  
  func (c *Client) refreshAccessToken() error {
      // ...
      tokenURL := "{{.Auth.TokenURL}}"
      if tokenURL == "" {
          return nil
      }
      // ...
  }
  ```

`.HasAuthCommand` is truthy for every auth-bearing CLI, including the simple-template (`auth_simple.go.tmpl`) catch-all. The simple template emits `status` / `set-token` / `logout` only — no `login` command that populates `RefreshToken`. So `refreshAccessToken` is unreachable through the printed CLI's surface, yet still emitted and called.

## Target

`internal/generator/templates/client.go.tmpl` — tighten both gates from `{{- if .HasAuthCommand}}` to `{{- if and .HasAuthCommand (ne .Auth.TokenURL "")}}`.

Suggested shape (method emission):

```
{{- if and .HasAuthCommand (ne .Auth.TokenURL "")}}

func (c *Client) refreshAccessToken() error {
    // ...
}

{{- end}}
```

Same gate on the call site around line 1039.

## Acceptance criteria

**Positive (catch case):** Regenerate `dub` (or any bearer-only spec). Inspect `internal/client/client.go`. It must NOT contain `func (c *Client) refreshAccessToken()`. The auth-header preflight must NOT call `c.refreshAccessToken()`.

**Negative (preserve case):** Regenerate an OAuth2 authorization_code API where the parser populates `auth.TokenURL`. `refreshAccessToken()` is still emitted, `tokenURL := "<real URL>"` is non-empty, the call site is still wired up.

**Regression test:** Add a sibling test next to [`internal/generator/client_invalidate_cache_test.go`](../../internal/generator/client_invalidate_cache_test.go) — that file is the model for template-emission gate tests. Two prongs:

1. Bearer-only fixture (e.g., `testdata/<bearer-only>.yaml` with no AuthorizationURL/TokenURL): `refreshAccessToken` not present, call site absent.
2. OAuth2 authorization_code fixture: `refreshAccessToken` present, call site present, `tokenURL := "https://..."` populated.

## Scope boundary

- Does **not** remove the `RefreshToken` field from `config.go.tmpl`. The field is harmless when empty; removing it changes the on-disk TOML shape and could surprise users with custom integrations that set it via external tooling. Pure subtraction here.
- Does **not** add OAuth-source detection (`bearerFormat: "OAuth 2.0 access token"`, description-mentions-OAuth-Playground heuristics, `x-google-*` extensions) that would route OAuth-bearer APIs to the full `auth.go.tmpl` login template. That's the natural follow-up once F1 ships, but a separate work unit.

## Frequency / catalog evidence

Every bearer-token API where `auth.AuthorizationURL` is empty (which is every bearer-token API in the catalog that doesn't ALSO declare OAuth2 flows). Today: 9 of the 10 surveyed published bearer CLIs. Will recur on every new bearer-token CLI generated until the template is gated.

## Complexity

small — two-line template gate, plus one new test in the existing template-test style.

## Related issues

- #1059 (open) — `[P3] WU-2: OAuth2 auth login template missing IsVerifyEnv() short-circuit`. **Related area.** Same `auth.go.tmpl` family but a different defect (verify-env short-circuit in the *full* OAuth2 login path, not dead `refreshAccessToken` emission in the *simple* path). Whoever picks up this WU should sanity-check the gate logic against the same template family.
- #1007 (open) — `[P2] WU-3: Phase 1.9 reachability gate probes OAuth grant flow for oauth2 APIs`. **Related area.** Different layer — pre-generation OAuth probe — but adjacent in the OAuth-emission territory.
- #1157 (open) — `[P2] Slug-derived auth env var rarely matches the canonical name users have set`. **Related area** in the auth-emission territory.

## Anti-pattern

Half-emitted feature scaffolding. The template emits a function body, type field, and helpers, but the rest of the generator chain leaves a critical constant empty. The code compiles and silent-no-ops. The user has no way to know the feature is dead without reading the source. Tighten template gates so a template emits either a working version or nothing at all.

## Evidence + audit trail

Survey of 10 published bearer-token CLIs in the public-library worktree (9 with `tokenURL := ""`, 1 with real URL). Per-CLI patches.json catalog at `~/printing-press/library/google-search-console/.printing-press-patches.json` records the oauth-login patch that hand-wired Google's tokenURL on top of the bearer-only template. Full retro at `manuscripts/google-search-console/20260509-053752/proofs/20260512-061631-retro-google-search-console-pp-cli.md` in the local manuscripts dir.

tmchow:
## Intent

Bearer-only printed CLIs (api_key, bearer_token, OAuth2 specs lacking a TokenURL) were shipping a `refreshAccessToken()` Go method that silently returned `nil` on every call because the templated runtime `tokenURL` was empty. `authHeader()` invoked it whenever an access token expired, hiding the real failure: users hit the OAuth refresh wall with no error and `auth status` reported `has_refresh_token: ...` as if everything was wired up. The issue counted nine of ten surveyed published bearer CLIs with `tokenURL := ""` baked in.

Issue: Closes #1200

## Approach

Tighten both `{{- if .HasAuthCommand}}` gates in `internal/generator/templates/client.go.tmpl` (the method definition and its call site inside `authHeader()`) to `{{- if and .HasAuthCommand (ne .Auth.TokenURL "")}}`. With the outer gate now guaranteeing `.Auth.TokenURL` is non-empty, the redundant inner `{{- if .Auth.TokenURL}}` block in the function body is dropped, and a now-unreachable runtime `if tokenURL == "" { return nil }` guard inside the function body goes with it. Pure subtraction.

The fix follows the co-gating pattern established by PR #704 / [docs/solutions/logic-errors/auth-none-dead-code-in-config-client-2026-05-08.md](docs/solutions/logic-errors/auth-none-dead-code-in-config-client-2026-05-08.md) — extending it from "any auth emits scaffolding" to "auth with a real TokenURL emits refresh plumbing."

## Scope

Primary area: `cli` (generator template + test).

Why this belongs in this repo: This is a machine change per AGENTS.md. Every bearer-token CLI generated today reproduces the bug; fixing it in one printed CLI's source would not generalize and would re-break on the next regen.

## Risk

Low, by design — pure subtraction of dead plumbing for CLIs that lack a TokenURL.

- **Bearer-only specs** (no TokenURL): the method and call site stop emitting. Their `refreshAccessToken()` already silently returned `nil`, so the runtime behavior was already "no-op" — only the dead symbols are removed.
- **OAuth2 authorization_code specs with TokenURL**: emission is unchanged. Pinned by the new `oauth2_authorization_code_emits_refresh` subtest.
- **OAuth2 client_credentials specs**: emission of `refreshAccessToken()` is unchanged (the function is still defined, although the CC flow uses its own `mintClientCredentials()` path inside `authHeader()` rather than the refresh call site). Pre-existing orphan-emission residual flagged by review — out of scope for this PR; worth a separate issue.
- **`auth.type: none` specs**: unchanged — `shouldEmitAuth()` returns false, no emission either way.

## Output Contract

Three golden client.go fixtures updated:

- `testdata/golden/expected/generate-golden-api/.../client.go`: removes the dead `refreshAccessToken()` definition and its call site inside `authHeader()`.
- `testdata/golden/expected/generate-tier-routing-api/.../client.go`: same.
- `testdata/golden/expected/generate-golden-api-oauth2-cc/.../client.go`: keeps `refreshAccessToken()` (the CC spec has a TokenURL), but removes the now-unreachable second `if tokenURL == "" { return nil }` guard inside the function body.

No other generated artifacts change.

## Verification

Commands actually run, on macOS / Go 1.24:

- `go fmt ./...` — clean
- `go vet ./...` — clean
- `go test ./... -short` — all packages green, including new `TestRefreshAccessToken_GatedByTokenURL` (4 subtests) and updated `TestTokenScaffoldingFollowsAuthSurface`
- `scripts/golden.sh verify` — 17/17 cases pass
- `golangci-lint run ./...` — 0 issues

## AI / Automation Disclosure

### DIFF
--- a/internal/generator/auth_none_dead_code_test.go
+++ b/internal/generator/auth_none_dead_code_test.go
@@ -10,36 +10,32 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-// Symbols that exist solely to support an `auth` subcommand. They must
-// disappear from generated config.go / client.go for `auth.type: "none"`
+// Config-side symbols that exist solely to support an `auth` subcommand.
+// They must disappear from generated config.go for `auth.type: "none"`
 // CLIs (no caller can populate them) and stay for any non-none auth flow.
-var (
-	configTokenScaffolding = []string{
-		"AccessToken",
-		"RefreshToken",
-		"TokenExpiry",
-		"ClientID",
-		"ClientSecret",
-		"SaveTokens",
-		"ClearTokens",
-	}
-	clientTokenScaffolding = []string{
-		"refreshAccessToken",
-		"c.Config.AccessToken",
-		"c.Config.RefreshToken",
-		"c.Config.TokenExpiry",
-	}
-)
+//
+// Client-side refresh plumbing (refreshAccessToken et al.) follows a
+// tighter gate (shouldEmitAuth + Auth.TokenURL != "") and is covered by
+// TestRefreshAccessToken_GatedByTokenURL in client_refresh_token_gate_test.go.
+var configTokenScaffolding = []string{
+	"AccessToken",
+	"RefreshToken",
+	"TokenExpiry",
+	"ClientID",
+	"ClientSecret",
+	"SaveTokens",
+	"ClearTokens",
+}
 
 // TestTokenScaffoldingFollowsAuthSurface pins all three branches of
 // shouldEmitAuth(): Auth.Type != "none", Auth.AuthorizationURL != "", and a
 // graphql_persisted_query traffic-analysis hint each independently keep the
-// OAuth-shape token scaffolding emitting; only when all three are absent do
-// the symbols disappear (otherwise they would be dead code with no caller on
-// the CLI surface). The cases also verify that the generated CLI still
-// compiles with the gating in place — gated imports falling out of sync with
-// gated function bodies would otherwise pass the symbol-absence assertions
-// but ship a non-buildable CLI.
+// OAuth-shape token scaffolding emitting in config.go; only when all three
+// are absent do the symbols disappear (otherwise they would be dead code
+// with no caller on the CLI surface). The cases also verify that the
+// generated CLI still compiles with the gating in place — gated imports
+// falling out of sync with gated function bodies would otherwise pass the
+// symbol-absence assertions but ship a non-buildable CLI.
 func TestTokenScaffoldingFollowsAuthSurface(t *testing.T) {
 	t.Parallel()
 
@@ -97,11 +93,6 @@ func TestTokenScaffoldingFollowsAuthSurface(t *testing.T) {
 			configSrc := readGeneratedFile(t, outputDir, "internal", "config", "config.go")
 			for _, sym := range configTokenScaffolding {
 				tt.expect(t, configSrc, sym)
-			}
-
-			clientSrc := readGeneratedFile(t, outputDir, "internal", "client", "client.go")
-			for _, sym := range clientTokenScaffolding {
-				tt.expect(t, clientSrc, sym)
 			}
 
 			// Catch import-gating mistakes: an orphan reference to a gated
--- a/internal/generator/client_refresh_token_gate_test.go
+++ b/internal/generator/client_refresh_token_gate_test.go
@@ -1,0 +1,106 @@
+package generator
+
+import (
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/mvanhorn/cli-printing-press/v4/internal/spec"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// Guards #1200: refreshAccessToken() and its call site must only be emitted
+// when the spec actually carries an OAuth token-refresh endpoint
+// (Auth.TokenURL != ""). For bearer-only CLIs (api_key, bearer_token with
+// no TokenURL) the method becomes a guaranteed no-op that silently swallows
+// every refresh call, leaving users to hit a 1-hour token expiry wall with
+// no error and `auth status` cheerfully reporting `has_refresh_token: ...`.
+//
+// Both prongs must hold: the method definition and the do()-path call site
+// share the same gate, so dropping one without the other leaves either an
+// orphan symbol or an undefined-symbol compile error.
+func TestRefreshAccessToken_GatedByTokenURL(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name string
+		auth spec.AuthConfig
+		want bool
+	}{
+		{
+			name: "bearer_only_omits_refresh",
+			auth: spec.AuthConfig{
+				Type:    "bearer_token",
+				Header:  "Authorization",
+				Format:  "Bearer {token}",
+				EnvVars: []string{"BEARER_ONLY_TOKEN"},
+			},
+			want: false,
+		},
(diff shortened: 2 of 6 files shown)


tmchow:
Accepted-for-later: the P2 on adding a dedicated `client_credentials` subtest is a real coverage gap for the orphan-emission contract (method defined, call site absent) but the contract itself is pre-existing — out of scope for #1200, which targets bearer-only CLIs. The CC golden `generate-golden-api-oauth2-cc` continues to pin the contract, and a follow-up issue is the right place to add the unit test alongside any decision on whether to tighten the method-definition gate to also exclude CC.

Issue #1249, opened 2026-05-12
Templates carry SQL injection + redundant-call patterns flagged by Greptile on PR mvanhorn/printing-press-library#416


ahujasachin92:
## Context

Greptile's review of [printing-press-library PR #416 (uk-train-goat)](https://github.com/mvanhorn/printing-press-library/pull/416) flagged four template findings, plus the same PR surfaced a fifth issue in the publish-side CI workflow. All five originate in **machine artifacts** (templates or the verify-library-conventions.yml workflow), not in uk-train-goat-specific code. The printed CLI was patched in [1c4d54f](https://github.com/mvanhorn/printing-press-library/commit/1c4d54f5923b1bdd735e448722be583b6fe6b769) to unblock the PR, but every CLI in the public library still ships the unpatched template-derived code, and the next regen of uk-train-goat overwrites the in-CLI fix.

Per [`AGENTS.md`](../blob/main/AGENTS.md#machine-vs-printed-cli) ("Default to machine changes. If a problem appears in a printed CLI, ask first whether the generator should have gotten it right"), this is a machine fix.

### 1. P1 — `store.ListIDs` SQL injection (template: `internal/store/store.go.tmpl`)

`resourceType` is interpolated verbatim into `SELECT id FROM %s` via `fmt.Sprintf`. Internal callers pass constants today, but the signature is `func (s *Store) ListIDs(resourceType string)` with no input gate.

```go
// in generator template
query := fmt.Sprintf("SELECT id FROM %s", resourceType)
rows, err := s.db.Query(query)
```

**Patch shape that worked in uk-train-goat (1c4d54f):**

```go
var safeIdentPattern = regexp.MustCompile(`^[a-zA-Z_][a-zA-Z0-9_]*$`)

func (s *Store) ListIDs(resourceType string) ([]string, error) {
    var rows *sql.Rows
    var err error
    if safeIdentPattern.MatchString(resourceType) {
        query := fmt.Sprintf("SELECT id FROM %s", resourceType)
        rows, err = s.db.Query(query)
    } else {
        err = fmt.Errorf("unsafe resource type")
    }
    if err != nil {
        rows, err = s.db.Query("SELECT id FROM resources WHERE resource_type = ?", resourceType)
        if err != nil { return nil, err }
    }
    defer rows.Close()
    // ...
}
```

Keeps the domain-table fast path for safe callers; unsafe inputs fall through to the parameterised `resources` query.

[...]

PR #1278, opened 2026-05-13
fix(cli): close ResolveByName json_extract injection + dedupe cacheKey AuthHeader call

tmchow:
## Intent

Close two remaining template-level findings from [Greptile's review of printing-press-library#416](https://github.com/mvanhorn/printing-press-library/pull/416). The first finding from that review (`ListIDs` SQL injection) was already closed by #1000; this PR finishes the other two so the next regen of any printed CLI doesn't overwrite the in-CLI patch.

Issue: Refs https://github.com/mvanhorn/cli-printing-press/issues/1249

Finding #4 from the issue thread (eval `--dataset` default path) is out of scope — there's no eval template in the generator (`internal/generator/templates/internal/cli/eval.go.tmpl` does not exist); eval in uk-train-goat is a hand-written novel command, not a generator-emitted one.

## Approach

**`store.go.tmpl` ResolveByName.** `json_extract` path components cannot be bound as SQL parameters, so the existing `fmt.Sprintf("$.%s", field)` splice is structurally required. Validate each `field` against `validIdentifierRE` (the same regex `ListField` already uses ~110 lines above) and `continue` past invalid entries. Same defense pattern, no new regex, no new helper.

**`client.go.tmpl` cacheKey.** The block already captured `authHeader := c.Config.AuthHeader()` for the empty check, then called `AuthHeader()` a second time to feed `sha256.Sum256`. Reuse the captured variable. Pure dedupe; no behavior change.

## Scope

Primary area: `internal/generator/templates/store.go.tmpl`, `internal/generator/templates/client.go.tmpl`.

Why this belongs in this repo: per AGENTS.md "Default to machine changes." Both flaws live in templates that ship with every printed CLI. The in-CLI patch in [printing-press-library@1c4d54f](https://github.com/mvanhorn/printing-press-library/commit/1c4d54f) closed it for uk-train-goat only and would be overwritten on next regen.

## Risk

**Behavior preserved.**
- `ResolveByName`: existing callers pass schema-controlled field names like `"name"`, `"key"`, `"email"` — all match `validIdentifierRE`. A current caller passing an unsafe field hits the `continue` (same as if no rows matched), so the function still returns the existing "not found" error rather than panicking.
- `cacheKey`: `authHeader` is the immediate return of `AuthHeader()`; no rotation can happen between the two calls in single-threaded request setup. The dedupe is a pure no-op behaviorally.

Backfill against the public library: every CLI in the public library still ships the old templates until natural regen cadence picks them up.

## Output Contract

Four golden fixtures pick up the template changes:
- 3 client.go fixtures gain the `authHeader` reuse.
- 1 store.go fixture (`sync-walker-golden`) gains the `validIdentifierRE` guard and its 4-line doc comment.

Diffs are exactly the patch shape and nothing else — confirmed via `scripts/golden.sh verify` before update.

## Verification

- `go build -o ./printing-press ./cmd/printing-press` — pass
- `go vet ./...` — pass
- `go test ./...` — 3905 pass, 0 fail
- `golangci-lint run ./...` — pass
- `scripts/golden.sh verify` — 17/17 pass
- `git diff --check` — clean

Regression coverage:
- Updated `TestClientCacheKeyScopesByBaseURLAndAuthIdentity` to assert the captured-reuse pattern and forbid the double-call pattern.
- Added `TestStoreResolveByNameValidatesField` to pin the `validIdentifierRE` guard on the json_extract splice.
- Both tests fail on the pre-fix templates.

## AI / Automation Disclosure

### DIFF
--- a/internal/generator/client_cache_scope_test.go
+++ b/internal/generator/client_cache_scope_test.go
@@ -30,7 +30,9 @@ func TestClientCacheKeyScopesByBaseURLAndAuthIdentity(t *testing.T) {
 
 	require.Contains(t, body, `"|base_url=" + c.BaseURL`, "cache keys must isolate staging/prod or per-tenant base URLs")
 	require.Contains(t, body, `"|auth_source=" + c.Config.AuthSource`, "cache keys should distinguish env/config/profile auth sources")
-	require.Contains(t, body, `sha256.Sum256([]byte(c.Config.AuthHeader()))`, "cache keys should include an auth fingerprint without storing the raw token")
+	require.Contains(t, body, `authHeader := c.Config.AuthHeader()`, "cache keys should capture AuthHeader() once")
+	require.Contains(t, body, `sha256.Sum256([]byte(authHeader))`, "cache keys should include an auth fingerprint without storing the raw token")
+	require.NotContains(t, body, `sha256.Sum256([]byte(c.Config.AuthHeader()))`, "cache keys should reuse the captured authHeader, not call AuthHeader() twice")
 	require.Contains(t, body, `sort.Strings(paramKeys)`, "cache keys should be deterministic for map params")
 }
 
--- a/internal/generator/store_resolve_by_name_guard_test.go
+++ b/internal/generator/store_resolve_by_name_guard_test.go
@@ -9,11 +9,6 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-// TestStoreResolveByNameValidatesField pins the SQL-injection guard on
-// ResolveByName's caller-supplied field names. json_extract path components
-// can't be bound as SQL parameters, so the splice is structurally required
-// — but every field must be validated against validIdentifierRE before
-// being substituted into the query.
 func TestStoreResolveByNameValidatesField(t *testing.T) {
 	t.Parallel()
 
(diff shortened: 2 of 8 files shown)


tmchow:
Thanks for the review. Addressing the findings:

**Issue 1 (test brittleness):** Fixed in [test(cli): tighten ResolveByName guard assertion](https://github.com/mvanhorn/cli-printing-press/pull/1278/commits/bcc916d44b4eb38a581cb74a12a3d31cc56c04f1). Swapped the standalone `continue` assertion for a regex that anchors the `continue` inside the `validIdentifierRE` guard block, so the assertion only passes when the new security guard is present.

**Issue 2 (silent skip vs ListField's explicit error):** Accepting for now. The two functions have intentionally different semantics — `ListField` takes a single field name and can't fall through, so erroring is the only useful response. `ResolveByName` is variadic and the issue#1249 patch shape explicitly specified `continue` past unsafe entries so partial validity (2 valid fields out of 3) still resolves. Surfacing an "all-invalid" error would be a useful follow-up but expands scope beyond this security fix.

tmchow:
@mergifyio queue

Earlier dequeue was the OSS-plan 'Merge Queue Batch' billing block. That root cause was fixed in #1321 (now on main): `batch_size: 1` + `max_parallel_checks: 1`, Greptile enforcement moved into Mergify. Safe to requeue.

tmchow:
@mergifyio dequeue

tmchow:
@mergifyio queue

In-place mode is now active on main. Re-queueing.

PR #1279, opened 2026-05-13
fix(cli): route in:query params to URL on PUT/DELETE/POST/PATCH handlers

### LINKED ISSUE #1275: feat(generator): route in:query parameters to URL query string on PUT/DELETE/POST handlers
## Summary

The generator's PUT, DELETE, and POST handler templates ignore OpenAPI parameters declared `in: query` and instead route them into the JSON request body. The GET template handles `in: query` correctly. The mismatch causes every printed CLI to send malformed requests when a spec puts query parameters on a non-GET endpoint.

## Reproduction

Spotify's OpenAPI spec declares `PUT /me/library` with `uris` as `in: query`:

```yaml
/me/library:
  put:
    operationId: save-library-items
    parameters:
    - name: uris
      required: true
      in: query
```

The printed handler `me_save-library-items.go` builds an empty body and never appends `uris` to the URL:

```go
path := "/me/library"
var body map[string]any  // ends up as `{}`
data, statusCode, err := c.Put(path, body)
```

Spotify responds with HTTP 400 `"Missing required field: uris"` on every invocation. The exact same pattern repeats for `DELETE /me/library` (`me_remove-library-items.go`).

## Cross-API impact

Any OpenAPI spec where a non-GET endpoint declares query parameters hits this. Examples in the public library that likely repeat the bug today:

- GitHub: `PUT /repos/{owner}/{repo}/topics` with `?names=...`
- Atlassian: `DELETE /rest/api/3/issue/{key}/votes` with `?notifyUsers=...`
- Most "save items to library" / "remove items from library" idioms across vendor APIs

So fixing it raises the floor on every printed CLI that wraps a non-GET-with-query-params endpoint, not just future Spotify-shaped specs.

## Proposed fix

1. In `internal/generator/templates/` (PUT, DELETE, POST handler templates):
   - Walk the endpoint's parameter list.
   - For each parameter with `in: query`, append it to a `params map[string]string` exactly the way the GET template does today.

2. In the generated `internal/client/client.go`:
   - Add `PutWithParams(path string, params map[string]string, body any)` (and `DeleteWithParams`, `PostWithParams`) that funnels params through `req.URL.Query()` instead of forcing the handler to hand-roll `path += "?key=" + url.QueryEscape(val)`.

3. The printed-CLI handler then becomes (PUT example):
   ```go
   path := "/me/library"
   params := map[string]string{"uris": flagUris}
   data, statusCode, err := c.PutWithParams(path, params, body)
   ```

## Test case for the golden harness

Add a fixture covering this spec snippet:

```yaml
paths:
  /widgets/{id}/tags:
    put:
      operationId: tag-widget
      parameters:
        - name: id
          in: path
          required: true
          schema: {type: string}
        - name: names
          in: query
          required: true
          schema: {type: string}
```

Expected generated PUT handler: builds `params := map[string]string{"names": flagNames}` and calls `c.PutWithParams("/widgets/"+id+"/tags", params, nil)`. Current behavior: builds an empty body, drops `names` on the floor.

## Where the bug was first surfaced

- `~/printing-press/library/spotify/internal/cli/me_save-library-items.go` (worked around by hardcoding `?uris=` into path)
- `~/printing-press/library/spotify/internal/cli/me_remove-library-items.go` (same workaround)
- spotify-pp-cli PR: mvanhorn/printing-press-library#525

The printed CLIs were patched to ship, but the patches sit under `// Generated by CLI Printing Press ... DO NOT EDIT.` headers and will regress on next regen. The fix has to land in the templates.

## Severity

High. Hits any spec with `in: query` parameters on non-GET methods. Failure mode is "every printed command of this shape silently sends malformed requests"; not a corner case.

tmchow:
## Intent

The PUT, DELETE, POST, and PATCH handler templates ignored OpenAPI parameters declared `in: query` — only the GET path built a `params` map. Non-GET endpoints with query parameters silently dropped them, so e.g. Spotify's `PUT /me/library?uris=...` returned HTTP 400 "Missing required field: uris" on every invocation. The MCP intent classifier had the matching gap: it routed every spec param into the body location regardless of the underlying `in:` value.

Issue: Closes #1275

## Approach

Three coordinated machine changes:

1. **Generator helper** — new `endpointHasQueryFlags(endpoint)` returns true when an endpoint declares any non-positional, non-path param (i.e., a query flag).
2. **Client template** — adds `*WithParams` and `*WithParamsAndHeaders` variants for each write verb (`Post`/`Put`/`Patch`) plus `DeleteWithParams` / `DeleteWithParamsAndHeaders`. The underlying `do(method, path, params, body, headers)` already plumbed params through to `req.URL.Query()`, so the new methods wire through what was already supported instead of introducing a new code path.
3. **Handler templates** — `command_endpoint.go.tmpl` and `command_promoted.go.tmpl` gate on the new helper: when `endpointHasQueryFlags` is true, emit the same `params := map[string]string{}` block the GET branch already uses and route through the `*WithParams` variant. Otherwise emit the existing call shape, so endpoints without query flags keep byte-identical generated output.

The change is additive at both the client and handler layers; published CLIs without `in:query` params on non-GET methods see zero generated-code drift.

## Scope

Primary area: `cli` (generator + templates).

Why this belongs in this repo: This is a machine change per AGENTS.md "Machine vs Printed CLI". Every spec with `in:query` params on non-GET methods reproduces the bug today (GitHub topics PUT, Atlassian votes DELETE, Spotify library PUT/DELETE). Fixing one printed CLI's handler would not generalize and would re-break on the next regen.

## Risk

Low, by design — gated emission and additive client surface.

- **Endpoints without `in:query` params on non-GET methods**: unchanged. `endpointHasQueryFlags` returns false, the existing call shape is emitted. No drift on regen.
- **Endpoints with `in:query` params on non-GET methods**: new params map is built and routed through `*WithParams`. The underlying `do()` already handled the params arg correctly; this just stops passing `nil`.
- **MCP surface**: param location classification is unchanged in the generator (already reads `in:` from the spec). The golden tools.go diff shows the location now reads `"query"` instead of the previous incorrect `"body"` for the new fixture endpoint, but this is the same classifier behavior — only the spec input changed.
- **Existing `*WithHeaders` methods**: untouched. The selection matrix (`bare` / `WithHeaders` / `WithParams` / `WithParamsAndHeaders`) routes through the variant matching the combination of query flags and per-endpoint header overrides.

## Output Contract

Golden fixture changes:

- `testdata/golden/fixtures/golden-api.yaml`: added an `in:query` `notify` flag to the `PATCH /projects/{projectId}/tasks/{taskId}` endpoint to cover the new branch.
- Generated client.go in `generate-golden-api`, `generate-golden-api-oauth2-cc`, and `generate-tier-routing-api` gains the 8 new `*WithParams`/`*WithParamsAndHeaders` methods.
- Generated `projects_tasks_update-project.go` builds a `params` map and calls `c.PatchWithParams`.
- Generated `tools.go` for the same endpoint classifies `notify` as `Location: "query"`.

No other generated artifacts change.

## Verification

- `go build ./...` ✓
- `go test ./...` — 3885 tests passed
- `go fmt ./...` clean
- `go vet ./...` clean
- `golangci-lint run ./...` clean
- `scripts/golden.sh verify` — 17/17 PASS
- `rtk git diff --check` clean


tmchow:
Addressing Greptile findings:

- **`golden-api.yaml` ticket-number comment**: fixed in the follow-up commit — the regression-guard prose stays, only the `#1275` reference goes.
- **Multipart/form non-GET endpoints with `in:query` params still drop query params**: accepted-for-later. Greptile correctly noted this is a pre-existing gap, not a regression introduced here. The original issue body specifically scopes the fix to JSON-body handlers; extending the same shape to multipart/form is a parallel, additive change (new client variants, separate template branches, golden fixture). Filed as #1282 to keep this PR focused on the JSON-body fix the issue calls out.

tmchow:
Update: pulled the multipart/form gap into this PR instead of a follow-up. Added 12 new client variants (`*MultipartWithParams`, `*MultipartWithParamsAndHeaders`, `*FormWithParams`, `*FormWithParamsAndHeaders` for POST/PUT/PATCH), gated the multipart and form template branches on `endpointHasQueryFlags`, and added a `PUT /projects/{projectId}/avatar` multipart fixture with an `in:query overwrite` flag to cover the new branch end-to-end. Closed #1282 since this PR now handles both shapes.

PR #1301, opened 2026-05-13
feat(cli): emit fetchFull<X> companion for GET-embedded paged sub-resources

### LINKED ISSUE #1277: feat(generator): emit fetchFull<X> companion when GET embeds a paginated sub-resource
## Summary

When an OpenAPI spec returns a paginated sub-resource embedded in a parent `GET /resource/{id}` response (e.g. Spotify's `GET /playlists/{id}` embeds `tracks` as a `PagingObject`), the generated GET handler returns whatever the API embedded in the first page — and the embed is hard-capped at 50–100 items by most vendors. Any consumer that treats the embedded array as "the full list" silently truncates.

The truncation is detectable from the spec: the embedded property's schema references the same paging type used for top-level paginated list endpoints (`PagingObject`, `Connection`, `PageResult`, etc.), and the parent response shape has both `items` and `next` nested under it.

## Reproduction

Spotify's `GET /playlists/{id}` returns:

```json
{
  "id": "...",
  "name": "...",
  "snapshot_id": "...",
  "tracks": {
    "items": [ ... up to 100 entries ... ],
    "next": "https://api.spotify.com/v1/playlists/{id}/tracks?offset=100&limit=100",
    "total": 247
  }
}
```

The printed `playlists get` handler (and any hand-built command that calls `c.Get("/playlists/"+id, nil)`) only sees the first 100 tracks. The `next` URL is right there in the response but neither the generator nor any helper consumes it.

In spotify-pp-cli this surfaced as three correctness bugs:

- **T1 `playlists diff`** stored an incomplete baseline snapshot, so any track at position 100+ was permanently reported as "removed" on the next diff.
- **T2 `playlists dedupe`** missed duplicates at positions 100+, since they weren't in the in-memory groups map.
- **T3 `playlists merge`** dropped the tail of any source playlist over 100 tracks before posting to the destination.

All three were patched in mvanhorn/printing-press-library#525 (commit `c221cce8`) by routing through a new `fetchFullPlaylist(c, id)` helper that calls `GET /playlists/{id}?fields=id,name,snapshot_id` for metadata and `fetchAllPaged(c, "/playlists/{id}/tracks", ...)` for tracks.

## Cross-API impact

Any OpenAPI spec that embeds a paginated sub-resource has the same hidden ceiling. Examples in the public library that probably hit this today:

- GitHub: `GET /repos/{owner}/{repo}/issues/{number}` embeds `comments_url` separately, but `GET /repos/{owner}/{repo}/pulls/{n}` embeds `commits`/`review_comments` paged objects.
- Slack: `GET /conversations.info` embeds `members` as a paged sub-resource.
- Notion: `GET /pages/{id}` embeds `properties` and `children` (the latter is hard-capped at 100).
- Atlassian Jira: `GET /issue/{key}` embeds `comments` as a paged sub-resource (capped at 50).

So a generator fix raises the floor across many CLIs.

## Proposed fix

Two flavors, in order of how much the generator does:

**Flavor A (minimum): emit a paginated companion handler.**

When the generator processes `GET /resource/{id}`, walk the response schema for properties whose type matches the spec's paging shape (`PagingObject`, `Page<T>`, `Connection<T>`). For each such property `P`, emit a sibling handler `fetchFull<Resource><P>(c, id)` alongside the standard handler. The standard handler stays as-is for cases where the caller only wants metadata; the companion paginates `/resource/{id}/<P>` (the URL pattern Spotify, Slack, Atlassian, Notion all follow) until exhausted.

**Flavor B (better): also rewrite the embedded shape.**

In the standard handler, replace the truncated `items` array in the unmarshalled response with `nil` and emit a warning that callers must use the companion helper. This forces every call site that depends on the full list to either acknowledge the truncation or opt into the paginated path. Otherwise the standard handler keeps quietly returning partial data.

## Detection heuristic for the generator

A property is a paginated sub-resource if BOTH:

1. Its schema is a struct/object with an `items` array (or vendor equivalent: `data`, `results`).
2. The same object has a `next` URL / `next_cursor` / `has_more` field.

This is the same heuristic `extractPaginatedItems` in `internal/cli/helpers.go` already uses for the top-level path. The generator can reuse it during spec walk.

## Severity

High. Failure mode is silent data corruption — the CLI returns a value that LOOKS complete (no error, valid JSON) but is wrong. Three separate transcendence features in spotify-pp-cli hit it in one session.

tmchow:
## Intent

Fix the silent-corruption class of bug described in [#1277](https://github.com/mvanhorn/cli-printing-press/issues/1277): when an OpenAPI GET response embeds a paged sub-resource (e.g. Spotify `GET /playlists/{id}` embeds `tracks` as a PagingObject capped at 100 entries), any caller that treats the embedded array as the complete list silently truncates. Three transcendence features in spotify-pp-cli hit it in one session — and the same shape exists across Slack, Atlassian, Notion, GitHub, and Stripe specs.

Issue: Closes #1277

## Approach

Implements Flavor A from the issue's proposal — detection plus emission of a companion helper, without changing existing GET-handler behavior:

1. **Parser (`internal/openapi/parser.go`).** For every GET endpoint with a path placeholder, walk the success response schema's top-level properties. A property qualifies when its nested schema has both an items-style array field (`items`/`data`/`results`/`messages`/`members`/`values`) and a next-page signal (`next`/`next_cursor`/`next_page_token`/`cursor` string or `has_more` bool). Property matching is case-insensitive so PascalCase schemas (`Items`, `NextPageToken`) are detected. List endpoints (no path placeholder) are skipped — they ARE the paged endpoint.

2. **Spec (`internal/spec/spec.go`).** New `EmbeddedPagedSubresource` type and `Endpoint.EmbeddedPagedSubresources []EmbeddedPagedSubresource` field. Additive (omitempty); existing internal YAML specs unaffected.

3. **Runtime helper (`helpers.go.tmpl`).** Single shared `fetchEmbeddedPagedSubresource(c, childPath, nextField, nextIsBoolean)` paginates the dedicated child endpoint. Page cap (500) bounds the loop against pathological servers. Absolute next URLs are relativized so the client's BaseURL+path concat doesn't double-prefix. Envelope lookups are case-insensitive at runtime to compensate for schema/wire casing drift. `has_more=true` emits a stderr `truncated` event (mirroring `emitTruncationWarning`) rather than masquerading partial results as complete. Emitted only when `HelperFlags.HasEmbeddedPaged` is set.

4. **Per-endpoint thin wrapper (`command_endpoint.go.tmpl`, `command_promoted.go.tmpl`).** Each detected sub-resource yields a `fetchFull<Resource><Endpoint><Property>(c, pathParams)` function that substitutes the path placeholders and delegates to the shared runtime helper. Available for hand-written novel commands; the standard GET handler is unchanged.

## Scope

Primary area: generator (parser + templates) plus a new shared runtime helper.

Why this belongs in this repo: a machine-side fix per AGENTS.md "Machine vs Printed CLI" — every printed CLI generated from a spec with embedded paged sub-resources benefits without per-CLI patching. Affects current and future printed CLIs across Spotify, Slack, Notion, Atlassian Jira, GitHub PRs, Stripe, and any vendor that follows the pattern.

## Risk

- **Generated output.** Additive: only CLIs whose specs trigger detection see new code. Existing 17/17 golden fixtures pass unchanged (none of them embed paged sub-resources).
- **MCP surface.** Unchanged — helpers are package-private internal helpers, not Cobra commands. MCP tool catalogue is identical.
- **Auth/catalog/publish/verifier/scorer.** No changes.
- **Existing CLIs in the public library.** No retroactive change — the new helper appears only when CLIs are regenerated from a spec that triggers detection.

## Verification

- `go build ./...` — clean
- `go test ./...` — 3915/3915 pass
- `scripts/golden.sh verify` — 17/17 PASS
- End-to-end generation against a minimal Spotify-shape spec produces a CLI that compiles and exposes `fetchFullPlaylistsGetTracks` with the expected delegation shape

## Tests added

- `internal/openapi/embedded_paged_test.go` — parser detection across Spotify shape, Stripe has_more shape, no-envelope negative, list-endpoint-skipped, multi-property GitHub shape, PascalCase casing, and a cross-package consistency test guarding drift between `itemsFieldNames` and runtime `extractPaginatedItems`.
- `internal/generator/embedded_paged_test.go` — generator emission for non-promoted, promoted, has_more, and empty (gate-off) endpoints, plus the shared runtime helper presence/absence assertions.

## Pre-merge review

Ran `/ce-simplify-code` and `/ce-code-review mode:headless` over the branch. Acted on findings:
- Extracted shared runtime helper to eliminate template duplication and reduce generated-output bloat.
- Fixed absolute-URL handling (Spotify next URLs would otherwise double-concat against BaseURL).
- Added case-insensitive runtime lookup for envelope properties.
- Wrapped page errors with index and unmarshal failures so partial results don't return with `nil` error.
- Emit truncation event when `has_more=true`.
- Removed vendor names from generator code/templates per AGENTS.md.
- Removed ticket-reference phrasing per AGENTS.md.
- Raised page cap to 500 (covers 50k-item collections at 100/page).


tmchow:
@mergifyio queue

Earlier dequeue was the OSS-plan 'Merge Queue Batch' billing block. That root cause was fixed in #1321 (now on main): `batch_size: 1` + `max_parallel_checks: 1`, Greptile enforcement moved into Mergify. Safe to requeue.

tmchow:
@mergifyio dequeue

tmchow:
@mergifyio queue

Sorry — I dequeued this to force re-evaluation under the new in-place-mode config but never re-queued it. Doing so now.

tmchow:
@mergifyio queue

Rebased onto current main + refreshed three golden fixtures (cosmetic update from #1280's wantsHumanTable guard). Verify passes locally.

PR #1341, opened 2026-05-13
fix(cli): sync pagination, auth aliases, narrative timing, and codex prompt

salmonumbrella:
## Summary

Six independent generator-level findings surfaced during a downstream CLI generation run against an offset-paginated REST API. Each is generic; the discovery context is not in the diff or test data.

### 1. AuthHeader OR-semantics for legacy `x-auth-env-vars`
`internal/spec/spec.go`

`IsAuthEnvVarORCase` only returned true when all `EnvVarSpecs` were non-required `per_call`. Specs using the legacy `x-auth-env-vars` list form (or a populated `EnvVars` slice with default `Required=true`) fell through to the canonical-only path: only the first env-var alias reached `AuthHeader`, the rest were populated by `Load()` but never read.

Updated `IsAuthEnvVarORCase` to also return true when:
- `EnvVarSpecs` has 2+ entries that are all request-credentials, OR
- legacy `EnvVars` list has 2+ entries.

**New pin:** `TestAuthHeader_LegacyEnvVarsList_OrSemantics`.

### 2. `validate-narrative` side-effect classifier
`internal/narrativecheck/narrativecheck.go`

`--full-examples` ran every command whose `--help` advertised `--dry-run`, including `auth set-token YOUR_TOKEN_HERE` and `auth logout`. The first persisted the literal placeholder to the user's `~/.config`; the second wiped saved credentials.

Added `isSideEffectfulNarrativeExample` skip-classifier covering `auth set-token`, `auth logout`, `auth setup`, `auth login`, `--launch`, and bare `--apply`. Skipped commands return `StatusUnsupported` with an explanatory reason rather than executing.

**New pins:** `TestRunFullExample_SkipsAuthSetToken`, `TestRunFullExample_SkipsAuthLogout`.

### 3. Empty-array marshalling guidance for Codex
`skills/printing-press/references/codex-delegation.md`

Codex-generated novel-feature commands declared `var results []T`, which marshals nil to JSON `null`. Empty list outputs broke `jq '.[]'` agent pipelines.

Updated the transcendence-command prompt template to require `results := make([]T, 0)` for all list-shape outputs, with a paragraph at the top of `CONVENTIONS` spelling out the contract.

### 4. Shipcheck leg ordering
`internal/cli/shipcheck.go`

`dogfood` ran first and synthesized `README.md` and `SKILL.md` from `research.json.novel_features_built` — including any planned commands whose command-paths didn't actually resolve against the built binary. `validate-narrative` ran later and caught them but the user-facing files were already wrong.

Reordered: `verify`, `validate-narrative`, `dogfood`, `workflow-verify`, `verify-skill`, `scorecard`. Comment updated to reflect the dependency: `verify` builds the binary; `validate-narrative` checks research.json against the binary BEFORE `dogfood` renders from it.

### 5. Offset+`has_more` pagination loop
`internal/generator/templates/sync.go.tmpl`

The sync loop's break-condition required `nextCursor != ""` alongside `hasMore=true`. APIs using offset+has_more (no cursor token in the response) returned `hasMore=true` with empty `nextCursor` on every page, so the loop broke after page 1 and only the first page worth of records reached the local store.

Fixed at both occurrences (linear and parent-child sync paths): when `cursorParam=="offset"` and the envelope returned no cursor despite `hasMore`, compute `nextOffset` client-side as `currentOffset + pageSize.limit`. Cursor-based APIs that genuinely return no cursor still break (existing behavior; silent infinite-loop guard).

**New pin:** `TestGeneratedSyncAdvancesOffsetWhenHasMoreWithoutCursor`.

### 6. Per-resource pagination param gate
`internal/profiler/profiler.go` + `internal/generator/templates/sync.go.tmpl`

Sync sent `?limit=` and `?offset=` to every list endpoint regardless of whether the spec declared those params for that endpoint. List endpoints that don't paginate (typical for `/me`, `/settings`, small reference resources) rejected the synthetic params with HTTP 400 "Invalid query parameter".

Added `profiler.SupportsPagination` per syncable+dependent resource (detected from endpoint params at generate time). Generator template emits `resourceSupportsPagination()` from the per-resource map; sync gates `limit`/`offset` setting on its return value.

**New pin:** `TestGeneratedSyncGatesPaginationParamsPerResource`.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` (30 packages, ~13 min runtime, 0 failures)


tmchow:
@salmonumbrella fix CI failures

tmchow:
Following up: this PR is still showing merge conflicts against `main` after the last push. Please rebase and resolve.

salmonumbrella:
@tmchow rebased onto current `main` (`875d64e9`). The four commits replay cleanly; the only diff vs. the pre-rebase tip is a comment merge in `internal/spec/spec.go` around `IsAuthEnvVarORCase` (the rebased comment now subsumes both branches).

tmchow:
@salmonumbrella next time do not combine this many independent fixes into a single PR, that's just lazy. your agent can figure out separate PRs.

salmonumbrella:
thanks for feedback @tmchow will do.

Issue #1332, opened 2026-05-13
Tenant-templated paths in sync emit empty path map and have no fill mechanism


piercekiltoff:
## What happened

`printing-press generate` for ServiceTitan Dispatch emitted a `syncResourcePath` function with an empty `paths := map[string]string{}` map. After hand-patching that to populate the eight tenant-scoped paths (`/tenant/{tenant}/teams`, etc.), `sync` still failed with `unfilled_path_key keys=["{tenant}"]` for every resource. The framework's path-templating engine treats `{tenant}` as a parent-context dependency that `--param tenant=<id>` does NOT satisfy — `--param` adds query params, not path params.

Symptom (live testing):

```
$ printing-press sync --param tenant=$ST_TENANT_ID --resources zones --json
{"event":"sync_warning","resource":"zones","reason":"unfilled_path_key","keys":["{tenant}"],"path":"/tenant/{tenant}/zones","message":"path /tenant/{tenant}/zones requires parent context ({tenant}); resource skipped"}
{"event":"sync_summary","total_records":0,"resources":1,"success":0,"warned":1,"errored":0,"duration_ms":0}
Error: 1 resource(s) skipped due to insufficient access
```

## Root cause

Two-part bug:

1. **Generator spec walker** doesn't populate `syncResourcePath`'s map for resources whose path templates contain non-`{id}` parent-context placeholders (`{tenant}`, `{workspace}`, `{org}`).
2. **Framework `internal/cli/sync.go` template** has no hook to inject a CLI-level "global" flag value into `{tenant}` / `{workspace}` / `{org}` slots before path expansion at sync request time.

## Cross-API check

Recurs across every per-tenant or per-workspace API. Direct evidence:
- **ServiceTitan Dispatch** (this run, `tenant-dispatch-v2.json`) — `/tenant/{tenant}/...` for all 24 paths
- **ServiceTitan JPM** (`~/printing-press/library/servicetitan-jpm/`, printed 2026-05-12) — same `/tenant/{tenant}/` shape; sync may be silently broken there too (user uses positional `<tenant>` args on typed commands, never tried `sync`)
- **ServiceTitan CRM** (`~/printing-press/library/servicetitan-crm/`) — same shape

Plus 22 more ServiceTitan module specs in the user's vault that all share `/tenant/{tenant}/` paths.

[...]

PR #1368, opened 2026-05-14
fix(cli): consume x-tenant-env-var so per-tenant sync paths ship populated

### LINKED ISSUE #1305: generator: x-tenant-env-var spec extension not consumed; defaultSyncResources + syncResourcePath ship empty
## Summary

When a spec declares a per-tenant path-positional parameter (e.g., `/tenant/{tenant}/<resource>`) and documents the env-var source via `info.x-tenant-env-var: ST_TENANT_ID`, the generator parses the extension but doesn't consume it: `defaultSyncResources()` returns `[]string{}`, and `syncResourcePath()`'s `paths` map is empty. Sync silently no-ops, and every offline transcendence command (which depends on synced data) ships broken.

This is a narrower follow-up to the JPM-retro's broader "tenant/account-wide path param not parameterized at config level" finding. JPM retro raised the same template gap at the broader level; CRM run hits the same gap and the durable fix is to consume the spec extension explicitly.

## What happened (CRM run)

Pre-generation enrichment added `info.x-tenant-env-var: ST_TENANT_ID` at the spec root as a documented hint. The OpenAPI parser parsed it (no error). After `printing-press generate`:

```go
// internal/cli/sync.go
func defaultSyncResources() []string {
    return []string{}    // <-- empty stub
}

func syncResourcePath(resource string) (string, error) {
    paths := map[string]string{}    // <-- empty stub
    if p, ok := paths[resource]; ok {
        return p, nil
    }
    return "", fmt.Errorf("unknown sync resource %q", resource)
}
```

Required hand-populating both stubs with all 6 top-level CRM resources + `/tenant/<ST_TENANT_ID>/<resource>` templates + `os.Getenv("ST_TENANT_ID")` lookup. After the patch, `sync --resources customers,leads --json` pulled 100 customers + 100 leads from the JKA tenant in 409ms.

## Cross-API check (3 named with evidence)

- **ServiceTitan-* (25 modules)**: every endpoint is `/tenant/{tenant}/<resource>`; tenant is account-wide.
- **AWS sigv4 services** (e.g., S3 buckets, Lambda functions in account scope): account-id path-positional that's account-wide.
- **Atlassian Cloud (Jira, Confluence)**: workspace ID is path-positional and account-wide for site-scoped tokens.

Frequency: multi-tenant SaaS APIs with per-tenant path prefixes. Common across ST modules + AWS account-scoped + Atlassian site-scoped APIs.

## Durable fix

When the generator sees `info.x-tenant-env-var: <NAME>` in the spec:

1. Add `Config.TenantID` field (named after the env var) with `os.Getenv` + `strings.TrimSpace` loader (same pattern as the apiKey fix in the sibling F1 issue).
2. Populate `defaultSyncResources()` with the snake-kebab-cased names of every resource that has a `getList` endpoint (the generator already enumerates these for absorbed-command emission).
3. Populate `syncResourcePath()` paths map with `/<positional>/<env-substitution>/<resource>` per resource.

Skip emission entirely when the spec lacks the extension (preserves single-tenant behavior).

## Acceptance

- **Positive**: a fresh generate of any ST module produces a working `sync` that pulls all resources without any post-gen patch.
- **Negative**: a generate of a single-tenant API (no `x-tenant-env-var`) doesn't emit a spurious env getenv or substitution.

## Related issues

- JPM-retro local doc (extends) — `manuscripts/servicetitan-jpm/20260512-085136/proofs/20260512-121549-retro-servicetitan-jpm-pp-cli.md` finding #4 ("Tenant/account-wide path param not parameterized at config level"). Same template gap; this issue narrows to the spec-extension contract.

## Context

**Source retro:** servicetitan-crm (run 20260512-143214)
**Sibling module:** servicetitan-jpm (printed 2026-05-12; retro filed locally)
**Press version:** 4.5.1 (HEAD c05d3bc + 2 pulled fixes for OAuth scheme selection + refresh-token gate)
**Run scorecard:** 87/100 Grade A, verify 100%, live dogfood 185/185 vs JKA tenant

## Artifacts

Same as F1 (catbox blocked; zips saved locally).

tmchow:
## Summary

Closes #1305. Cross-references #1332 (same code path, broader generic
\`--path-context\` fix is separate work).

ServiceTitan-shape APIs (every path is \`/tenant/{tenant}/<resource>\`) ship
a broken sync surface today: the generator emits empty stubs for
\`defaultSyncResources()\` and \`syncResourcePath()\`, sync silently no-ops
(exit 0, \`resources:0\`), and every offline command that depends on synced
data fails downstream. The cross-CLI tally in #1305 shows 4 confirmed hits
plus ~22 sibling ServiceTitan specs sharing the shape.

## What changed

Rather than introduce a parallel tenant-only codepath, this wires the
\`info.x-tenant-env-var: <ENVNAME>\` OpenAPI extension into the existing
\`EndpointTemplateVars\` + \`buildURL\` substitution mechanism that Shopify
\`{shop}\`/\`{api_version}\` already uses.

- **Parser** (\`internal/openapi/parser.go\`): \`x-tenant-env-var: ST_TENANT_ID\`
  registers \`"tenant"\` as an \`EndpointTemplateVar\` and records the env-var
  name in a new \`APISpec.EndpointTemplateEnvOverrides\` map so downstream
  emitters use the spec-declared name instead of the conventional
  \`<APINAME>_<PLACEHOLDER>\`.
- **Profiler** (\`internal/profiler/profiler.go\`): paths whose only
  \`{placeholder}\`s are \`EndpointTemplateVars\` are now treated as
  standalone-listable, so \`/tenant/{tenant}/customers\` becomes a flat
  \`SyncableResource\` rather than landing in the parameterized bucket. Mixed
  paths like \`/tenant/{tenant}/channels/{channel_id}\` stay parameterized
  (regression test added).
- **Templates** (\`config.go.tmpl\`, \`url.go.tmpl\`, \`readme.md.tmpl\`): route
  env-var name resolution through a new \`endpointTemplateEnvName\` FuncMap
  helper that consults the override map.
- **sync.go.tmpl**: filters \`EndpointTemplateVars\` placeholders out of the
  unresolved-key warning so \`{tenant}\` paths don't get skipped as
  \"requires parent context\". Conditionally emitted — non-template-var
  specs render byte-identical output (golden harness clean).
- **Manifest plumbing** (\`internal/pipeline/climanifest.go\`,
  \`mcpb_manifest.go\`, \`publish.go\`): \`EndpointTemplateEnvOverrides\` flows
  through publish + refresh so MCPB \`user_config\` / env names use the
  override. The previously-local default-name branch in
  \`endpointTemplateEnvVar\` is replaced with
  \`spec.DefaultEndpointTemplateEnvName\` — single source of truth (the
  prior copy silently disagreed with the spec helper on hyphenated API
  names like \`servicetitan-crm\`).
- **Docs**: \`docs/SPEC-EXTENSIONS.md\` documents the new extension under
  Info Extensions, per AGENTS.md \"Pointer-rot rule.\"

## Acceptance (from #1305)

- **Positive**: a fresh generate of any ServiceTitan module produces a
  working \`sync\` that pulls all resources without any post-gen patch.
  Anchored by \`TestGenerateEndpointTemplateEnvOverridesWireThrough\` in
  \`internal/generator/generator_test.go\`, which asserts the override
  flows into \`url.go\`, \`config.go\`, \`sync.go\`, and the README.
- **Negative**: a generate of a single-tenant API (no
  \`x-tenant-env-var\`) doesn't emit a spurious env getenv or substitution.
  Anchored by the existing \`TestGenerateNoEndpointTemplateVarsByteCompat\`
  + golden harness (18 cases, all PASS).

## Scope it leaves on the floor

- A generalized \`x-path-template-env-vars: { tenant: ENV1, workspace:
  ENV2 }\` map-shape (or a runtime \`--path-context\` flag from #1332). The
  current extension hardcodes the placeholder name to \`tenant\` because
  that's the actual ServiceTitan shape; the docs flag the generalization
  path for future work.
- A small behavior change in \`config.go.tmpl\`: \`TemplateVars\` values are
  now trimmed via \`strings.TrimSpace(v)\` on read. Helps users who paste
  tenant IDs with trailing whitespace and doesn't move generator output
  for existing Shopify-shape specs (verified via golden).

## Test plan

- [x] \`go build ./...\`
- [x] \`go test ./...\` (3949 pass)
- [x] \`go vet ./...\`
- [x] \`golangci-lint run ./...\`
- [x] \`scripts/golden.sh verify\` (18 cases pass)
- [x] \`govulncheck ./...\` (clean)

### DIFF
--- a/docs/SPEC-EXTENSIONS.md
+++ b/docs/SPEC-EXTENSIONS.md
@@ -219,6 +219,51 @@ x-mcp:
   endpoint_tools: hidden
 ```
 
+### `x-tenant-env-var`
+
+Declares the env-var name that resolves the implicit `{tenant}` path
+placeholder for multi-tenant SaaS APIs whose every path is
+`/tenant/{tenant}/<resource>`. Without this annotation, the generator
+classifies tenant-templated paths as parent-context-dependent and emits an
+empty `defaultSyncResources` / `syncResourcePath` map; sync silently no-ops
+and every downstream offline command ships broken.
+
+Parsed fields: `APISpec.EndpointTemplateVars` (`tenant` added) and
+`APISpec.EndpointTemplateEnvOverrides["tenant"]` (env-var name).
+
+Rules:
+- Optional. Specs without `x-tenant-env-var` keep single-tenant behavior;
+  no `{tenant}`-aware emission, no spurious env reads.
+- Declared under `info` only (path-positional templates are spec-wide).
+- Value must be a non-empty string after `TrimSpace`. Whitespace-only
+  values are treated as absent.
+- The placeholder name is `tenant`. Specs that use a different
+  placeholder (`{workspace}`, `{org}`) should set
+  `EndpointTemplateVars` + `EndpointTemplateEnvOverrides` directly in
+  internal YAML until this extension generalizes.
+
+Effect on generated output (when set):
+- The profiler treats `/.../{tenant}/...` paths as standalone-listable, so
+  the resource becomes a flat `SyncableResource` rather than a
+  `DependentSyncResource`.
+- The emitted `config.go` reads the override env-var name (e.g.
+  `ST_TENANT_ID`) into `Config.TemplateVars["tenant"]` at `Load()` time.
+- The emitted `url.go` `buildURL` substitutes `{tenant}` from
+  `Config.TemplateVars` at request time and names the override env var in
+  the actionable error when the value is missing.
+- The emitted `sync.go` filters `{tenant}` out of the unresolved-key
+  warning so per-tenant paths don't get skipped as "requires parent
+  context".
+
+Example:
+
+```yaml
+info:
+  title: ServiceTitan CRM
+  version: 1.0.0
+  x-tenant-env-var: ST_TENANT_ID
+```
+
 ## Security Scheme Extensions
 
 Security scheme extensions are read from
--- a/internal/generator/generator.go
+++ b/internal/generator/generator.go
@@ -336,8 +336,17 @@ func New(s *spec.APISpec, outputDir string) *Generator {
 		// `?limit=N` query param without honoring it; truncating client-
 		// side means the user-facing --limit flag works regardless.
 		// Surfaced by hackernews retro #350 finding F6.
-		"endpointNeedsClientLimit":       endpointNeedsClientLimit,
-		"envName":                        naming.EnvPrefix,
+		"endpointNeedsClientLimit": endpointNeedsClientLimit,
+		"envName":                  naming.EnvPrefix,
+		// endpointTemplateEnvName resolves the env-var name for a
+		// {placeholder} in EndpointTemplateVars. Returns the spec-declared
+		// override (e.g. ST_TENANT_ID for {tenant}) when one exists, else
+		// the conventional <APINAME>_<UPPER_PLACEHOLDER>. Bound to the
+		// generator's current spec; callers in templates pass just the
+		// placeholder name.
+		"endpointTemplateEnvName": func(placeholder string) string {
+			return s.EndpointTemplateEnvName(placeholder)
+		},
 		"safeName":                       safeSQLName,
 		"resourceIDFieldOverrideEntries": resourceIDFieldOverrideEntries,
 		"criticalResourceEntries":        criticalResourceEntries,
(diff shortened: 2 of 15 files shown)


tmchow:
Both prior threads addressed and resolved. @greptileai please re-review the new HEAD (0cbd0c9b).

PR #1401, opened 2026-05-14
feat(cli): add x-path-template-env-vars spec extension

brysonerandall-a11y:
## Intent

Adds a generic, map-shaped successor to `x-tenant-env-var` so specs with arbitrary path placeholders can resolve them either at runtime (env-var override, parallel to the tenant flow) or at generation time (literal substitution that bakes the value into operation paths and drops the matching path parameter).

Issue: N/A

## Approach

Introduces `x-path-template-env-vars`, read from `info`, in `internal/openapi/parser.go`. Each entry binds a placeholder name to an object with two optional fields:

- `env` registers the placeholder for runtime BaseURL substitution through the existing `EndpointTemplateVars` / `EndpointTemplateEnvOverrides` bucket, the same path `x-tenant-env-var` already takes.
- `default` bakes a literal into operation paths at generation time and drops the matching path parameter through a post-parse sweep (`applyPathParamDefaults`), so downstream consumers (profiler, generator, manifest emitter) see fully resolved paths and never re-emit a flag for the resolved placeholder.

When both are set on the same entry, `default` wins and `env` is ignored. The legacy `x-tenant-env-var` path is preserved and coexists with the new extension; both feed the same template-vars bucket.

Adds `APISpec.EndpointPathParamDefaults map[string]string` to carry the generation-time substitutions, populated from the extension or set directly in internal YAML.

## Scope

Primary area: machine (generator parser, internal spec shape, docs).

Why this belongs in this repo: this is a generator capability that benefits any future spec with canonical always-valid path params (Gmail's `userId=me`, etc.) or BaseURL templating beyond the implicit `{tenant}` slot, not a printed-CLI fix.

## Risk

- Specs without the extension keep byte-identical output: the new code paths short-circuit when no entries are present, and `applyPathParamDefaults` is a no-op when `EndpointPathParamDefaults` is empty. Confirmed by `scripts/golden.sh verify` (18/18 cases pass).
- `applyPathParamDefaults` is the only mutating side of the change. It walks `Resources` and their `SubResources` recursively, replacing matching placeholders and dropping matching params on each endpoint.
- No changes to auth, MCP surface, catalog, publish flow, verifier, scorer, or release behavior.

## Output Contract

No change to existing generated output. New behavior activates only when a spec declares `x-path-template-env-vars` (or sets the new `EndpointPathParamDefaults` field in internal YAML).

## Verification

- `go fmt ./...` — clean
- `go vet ./...` — clean
- `go build -o ./printing-press ./cmd/printing-press` — clean
- `go test ./internal/openapi/ ./internal/spec/ -count=1` — pass
- `go test ./...` — all packages pass
- `scripts/golden.sh verify` — 18/18 cases pass
- `TestParsePathTemplateEnvVarsExtension` covers env-only, default-only, both-set-default-wins, coexists-with-`x-tenant-env-var`, absent, and whitespace-only entry shapes.

## AI / Automation Disclosure


brysonerandall-a11y:
Addressed all three Greptile findings in 16bfe5a5.

1. `applyPathParamDefaults` now consults `isPathSubstitutionParam` before dropping a param, so a same-named query or body param survives the resolution. New sub-test: `default_drop_preserves_a_same-named_non-path_param_on_the_same_operation`.
2. `parseEndpointTemplateExtensions` now scrubs `EndpointTemplateVars` and `EndpointTemplateEnvOverrides` of any key declared in `defaults`, so a `tenant` placeholder governed by both `x-tenant-env-var` and `x-path-template-env-vars: tenant: {default: ...}` resolves cleanly with no stale runtime config left behind. New sub-test: `default_in_x-path-template-env-vars_overrides_a_conflicting_x-tenant-env-var_entry`.
3. `docs/SPEC-EXTENSIONS.md` splits the previous single example into two coherent ones: a runtime-override example using Atlassian `{workspace}` and a build-time default example using Gmail `{userId}`.

Local verification: `go test ./...`, `go vet ./...`, and `scripts/golden.sh verify` (18 of 18) all pass.

brysonerandall-a11y:
Adding one more commit (317bbcb4) that is adjacent to this PR's stated scope but only became visible because of the extension itself.

**What:** moved `pipeline.WriteToolsManifest` into `runGenerateProject` so a fresh `printing-press generate` run emits `tools-manifest.json` alongside `cmd/` and `internal/mcp/`. Previously the manifest was only written by `mcp-sync` and `publish`.

**Why this PR:** adopting `x-path-template-env-vars: { userId: { default: me } }` on a Gmail spec, then running `generate --force`, leaves the runtime correctly serving `/users/me/...` while the stale `tools-manifest.json` on disk still advertises `/users/{userId}/...` with a `userId` path parameter. The two surfaces diverge silently until the user runs `mcp-sync`. Hitting that during the Gmail adoption is what surfaced this.

**Side effect on goldens:** `generate-golden-api`'s scorecard now evaluates the `mcp_description_quality` dimension (3/10) because the manifest exists. The dimension moves out of `unscored_dimensions`, total drops 82 to 80 — real signal that was previously hidden, not a regression.

**Happy to split** this into a separate PR if you'd rather keep the original scope tight; let me know and I will pull it back to a new branch.

tmchow:

PR #1406, opened 2026-05-14
fix(cli): derive doctor probe path from auth.verify_path or me-shaped endpoint

### LINKED ISSUE #1118: [P2] Doctor reachability picks a real auth-probe path, not /
## Summary

`doctor` probes `/` for reachability, which for subdomain-customer APIs is the marketing/help-center page — often Cloudflare-fronted with a 403 interstitial. `doctor` reports "API: blocked by Cloudflare interstitial" while the actual API at `/api/v2/...` returns 200.

**Reproduction:** Generate Zendesk against any subdomain; run `doctor --json` — `api` field shows the interstitial message, `auth` shows "configured" and credentials work. The two signals contradict each other.

## Cross-API evidence

- **Zendesk** — `<sub>.zendesk.com/` is the help center, `<sub>.zendesk.com/api/v2/...` is the API.
- **Notion** — `notion.so/` is marketing; the analog holds for any vendor with a CDN-fronted root.
- **Linear** — `linear.app/` is the dashboard behind a bot-detection layer; `api.linear.app/graphql` is the API.

## Proposed fix

In `internal/generator/templates/doctor.go.tmpl` reachability step, replace the unconditional `c.Get("/", nil)` with a probe-path resolver:

1. If the spec declares `x-verify-path` (OpenAPI) or `auth.verify_path` (internal-spec), probe that.
2. Else, prefer a known small auth-only GET endpoint (a `me`-shaped path with no required params), picked at generate-time.
3. Fall back to `/` only when neither heuristic produces a candidate.

The probe path the generator picks should be emitted as a const in the generated `doctor.go` so the runtime check has a single string to send.

## Acceptance

- Positive: generate from a spec with an authed `me`-style endpoint (Zendesk's `/api/v2/users/me.json`, GitHub's `/user`, Linear's `viewer { id }`); doctor probes that path and reports `api: reachable` against the live API.
- Negative: spec with no candidate path falls back to current `/` behavior (still useful as a connectivity check).

## Component

`comp:generator` — `internal/generator/templates/doctor.go.tmpl`

tmchow:
## Summary

Subdomain-customer APIs (Zendesk, Notion, Linear) front `/` with the marketing or help-center page, which Cloudflare often answers with a 403 interstitial. The generated `doctor` previously probed `/` unconditionally and reported `api: blocked by Cloudflare interstitial` while the actual API at `/api/...` was reachable, contradicting the `credentials: configured` signal.

This change derives a real API probe path at generate-time. The doctor template already had a `{{- if .HealthCheckPath}}` branch — the fix just populates `spec.HealthCheckPath` in `Generator.Generate()` when unset.

Priority chain:
1. `spec.HealthCheckPath` — explicit override, never clobbered.
2. `Auth.VerifyPath` — operator-vetted authenticated GET; an unauth probe returns 401, which the existing doctor template classifies as reachable.
3. Heuristic me-shaped GET endpoint discovered in the spec (no required params, no `{placeholders}`). Patterns: `users/me.json`, `users/me`, `users/@me`, `current_user`, `me.json`, `whoami`, `viewer`, `account`, `self`, `user`, `me`. Order is meaningful — specific tails win over `me`.
4. Empty → template falls back to `/`. Preserves pre-derivation behavior for specs with nothing better.

`findEndpointMatch` is added as a sibling to the existing `anyEndpointMatches` walker. It iterates resource and endpoint keys in sorted order so the derived path stays deterministic across runs even when two endpoints in different resources match the same tail.

Closes #1118

## Test plan

- [x] `go test ./...` — 4089 tests pass (added 14 unit + integration tests for the priority chain, heuristic table, sub-resource walk, sub-string rejection, same-tail determinism (50-run loop), idempotency)
- [x] `scripts/golden.sh verify` — 18 cases pass unchanged. Existing golden specs don't declare `Auth.VerifyPath` and don't ship me-shaped endpoints, so emitted output for those fixtures is byte-identical; the new derivation paths are covered by string-asserting `TestGeneratedDoctor_*` cases in `doctor_health_path_test.go`.
- [x] `golangci-lint run ./...` — 0 issues
- [x] `go vet ./...` — clean

### Golden-fixture decision (per AGENTS.md "Generator Output Stability")

New deterministic generator behavior (HealthCheckPath derivation) is covered by string-asserting integration tests in `doctor_health_path_test.go` that lock all four branches of the rendered `doctor.go`:
- `TestGeneratedDoctor_DerivesHealthCheckPathFromVerifyPath`
- `TestGeneratedDoctor_DerivesHealthCheckPathFromMeEndpoint`
- `TestGeneratedDoctor_KeepsExplicitHealthCheckPath`
- `TestGeneratedDoctor_NoCandidateFallsBackToRoot`

A new golden fixture would duplicate this coverage at higher maintenance cost; skipped intentionally. Existing goldens continue to verify the no-derivation `/`-probe branch.

### DIFF
--- a/internal/generator/doctor_health_path.go
+++ b/internal/generator/doctor_health_path.go
@@ -1,0 +1,81 @@
+package generator
+
+import (
+	"strings"
+
+	"github.com/mvanhorn/cli-printing-press/v4/internal/spec"
+)
+
+// meShapedPathTails enumerates path suffixes that conventionally identify a
+// small auth-only GET endpoint suitable as a reachability probe (the same
+// shape Auth.VerifyPath calls out in spec.go). Order is meaningful only when
+// multiple endpoints qualify in the same spec — earliest match wins so that
+// the more specific `users/me` beats a bare `me` on APIs that ship both.
+var meShapedPathTails = []string{
+	"users/me.json",
+	"users/me",
+	"users/@me",
+	"current_user",
+	"me.json",
+	"whoami",
+	"viewer",
+	"account",
+	"self",
+	"user",
+	"me",
+}
+
+// deriveHealthCheckPath chooses the path the generated `doctor` command should
+// hit for its unauthenticated reachability probe.
+//
+// Priority:
+//  1. spec.HealthCheckPath when set (explicit user override, never clobbered).
+//  2. Auth.VerifyPath when set (already vetted by the operator as a known-good
+//     authenticated GET; an unauthenticated probe against it returns 401 from
+//     the real API surface, which the doctor classifies as "reachable").
+//  3. A heuristic me-shaped GET endpoint discovered in the spec (no required
+//     path/query params and a tail matching meShapedPathTails).
+//  4. Empty string. The template falls back to probing `/`, preserving the
+//     pre-derivation behavior for specs with nothing better to offer.
+//
+// Returning empty in case 4 (rather than `"/"`) keeps the existing template
+// branch in doctor.go.tmpl authoritative for the fallback — one source of
+// truth for the probe path the runtime sends.
+func deriveHealthCheckPath(s *spec.APISpec) string {
+	if s == nil {
+		return ""
+	}
+	if s.HealthCheckPath != "" {
+		return s.HealthCheckPath
+	}
+	if s.Auth.VerifyPath != "" {
+		return s.Auth.VerifyPath
+	}
+	for _, tail := range meShapedPathTails {
+		if e, ok := findEndpointMatch(s, func(e spec.Endpoint) bool {
+			return isMeShapedEndpoint(e, tail)
+		}); ok {
+			return e.Path
+		}
+	}
+	return ""
+}
+
+func isMeShapedEndpoint(e spec.Endpoint, tail string) bool {
+	if !strings.EqualFold(e.Method, "GET") {
+		return false
+	}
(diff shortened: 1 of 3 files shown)


tmchow:
Acknowledging Greptile's two non-blocking observations:

1. **No memoization when `deriveHealthCheckPath` returns `""`** — `g.Spec.HealthCheckPath` stays empty and the heuristic walk re-runs on each `Generate()` call. The walk is O(11 tails × tree size) for specs that don't trigger any tier, runs once per CLI generation in practice, and the empty result is the negative-cache value with no observable effect (template falls back to `/`). Memoizing the empty result would require either a sentinel string (bleeds derivation state into the spec) or a separate boolean (parallel-state pattern AGENTS.md discourages). Accepted as-is.

2. **`user` and `account` tails are broad** — the GET method + no-required-params filter + boundary-anchored `/<tail>` suffix match keep this conservative in practice. The doctor's reachability classifier treats any HTTP response as "reachable", so a false-positive endpoint pick still produces a correct reachability verdict; the credentials probe (separate path) handles auth-validity classification. Accepted as-is.

Also note: the template's `if !strings.HasPrefix(healthPath, "/") { healthPath = "/" + healthPath }` block at `doctor.go.tmpl:366-368` already normalizes the probe path, so the related "what if the spec ships a path without `/`?" edge case is template-handled.

PR #1445, opened 2026-05-15
fix(cli): kebab-case SKILL.md and README endpoint examples

### LINKED ISSUE #1270: SKILL.md emits snake_case subcommand names; Cobra commands are kebab-case
## Summary
SKILL.md examples emit snake_case subcommand names (`dns get_hosts`, `set_email_forwarding`) but the actual generated Cobra subcommands use kebab-case (`dns get-hosts`, `set-email-forwarding`). `verify-skill` flags this as an error because it treats `dns` as the command and `get_hosts` as a positional arg.

## Repro
Generate a CLI with any multi-word operation (e.g., `dns.getHosts`, `dns.setEmailForwarding`). The generator emits `Available Commands: get-hosts` under `dns --help`, but SKILL.md `## Common commands` lists `namecheap-pp-cli dns get_hosts`.

## Proposed fix
Generator template that emits SKILL.md "Common commands" and example lines must use the same `kebabCase(opID)` helper the Cobra command-name generator uses. They should share one helper.

## Acceptance
- Positive: generate a CLI with operation `getEmailForwarding`; SKILL.md contains `dns get-email-forwarding`. verify-skill exits 0 without manual edits.
- Negative: single-word operations (`check`, `create`) are unchanged.

## Evidence
- namecheap-pp-cli run 2026-05-13: 21 snake→kebab perl replacements in SKILL.md to pass verify-skill.

## From retro
- /Users/marcos/printing-press/manuscripts/namecheap/20260513-063855/proofs/20260513-073354-retro-namecheap-pp-cli.md
- Finding F1; WU-1; Frequency: every API with multi-word operations.

tmchow:
## Summary
- Route SKILL.md and README per-endpoint emission through the same `kebab` template helper Cobra already uses for the command `Use:` string, so spec keys like `get_hosts` or `setEmailForwarding` render as `get-hosts` / `set-email-forwarding` in user-facing docs instead of phantom command paths.
- Kebab-case the endpoint name inside `firstCommandExample` for the same reason, so the first-command example block stays in sync with the actual cobra command tree.
- Add coverage: positive cases for snake_case + camelCase multi-word endpoints in both SKILL.md and README.md, and a negative case asserting single-word endpoints (`create`, `list`, `get`, `check`) remain unchanged.

Closes #1270

## Description
The generator's `command_endpoint.go.tmpl` registers each command with `Use: "{{kebab .EndpointName}}{{positionalArgs .Endpoint}}"`, so a spec endpoint key of `getEmailForwarding` ships as the cobra subcommand `get-email-forwarding`. The SKILL.md and README "Command Reference" templates iterated the same endpoint map and emitted `{{$eName}}` directly — the raw spec key — so docs advertised `<cli> dns get_hosts` while the binary only knew `<cli> dns get-hosts`. `verify-skill` rejected the example as a phantom path, and the polish skill batch-fixed those lines with mechanical `_` → `-` replacements after the fact on every CLI with multi-word operation names (issue body notes 21 fixes on namecheap, 14 on openart).

This change moves the doc-side emission to the same kebab helper as the cobra registration so both surfaces share one transform. `firstCommandExample` gets the same treatment so the SKILL.md first-command example block doesn't fall out of sync.

## Test plan
- [x] `go test ./...` — 4201 passed in 34 packages
- [x] `scripts/golden.sh verify` — 19 cases pass unchanged (existing fixtures use single-word endpoint names, so the kebab pass is a no-op there)
- [x] `go vet ./...`, `golangci-lint run ./...`, `go build -o ./printing-press ./cmd/printing-press`
- [x] New unit test cases in `first_command_example_test.go` cover multi-word snake_case and camelCase endpoint keys.
- [x] New integration test `skill_endpoint_kebab_test.go` asserts both SKILL.md and README emit `dns get-hosts` / `audio create-speech` and never the snake_case / camelCase phantom paths, plus a negative case that single-word endpoints are unchanged.

### DIFF
--- a/internal/generator/first_command_example.go
+++ b/internal/generator/first_command_example.go
@@ -42,7 +42,7 @@ func firstCommandExample(resources map[string]spec.Resource) string {
 	pathFor := func(rName string, r spec.Resource, eName string, ep spec.Endpoint) string {
 		parts := []string{rName}
 		if !isPromotableSingleEndpoint(rName, r) {
-			parts = append(parts, eName)
+			parts = append(parts, toKebab(eName))
 		}
 		parts = append(parts, readmeExampleArgs(ep)...)
 		return strings.Join(parts, " ")
--- a/internal/generator/first_command_example_test.go
+++ b/internal/generator/first_command_example_test.go
@@ -280,6 +280,38 @@ func TestFirstCommandExampleHonorsPromotion(t *testing.T) {
 			},
 			want: "feed 2026-04-27",
 		},
+		{
+			// Issue #1270: snake_case endpoint keys must be kebab-cased so
+			// the example matches the actual cobra `Use:` string (which is
+			// already kebab) instead of advertising a phantom command path.
+			name: "multi-word snake_case endpoint key is kebab-cased",
+			resources: map[string]spec.Resource{
+				"dns": {
+					Endpoints: map[string]spec.Endpoint{
+						"get_hosts":            {Method: "GET", Path: "/dns/hosts"},
+						"set_email_forwarding": {Method: "POST", Path: "/dns/forwarding"},
+					},
+				},
+			},
+			want: "dns get-hosts",
+		},
+		{
+			// Same kebab pass for camelCase endpoint keys. Mirrors the
+			// command_endpoint.go.tmpl `Use: {{kebab .EndpointName}}` rule
+			// for cobra command names. The preferred-verb scan misses
+			// `createSpeech` and `cancelJob`, so the fallback alphabetical
+			// pick is `cancelJob` (c < other letters).
+			name: "multi-word camelCase endpoint key is kebab-cased",
+			resources: map[string]spec.Resource{
+				"audio": {
+					Endpoints: map[string]spec.Endpoint{
+						"createSpeech": {Method: "POST", Path: "/audio/speech"},
+						"cancelJob":    {Method: "POST", Path: "/audio/cancel"},
+					},
+				},
+			},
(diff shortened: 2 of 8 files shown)


tmchow:
Addressed the P2 finding: mirrored the `PromotedResourceNames` guard from `skill.md.tmpl`'s Command Reference loop in `readme.md.tmpl`'s Commands section. Now both surfaces emit the leaf form (`<cli> resource`) when cobra promotes a single-op resource, and the per-endpoint form (`<cli> resource <kebab-endpoint>`) otherwise.

Pushed in 4d6b... — extended the integration test (`TestReadmePromotionGuardMirrorsSkill`) and updated `TestOutputFormatsUsesRealCommandExample` to reflect promotion. Two README goldens (`generate-golden-api`, `generate-golden-api-rich-auth`) updated: previously phantom `currencies list`/`public get-status`/`items list` lines now correctly render as `currencies`/`public`/`items`. Full test + golden + vet + lint suites green locally.

tmchow:
@mergifyio queue

Re-queueing now that #1459 has merged the contract-test fix that was causing `test (pipeline)` failures on every in-place rebase. No rebase needed on this PR — Mergify will pull current main during in-place.

tmchow:
@mergifyio update

Branch is on an old main and CI is stuck at FAILURE from before #1459 landed the contract-test fix. `queue_conditions` needs `check-success = build-and-test` and `check-success = test` to be green before re-entering the queue. Rebasing this branch against current main will retrigger CI on a SHA that includes the fix.

PR #1449, opened 2026-05-15
fix(cli): strip root-level binaries from staged publish tree

### LINKED ISSUE #1396: skill(publish): binary cleanup misses <api-slug>-pp-mcp; root-level executables not stripped from staged dir
> Filed by `/printing-press-retro` from a run of **PushPress /v3** (scorecard 89/100, 100% verify pass).

## Summary

The publish skill's binary cleanup step removes a hardcoded pair (`<api-slug>`, `<cli-name>` — i.e. `pushpress`, `pushpress-pp-cli`) but never the `<api-slug>-pp-mcp` binary that the generator template emits as a peer. Every CLI with MCP vision set produces a second binary at the staged-dir root whenever the agent runs `go build ./cmd/...` (which the generated `Makefile` `build-mcp` target and several validation steps do), and that 21MB Mach-O slips into the publish PR unless the operator catches it by hand. PushPress: caught manually before push.

## Where to look

- **Component:** skill
- **Likely area:** `skills/printing-press-publish/SKILL.md` line ~423 — the `rm -f` cleanup line. Optional defense-in-depth landing site: `internal/cli/publish.go` `runPackage` (or whichever function runs after `pipeline.CopyDir(dir, outCLIDir)`), alongside the existing `os.RemoveAll(filepath.Join(outCLIDir, "build"))` strip.
- **Triggered when:** Any CLI with `VisionSet.MCP=true` (the default; the generator emits `cmd/<api-slug>-pp-mcp/` per `internal/generator/templates/makefile.tmpl` lines 19-25) where any preceding step builds binaries at the cli-dir root.

## What we observed

The publish skill cleanup line in `skills/printing-press-publish/SKILL.md`:

```bash
rm -f "$PUBLISH_REPO_DIR/library/<category>/<api-slug>/<api-slug>" "$PUBLISH_REPO_DIR/library/<category>/<api-slug>/<cli-name>"
```

For the PushPress run that means:

```bash
rm -f .../pushpress .../pushpress-pp-cli
```

`pushpress` is a phantom path (the generator's binary convention is `<api-slug>-pp-cli` and `<api-slug>-pp-mcp`, never bare `<api-slug>`). `pushpress-pp-cli` is one of the two real binaries. `pushpress-pp-mcp` is the other — and the cleanup misses it.

Concrete state from this run, in `~/printing-press/library/pushpress/`:

```
-rwxr-xr-x  1 alexpuckhaber  staff  18696690 May 14 09:18 pushpress-pp-cli
-rwxr-xr-x  1 alexpuckhaber  staff  21595250 May 14 08:20 pushpress-pp-mcp
```

Both are Mach-O 64-bit executable arm64. Verified via `file`.

The publish-package code that runs *before* the skill's cleanup already strips `build/` via `os.RemoveAll(filepath.Join(outCLIDir, "build"))` (see `internal/cli/publish.go` after the `pipeline.CopyDir(dir, outCLIDir)` call). That strip catches the `<cliDir>/build/stage/bin/<api-slug>-pp-mcp` that `autoBundleForHost` produces. It does NOT catch root-level binaries that get there via the agent or `Makefile` running `go build ./cmd/...` in the cli dir.

The test in `internal/cli/publish_test.go` line ~540 (`TestPublishPackageStripsBuildDir`-shape) only asserts `build/` is gone — there's no analogous test for root-level executables.

## Suspected root cause

**Certain.** The skill's cleanup enumerates two filenames where it should match by convention. Two layered cheap fixes apply.

## Suggested direction

Two layers; either suffices, both are better:

**1. Skill-level (durable for the next publish run).** Replace the enumerated `rm -f` with one of:

- Explicit pair: `rm -f .../<api-slug>-pp-cli .../<api-slug>-pp-mcp` (no false-positive on weirdly named user files).
- Mode-based: `find "$PUBLISH_REPO_DIR/library/<category>/<api-slug>" -maxdepth 1 -type f -perm +111 -delete` (catches future binary names without code change).

**2. Binary-level (defense in depth).** In `internal/cli/publish.go` `runPackage`, after `pipeline.CopyDir(dir, outCLIDir)` and the `build/` strip, walk the staged-dir root and refuse to stage executable files. Mirrors the existing `os.RemoveAll(...)` build strip pattern. Either return an error pointing at the offending file:line, or silently remove it (the file mode is the signal — committing executables to the library is never correct).

The skill change is the immediate stop-gap; the binary-level guard is the long-term floor that catches this even when a future skill rewrite forgets the cleanup line.

## Acceptance criteria

- **positive:** A publish-package run with `<api-slug>-pp-cli`, `<api-slug>-pp-mcp`, and any other executable at the cli-dir root produces a staged tree with no executables at the root. Reproduce with a fresh GHL or Notion publish after the fix lands.
- **negative:** Non-executable root files (`README.md`, `go.mod`, `spec.yaml`, `.printing-press.json`, `manifest.json`, `tools-manifest.json`, `dogfood-results.json`, `workflow-verify-report.json`, `LICENSE`, `NOTICE`, `AGENTS.md`, `SKILL.md`) are preserved.
- **regression test (if binary-level guard added):** Sibling case to `TestPublishPackageStripsBuildDir` in `internal/cli/publish_test.go` asserting executables at the root are stripped (or refused).

## Frequency

most APIs — every printed CLI with `VisionSet.MCP=true`. Named with evidence:

- **GHL CLI** (`library/sales-and-crm/ghl/`, manuscripts/ghl/20260513-202134) — `cmd/ghl-pp-mcp/` exists, MCP surface shipped, same publish-skill flow.
- **Notion CLI** (any library entry under `productivity/notion`) — MCP vision set, identical generator template path.
- **Linear CLI** (`library/productivity/linear/`) — same shape; `cmd/linear-pp-mcp/` per generator convention.

The MCP vision is the default; the rare CLI shipped without MCP would not produce the second binary but still benefits from the defense-in-depth check.

## Complexity

small

## Scope boundary

Does NOT include cleanup of nested `cmd/<api-slug>-pp-*/` source directories (those are the generated entry points and must stay). Does NOT change `build/` strip (already covered). Does NOT change CI release-binary behavior — those are built by CI from source, not committed to the tree.

## Dependencies

None.

## Related issues

- #971 — open-issue (`related-area`): publish SKILL Step 6 staged_dir guidance to api-slug keying. Different symptom (path keying), same skill file, same family of slug-vs-cli-name confusions; an editor touching the cleanup line should re-read #971 to see whether the path-keying fix and the binary-cleanup fix want to land together.

---

tmchow:
## Summary

Fixes #1396. The publish skill's binary cleanup listed `<api-slug>` and `<cli-name>`, missing the `<api-slug>-pp-mcp` peer the generator emits for every MCP-enabled CLI. When the operator (or any preceding step) ran `go build ./cmd/...` at the CLI root, the 21MB MCP binary slipped into the staged publish tree.

## Changes

- `internal/cli/publish.go`: after `CopyDir(dir, outCLIDir)` and the existing `build/` strip, remove the named root-level binaries from the staged tree. New helper `stagedBinaryNames` returns the canonical set (bare slug, `<api-slug>-pp-cli`, `<cli-name>`, `<api-slug>-pp-mcp`).
- `skills/printing-press-publish/SKILL.md`: Step 6 cleanup line now also lists `<api-slug>-pp-mcp` so following the SKILL verbatim matches the code-level guarantee.
- New tests:
  - `TestPublishPackageStripsRootBinaries`: writes bare-slug + pp-cli + pp-mcp at the source root, packages, asserts the source binaries survive and the staged tree contains none of them.
  - `TestStagedBinaryNamesDeduplicates`: pins the helper's name list and dedup behavior.

## Test plan

- [x] `go build -o ./printing-press ./cmd/printing-press`
- [x] `go test ./internal/cli/... ./internal/pipeline/...` (1683 passed)
- [x] `go vet ./...` (clean)
- [x] `go fmt ./...` (no changes)
- [x] `golangci-lint run ./internal/cli/...` (0 issues)
- [x] `scripts/golden.sh verify` (19 cases, no goldens cover root-level binaries)

Closes #1396

### DIFF
--- a/internal/cli/publish.go
+++ b/internal/cli/publish.go
@@ -361,6 +361,19 @@ func newPublishPackageCmd() *cobra.Command {
 				return &ExitError{Code: ExitPublishError, Err: fmt.Errorf("stripping build dir: %w", err)}
 			}
 
+			// Strip root-level binaries that local `go build ./cmd/...` (or
+			// `make build` / `make build-mcp` without `-o bin/...`) drops at
+			// the CLI dir root. The skill cleanup line tries to catch these
+			// after the copy, but routinely misses `<api-slug>-pp-mcp`; this
+			// strip applies the same convention before the copy so the staged
+			// tree is binary-free regardless of which path the operator took.
+			for _, name := range stagedBinaryNames(cliName, dirName) {
+				if err := os.Remove(filepath.Join(outCLIDir, name)); err != nil && !os.IsNotExist(err) {
+					cleanupOnFailure()
+					return &ExitError{Code: ExitPublishError, Err: fmt.Errorf("stripping staged binary %s: %w", name, err)}
+				}
+			}
+
 			// Rewrite go.mod module path if --module-path is set
 			if modulePath != "" {
 				oldModPath := cliName // generated CLIs use bare CLI name as module path
@@ -992,6 +1005,35 @@ func buildArtifactCandidates(dir, cliName string) []string {
 		filepath.Join(dir, cliName),
 		filepath.Join(dir, "cmd", cliName, cliName),
 	}
+}
+
+// stagedBinaryNames returns the root-level filenames that local builds
+// (`go build ./cmd/...`, `make build`, `make build-mcp` without `-o`)
+// drop alongside the source tree. The Makefile template emits
+// `<api-slug>-pp-cli` and `<api-slug>-pp-mcp` peers, and verifier paths
+// may also leave a bare `<api-slug>` artifact. The list is intentionally
+// a small, named superset rather than a glob so the strip step never
+// removes a tracked source file by accident.
+func stagedBinaryNames(cliName, apiSlug string) []string {
+	seen := map[string]struct{}{}
+	var names []string
+	add := func(n string) {
+		if n == "" {
+			return
+		}
+		if _, ok := seen[n]; ok {
+			return
+		}
+		seen[n] = struct{}{}
+		names = append(names, n)
+	}
+	add(apiSlug)
+	add(cliName)
+	if apiSlug != "" {
+		add(apiSlug + "-pp-cli")
+		add(apiSlug + "-pp-mcp")
+	}
+	return names
 }
 
 type fileSnapshot struct {
--- a/internal/cli/publish.go
+++ b/internal/cli/publish.go
@@ -363,10 +363,11 @@ func newPublishPackageCmd() *cobra.Command {
 
 			// Strip root-level binaries that local `go build ./cmd/...` (or
 			// `make build` / `make build-mcp` without `-o bin/...`) drops at
-			// the CLI dir root. The skill cleanup line tries to catch these
-			// after the copy, but routinely misses `<api-slug>-pp-mcp`; this
-			// strip applies the same convention before the copy so the staged
-			// tree is binary-free regardless of which path the operator took.
+			// the CLI dir root. The skill's downstream `cp -r ... PUBLISH_REPO_DIR`
+			// has a parallel `rm -f` line that historically missed
+			// `<api-slug>-pp-mcp`; doing the strip here on the staged copy
+			// makes the publish path binary-free regardless of which downstream
+			// step (skill or other tooling) the operator runs next.
 			for _, name := range stagedBinaryNames(cliName, dirName) {
 				if err := os.Remove(filepath.Join(outCLIDir, name)); err != nil && !os.IsNotExist(err) {
 					cleanupOnFailure()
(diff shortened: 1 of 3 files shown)


tmchow:
@mergifyio queue

Re-queueing now that #1459 has merged the contract-test fix that was causing `test (pipeline)` failures on every in-place rebase. No rebase needed on this PR — Mergify will pull current main during in-place.

tmchow:
@mergifyio update

Branch is on an old main and CI is stuck at FAILURE from before #1459 landed the contract-test fix. `queue_conditions` needs `check-success = build-and-test` and `check-success = test` to be green before re-entering the queue. Rebasing this branch against current main will retrigger CI on a SHA that includes the fix.

PR #1462, opened 2026-05-15
fix(cli): query generic resources_fts from search empty-type branch

### LINKED ISSUE #1390: generator: search.go template's empty-type branch only queries one typed FTS source
## Summary

Generated `internal/cli/search.go` has a comment **"Search all FTS-enabled tables individually to avoid duplicates"** followed by code that only calls `db.SearchRoles(query, limit)` — every other FTS-enabled table is silently ignored when the user runs `<cli> search "<query>"` without `--type`.

This retro's Asana CLI had FTS5 indexed over 27,704 rows in `resources_fts` plus per-resource typed FTS tables; `search "<example-query>"` returned zero results despite direct `SELECT ... FROM resources_fts WHERE resources_fts MATCH '<example-query>'` returning 249 hits.

## Root cause

The template that emits `internal/cli/search.go` enumerates FTS-enabled typed tables. For this CLI it enumerated only `roles_fts` in the `case "":` branch. Either the enumeration logic stops at the first match, or the loop body wasn't templated to repeat per typed table. The comment-vs-code discrepancy is the giveaway.

## Cross-API blast radius

Every CLI emitted by the affected template where more than one typed FTS table exists.

- **Asana** (confirmed this retro)
- **Jira** (published): typed FTS for issues, projects, users — same template
- **Linear** (published): issues, projects, teams typed tables — same template

## Fallback if unfixed

Zero. User runs `search`, gets empty results, has no signal that indexes exist but aren't being queried. Returns `results: []` with no warning.

## Durable fix (in-session validated)

Patch `search.go` template so `case "":` queries `db.Search` (generic `resources_fts`) in addition to any typed `Search<Resource>` helpers it enumerates. In-session patch added the `db.Search` call after `SearchRoles` and merged via string-key dedup — after rebuild, `search "<example-query>"` returned matching project + tasks rows.

## Acceptance criteria

- **Positive**: Generate a CLI for a multi-resource API, sync to populate multiple typed FTS tables AND the generic `resources_fts`. Run `<cli> search "<term known in multiple tables>" --data-source local`, assert non-empty results.
- **Negative**: Single-resource API — assert no change.
- **Code/comment alignment**: The comment "Search all FTS-enabled tables individually" must accurately describe the code below it.

## Scope boundary

Does NOT need perfect cross-index dedup — string-key dedup is sufficient. Does NOT need to add new FTS schemas; just iterate existing ones.

## Complexity

Small (template fix; loop over the FTS-table list the template already knows about).

## Artifacts

- [Retro doc](https://files.catbox.moe/unp7t0.md)
- [Scrubbed manuscripts](https://files.catbox.moe/9fvwft.zip)
- [Scrubbed CLI source](https://files.catbox.moe/nfpusx.zip)

Anti-pattern: comments that contradict code. The bug-discovery signal is the discrepancy.

tmchow:
## Summary

Fixes #1390. The generated \`internal/cli/search.go\`'s \`case \"\":\` branch iterated only the typed \`Search<Resource>\` helpers, so rows indexed in \`resources_fts\` (the generic FTS table) but not in any typed FTS table silently returned zero from \`<cli> search \"<query>\"\` without \`--type\`. Confirmed on Asana CLI where direct \`SELECT FROM resources_fts MATCH ...\` returned 249 hits but \`search \"<term>\"\` reported zero.

## Changes

- \`internal/generator/templates/search.go.tmpl\`: append a \`db.Search(query, limit)\` call to the existing seen-map dedup loop in the \`case \"\":\` branch. The generic resources_fts is now queried alongside the typed helpers, and dedup-by-raw-JSON keeps rows that are indexed in multiple FTS sources from appearing twice.
- \`internal/generator/generator_test.go\`: \`TestSearchTemplateEmptyTypeQueriesGenericFTS\` pins the call site and the contextual error message inside the empty-type branch.

## Test plan

- [x] \`go build ./...\`
- [x] \`go test ./internal/generator/... -run TestSearchTemplate\` (2 pass)
- [x] \`go vet ./...\` clean
- [x] \`go fmt ./...\` no changes
- [x] \`scripts/golden.sh verify\` (19 cases pass; no goldens cover search.go output)
- [x] \`golangci-lint run ./internal/generator/...\` (0 issues)

Closes #1390

### DIFF
--- a/internal/generator/templates/search.go.tmpl
+++ b/internal/generator/templates/search.go.tmpl
@@ -176,7 +176,12 @@ In local mode: searches locally synced data only.`,
 {{- end}}
 {{- end}}
 			case "":
-				// Search all FTS-enabled tables individually to avoid duplicates.
+				// Search every FTS-enabled source — typed per-resource tables
+				// AND the generic resources_fts — and dedup by raw JSON so a
+				// row indexed in multiple FTS sources appears once. Without
+				// the generic-search call, rows that landed in resources_fts
+				// but not in any typed FTS table (e.g., a resource whose sync
+				// populated only the generic index) silently return zero.
 				seen := make(map[string]bool)
 				_ = seen // prevent unused error when no FTS tables exist
 {{- range .Tables}}
@@ -196,6 +201,19 @@ In local mode: searches locally synced data only.`,
 				}
 {{- end}}
 {{- end}}
+				{
+					partial, searchErr := db.Search(query, limit)
+					if searchErr != nil {
+						return fmt.Errorf("search resources_fts failed: %w", searchErr)
+					}
+					for _, r := range partial {
+						key := string(r)
+						if !seen[key] {
+							seen[key] = true
+							results = append(results, r)
+						}
+					}
+				}
 			default:
 				// Unrecognized type — fall back to generic search
 				results, err = db.Search(query, limit)
--- a/internal/generator/generator_test.go
+++ b/internal/generator/generator_test.go
@@ -10203,8 +10203,11 @@ func TestSearchTemplateEmptyTypeQueriesGenericFTS(t *testing.T) {
 	caseEmptyIdx := strings.Index(body, `case "":`)
 	require.GreaterOrEqual(t, caseEmptyIdx, 0, "search.go.tmpl must have a case \"\": branch")
 
-	defaultIdx := strings.Index(body[caseEmptyIdx:], "default:")
-	require.GreaterOrEqual(t, defaultIdx, 0, "search.go.tmpl must have a default: after case \"\":")
+	// Anchor on the column-zero `default:` that closes the empty-type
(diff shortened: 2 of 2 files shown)


internal/generator/templates/search.go.tmpl
```diff
@@ -196,6 +201,19 @@ In local mode: searches locally synced data only.`,
 				}
 {{- end}}
 {{- end}}
+				{
+					partial, searchErr := db.Search(query, limit)
```
tmchow, line 205: Leaving the unconditional db.Search(query, limit) call as-is for parity with the existing typed Search<X> loop above — each iteration there also passes the full limit. The seen-map dedup discards duplicates so extra rows are work-only, not correctness-affecting; capping would deviate from the surrounding pattern. Accepting for later if we revisit the limit semantics across the whole branch.

PR #1471, opened 2026-05-15
fix(cli): short-circuit sync on dry-run sentinel response

### LINKED ISSUE #1382: generator: sync --dry-run mode falls through to upsert with mock-empty data, exits 1
> Filed by `/printing-press-retro` from a run of **Petstore** (scorecard 84/100, 100% verify pass).

## Summary

Sync's `--dry-run` mode short-circuits the API call (prints the would-be HTTP request without sending it) but then proceeds to the upsert path with a mock-empty response. Per-resource errors fire ("missing id for pet/store") and sync exits 1. Any narrative quickstart that includes `sync` (which the SKILL recommends — local-store transcendence features depend on it) blocks shipcheck because `validate-narrative --strict --full-examples` automatically appends `--dry-run` to advertised commands.

## Where to look

- **Component:** generator
- **Likely area:** `internal/generator/templates/internal/cli/sync.go.tmpl` (and any equivalent intermediate file)
- **Triggered when:** `<cli> sync [--full] --dry-run` is invoked under `PRINTING_PRESS_VERIFY=1` (validate-narrative does this automatically)

## What we observed

Run produced by Phase 4 shipcheck on Petstore CLI:

```
$ PRINTING_PRESS_VERIFY=1 ./petstore-pp-cli sync --dry-run
{"event":"sync_start","resource":"pet"}
{"event":"sync_start","resource":"store"}
{"event":"sync_warning","resource":"pet","reason":"resource_not_incremental","message":"endpoint does not declare a temporal filter parameter; incremental sync has no effect for this resource"}
GET /api/v3/pet/findByStatus
  ?limit=100

(dry run - no request sent)
{"event":"sync_error","resource":"pet","error":"missing id for pet"}
{"event":"sync_error","resource":"store","error":"missing id for store"}
{"event":"sync_summary","total_records":0,"resources":2,"success":0,"warned":0,"errored":2,"duration_ms":4}
Error: 2 resource(s) failed to sync
```

Validate-narrative output blocking shipcheck:

```
FAILED [quickstart]: petstore-pp-cli sync --full
  → full example failed: petstore-pp-cli sync --full --dry-run: exit status 1
```

The "(dry run - no request sent)" line confirms the network call WAS skipped — but the upsert path runs anyway against the mock-empty response and fails because there's no ID to upsert.

## Suspected root cause

The dry-run guard in the sync template prints the request and skips the network call but does NOT skip the per-record upsert that depends on the response. Likely a control-flow gap: the guard short-circuits at the http call, then the loop proceeds to upsert with a zero-record payload.

## Suggested direction

In the sync template's per-resource loop, after the dry-run "(dry run - no request sent)" print, return early with a synthetic success event (e.g., `{"event":"sync_dryrun","resource":"<name>"}`) instead of falling through to the upsert. Two-line change.

## Acceptance criteria

- positive: Generate any CLI from spec; `PRINTING_PRESS_VERIFY=1 <cli> sync --dry-run` exits 0; no `sync_error` events in output
- negative: With `--dry-run` removed and a live API, sync still upserts normally

## Frequency

every API with sync (i.e., every printed CLI)

## Complexity

small

## Scope boundary

Do NOT change sync's live-mode behavior. Do NOT change validate-narrative's `--full-examples` flag (it should keep auto-appending `--dry-run`).

## Dependencies

None.

## Related issues

None.

## Artifacts

| Artifact | Link |
|----------|------|
| Retro document | https://files.catbox.moe/zbpk24.md |
| Manuscripts (research + proofs) | https://files.catbox.moe/9aj1sv.zip |
| Generated CLI source code | https://files.catbox.moe/5cs40o.zip |

---

tmchow:
## Summary

Fixes #1382. \`client.dryRun\` returns \`{\"dry_run\": true}\` instead of a real response when \`--dry-run\` is set. The sync template fell through to \`upsertSingleObject\` against this sentinel, failing with \`missing id for <resource>\` because the sentinel has no items and no id. Every CLI shipped with a \`sync\` quickstart blocked shipcheck because \`validate-narrative --full-examples\` auto-appends \`--dry-run\`.

## Changes

- \`internal/generator/templates/sync.go.tmpl\`: add \`isDryRunResponse(data)\` helper near \`isEmptyPageResponse\`. In \`syncResource\`, check the helper right after \`c.Get\` and emit a structured \`sync_dryrun\` event + zero-count syncResult on match, before any extract/upsert logic runs.
- Three golden fixtures regenerated for the new short-circuit (\`generate-golden-api\`, \`generate-sync-walker-api\`, \`generate-tier-routing-api\`).
- \`internal/generator/generator_test.go\`: \`TestSyncTemplateShortCircuitsOnDryRunSentinel\` pins (a) the helper definition, (b) the \`sync_dryrun\` event string, and (c) the ordering — \`isDryRunResponse\` check must run before \`upsertSingleObject\` or the bug recurs.

Scope: the dependent-resource sync path (\`syncDependent\`) is already safe — it breaks out cleanly on \`len(items) == 0\` without an upsert, so the sentinel hits no error.

## Test plan

- [x] \`go build ./...\`
- [x] \`go test ./internal/generator/... -run TestSyncTemplateShortCircuitsOnDryRunSentinel\` (pass)
- [x] \`go test ./...\` (full suite, no failures)
- [x] \`go vet ./...\` clean
- [x] \`go fmt ./...\` no changes
- [x] \`scripts/golden.sh verify\` (19 cases pass after intentional update)
- [x] \`golangci-lint run ./internal/generator/...\` (0 issues)

Closes #1382

### DIFF
--- a/internal/generator/templates/sync.go.tmpl
+++ b/internal/generator/templates/sync.go.tmpl
@@ -808,12 +808,24 @@ func extractPageItems(data json.RawMessage, cursorParam string) ([]json.RawMessa
 // client.dryRun returns instead of a real API response. The sync loop
 // uses this to short-circuit before the upsert path, which would
 // otherwise fail with "missing id for <resource>" against a sentinel
-// that has no items and no id.
+// that has no items and no id. The check requires exactly one key
+// (dry_run) so a live API response that happens to include a top-level
+// dry_run field alongside real data isn't misclassified as the
+// sentinel and silently zero out the sync count.
 func isDryRunResponse(data json.RawMessage) bool {
-	var probe struct {
-		DryRun bool `json:"dry_run"`
+	var envelope map[string]json.RawMessage
+	if err := json.Unmarshal(data, &envelope); err != nil {
+		return false
 	}
-	return json.Unmarshal(data, &probe) == nil && probe.DryRun
+	if len(envelope) != 1 {
+		return false
+	}
+	raw, ok := envelope["dry_run"]
+	if !ok {
+		return false
+	}
+	var v bool
+	return json.Unmarshal(raw, &v) == nil && v
 }
 
 func isEmptyPageResponse(data json.RawMessage) bool {
--- a/internal/generator/templates/sync.go.tmpl
+++ b/internal/generator/templates/sync.go.tmpl
@@ -536,6 +536,19 @@ func syncResource(c interface {
 			return syncResult{Resource: resource, Count: totalCount, Err: fmt.Errorf("fetching %s: %w", resource, err), Duration: time.Since(started)}
 		}
 
+		// Dry-run sentinel: client.dryRun returns `{"dry_run": true}` instead
+		// of a real response when --dry-run is set. The upsert path below
+		// would otherwise fail with "missing id for <resource>" because the
+		// sentinel has no items and no id; emit a synthetic success event so
+		// validate-narrative --full-examples (which auto-appends --dry-run)
+		// sees a clean exit.
+		if isDryRunResponse(data) {
+			if !humanFriendly {
+				fmt.Fprintf(os.Stdout, `{"event":"sync_dryrun","resource":"%s"}`+"\n", resource)
+			}
+			return syncResult{Resource: resource, Count: 0, Duration: time.Since(started)}
+		}
+
 		// Try to extract items from the response.
 		// Strategy: try array first, then common wrapper keys.
 		items, nextCursor, hasMore := extractPageItems(data, pageSize.cursorParam)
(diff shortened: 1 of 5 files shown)


internal/generator/templates/sync.go.tmpl
```diff
@@ -528,6 +528,19 @@ func syncResource(c interface {
...
+		// Dry-run sentinel: client.dryRun returns `{"dry_run": true}` instead
+		// of a real response when --dry-run is set. The upsert path below
+		// would otherwise fail with "missing id for <resource>" because the
+		// sentinel has no items and no id; emit a synthetic success event so
+		// validate-narrative --full-examples (which auto-appends --dry-run)
+		// sees a clean exit.
+		if isDryRunResponse(data) {
+			if !humanFriendly {
+				fmt.Fprintf(os.Stdout, `{"event":"sync_dryrun","resource":"%s"}`+"\n", resource)
+			}
```
tmchow, line 548: Leaving the humanFriendly path silent on dry-run for parity with the surrounding sync error/warning branches in syncResource, which also only emit JSON events (the human-friendly path relies on the outer caller's progress output). Adding stderr noise just for dry-run would be inconsistent; if we want richer human-friendly sync output it should be a separate pass across all event paths. Accepting for later.

tmchow:
@mergifyio queue

tmchow:
@mergifyio refresh

tmchow:
@mergifyio queue

PR #1506, opened 2026-05-16
fix(cli): route promoted-command file stem through safeResourceFileStem

### LINKED ISSUE #1489: bug(generator): promoted_test.go is excluded from go build, uncovered case for #1020
## Summary

Closely related to #1020 (closed by #1021). #1020 fixed `safeResourceFileStem()` for the `<resource>_<verb>.go` filename shape (e.g., `webhook_test.go`). However, **promoted commands use a different filename convention** (`promoted_<resource>.go`) and the fix in #1021 does not cover the `promoted_test.go` case.

When an OpenAPI resource is named `test` (e.g., from `/test/generate`), the generator emits `internal/cli/promoted_test.go`. Go's build system excludes any file ending in `_test.go` from the normal build, so the constructor `newTestPromotedCmd` defined in that file is invisible to `root.go`'s `rootCmd.AddCommand(newTestPromotedCmd(flags))`, producing the exact same `undefined: newTestPromotedCmd` failure mode as the original #1020.

## Repro

Observed while generating a CLI for a FastAPI OpenAPI 3.1 spec.

- **Public repro spec:** https://api-staging.quotan.it/openapi.json
- **Generator version:** v4.6.1
- **Path triggering it:** `POST /test/generate` → resource `test` → file `internal/cli/promoted_test.go`

Generator output during `printing-press generate --validate`:

```
FAIL govulncheck ./...
.../internal/cli/root.go:241:21: undefined: newTestPromotedCmd
exit status 1
```

Manual workaround that unblocks the build:

```bash
mv internal/cli/promoted_test.go internal/cli/promoted_test_resource.go
```

## Suggested fix

Apply the same `_test` stem sanitization that #1021 introduced for `<resource>_<verb>.go` to the promoted-command file naming codepath (likely a sibling function or template emitter). Audit the generator for any other emitter that could produce a `<...>_test.go` filename, and centralize the "must not end in _test.go" rule on the final emitted-filename derivation rather than on individual call sites.

## Refs

- #1020 / PR #1021 (the original fix that didn't cover this codepath)

---

Internal downstream report: legacynik/preventivi-ai#211

tmchow:
Closes #1489.

## What

#1021 fixed `safeResourceFileStem` to rename trailing GOOS/GOARCH/`test` tokens to `_cmd`, but only for the `<resource>_<verb>.go` emit path. The promoted-command emitter at [`internal/generator/generator.go:2605`](https://github.com/mvanhorn/cli-printing-press/blob/main/internal/generator/generator.go#L2605) built its path as `"promoted_" + pc.PromotedName + ".go"` without the safe-stem wrapper, so an OpenAPI resource named `test` produces `internal/cli/promoted_test.go`. Go treats `*_test.go` as a test file and excludes it from the normal package build, leaving `root.go`'s `newTestPromotedCmd` reference undefined.

Wrap the stem in `safeResourceFileStem`. The function's existing `_test`/GOOS/GOARCH trailing-token rules now cover both emit paths.

## Repro (from the issue)

OpenAPI spec exposes `POST /test/generate` -> resource `test` -> file `internal/cli/promoted_test.go`:

```
FAIL govulncheck ./...
.../internal/cli/root.go:241:21: undefined: newTestPromotedCmd
```

After fix: emits `internal/cli/promoted_test_cmd.go`; package builds.

## Fix

```diff
- promotedPath := filepath.Join("internal", "cli", "promoted_"+pc.PromotedName+".go")
+ promotedPath := filepath.Join("internal", "cli", safeResourceFileStem("promoted_"+pc.PromotedName)+".go")
```

The function already handles `promoted_test` -> `promoted_test_cmd` (covered by the existing `webhook_test` -> `webhook_test_cmd` case in `safe_filename_test.go`); the bug was a missed call site.

## Tests

- New: `TestGeneratedOutput_PromotedCommand_TestResourceCompiles` builds an APISpec with a single-endpoint resource named `test`, asserts emission as `promoted_test_cmd.go` (not `promoted_test.go`), and runs `go build ./...` to catch the silent-exclusion regression.
- Pre-existing `TestSafeResourceFileStem` already covers the function-level behavior.

## Test plan

- [x] `go test ./internal/generator/ -run TestGeneratedOutput_PromotedCommand_TestResourceCompiles -v` passes
- [x] `go test ./internal/generator/...` clean (923/923)
- [x] `go vet ./...` / `go fmt ./...` clean
- [x] `golangci-lint run ./...` no issues
- [x] `scripts/golden.sh verify` passes (20/20)

### DIFF
--- a/internal/generator/generator.go
+++ b/internal/generator/generator.go
@@ -2602,7 +2602,7 @@ func (g *Generator) renderPromotedCommandFiles(promotedCommands []PromotedComman
 			IsReadOnly:      endpointIsReadCommand(pc.Endpoint, pc.EndpointName),
 			APISpec:         g.Spec,
 		}
-		promotedPath := filepath.Join("internal", "cli", "promoted_"+pc.PromotedName+".go")
+		promotedPath := filepath.Join("internal", "cli", safeResourceFileStem("promoted_"+pc.PromotedName)+".go")
 		if err := g.renderTemplate("command_promoted.go.tmpl", promotedPath, promotedData); err != nil {
 			return fmt.Errorf("rendering promoted command %s: %w", pc.PromotedName, err)
 		}
--- a/internal/generator/generator_test.go
+++ b/internal/generator/generator_test.go
@@ -5509,6 +5509,41 @@ func TestGeneratedOutput_PromotedCommandCompiles(t *testing.T) {
 	runGoCommand(t, outputDir, "build", "./...")
 }
 
+// A resource named `test` produces `promoted_test.go`; Go treats *_test.go as a
+// test file and excludes it from the normal package build, so root.go's
+// reference to newTestPromotedCmd fails to compile. Mirrors the safe-stem fix
+// that #1021 applied to `<resource>_<verb>.go`.
+func TestGeneratedOutput_PromotedCommand_TestResourceCompiles(t *testing.T) {
+	t.Parallel()
+
+	apiSpec := &spec.APISpec{
+		Name:    "promotedtest",
+		Version: "0.1.0",
+		BaseURL: "https://api.example.com",
+		Auth:    spec.AuthConfig{Type: "api_key", Header: "X-Api-Key", EnvVars: []string{"PT_API_KEY"}},
+		Config:  spec.ConfigSpec{Format: "toml", Path: "~/.config/promotedtest-pp-cli/config.toml"},
+		Resources: map[string]spec.Resource{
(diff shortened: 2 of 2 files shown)


tmchow:
@mergifyio queue

CI rerun (attempt 2) was green; the prior `test (pipeline)` failure was a Linux `text file busy` race in `dogfood_test.go:2031` unrelated to this one-line generator fix.

Issue #981, opened 2026-05-10
[P3] WU-1: cookieAuth in: cookie securityScheme wired as request header instead of HTTP cookie


richtheworld:
## Goal

When an OpenAPI spec declares `securitySchemes[*]: {type: apiKey, in: cookie, name: <N>}`, generated clients must send the value in the HTTP `Cookie` header, not as a request header named `<N>`.

## What happened

Generated a CLI from a synthesized OpenAPI 3.0 spec for Paraform's private tRPC API. The spec declared:

```yaml
components:
  securitySchemes:
    cookieAuth:
      type: apiKey
      in: cookie
      name: __Secure-next-auth.session-token
```

The generated `internal/client/client.go` emitted:

```go
if authHeader != "" {
    req.Header.Set("__Secure-next-auth.session-token", authHeader)
}
```

This sets `__Secure-next-auth.session-token` as a request header. The server (NextAuth) requires the value in the `Cookie:` header. Real GET returned HTTP 401 UNAUTHORIZED.

Patched locally to:

```go
if authHeader != "" {
    if strings.Contains(authHeader, "=") {
        req.Header.Set("Cookie", authHeader)
    } else {
        req.AddCookie(&http.Cookie{Name: "__Secure-next-auth.session-token", Value: authHeader})
    }
}
```

Real GET returned HTTP 400 after patch (auth past, tRPC validation error on a separate concern — see related issue link).

## Where to fix

`internal/generator/templates/client/*` (or equivalent template that emits the auth-injection block). Branch on `securitySchemes[*].in`:

- `in: header` → existing `req.Header.Set(<name>, <val>)` path
- `in: cookie` → `req.AddCookie(&http.Cookie{Name: <name>, Value: <val>})` path
- `in: query` → URL query param injection path

The fix is gated on the `in` field; APIs using header or query auth are unaffected.

## Doctor caveat

The defect is silent — `paraform-trpc-pp-cli doctor` reports `OK Auth: configured` because the env var was present. Doctor doesn't validate that the request actually carries the cookie. Worth a follow-up: doctor's auth check could perform one read call with the auth applied and confirm 2xx/4xx-other-than-401.

## Acceptance criteria

- **Positive:** generate a CLI from a spec with `securitySchemes.x.in: cookie`; capture HTTP request via mitmproxy/httpbin and confirm the `Cookie:` header carries the value.
[...]

PR #1518, opened 2026-05-16
fix(cli): filter auth login --chrome cookies by spec auth.cookies allowlist

### LINKED ISSUE #1292: generator: auth login --chrome saves unfiltered cookie blob, ignores spec auth.cookies allowlist
## Symptom

`<cli> auth login --chrome` (generated by the printing-press generator from any spec with `auth.type: cookie`) saves **every** cookie the target domain has set, concatenated as `name1=value1; name2=value2; name3=value3` into the config.toml. The spec's `auth.cookies: [<name>]` allowlist is ignored — both for *which* cookie to keep AND for *which TOML field* to store it under.

Combined with #981 (cookieAuth sent as Authorization header), this means cookie-auth CLIs ship runtime-broken on first call to any auth-required endpoint, even when the user successfully runs `auth login --chrome` and `auth status` reports `Authenticated`.

## Evidence (Skool, run 20260513-090632)

Spec declared:
```yaml
auth:
  type: cookie
  cookie_domain: skool.com
  cookies:
    - auth_token
```

After `skool-pp-cli auth login --chrome`:
```toml
access_token = 'client_id=23f48d8adea74cb3b29e2ea7c711bfb1; auth_token=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9....; aws-waf-token=66a461a4-386a-...'
auth_token = ''   # ← the field the runtime READS via SkoolAuthToken
```

Two compounding bugs:
1. **All 3 skool.com cookies stored** (`client_id`, `auth_token`, `aws-waf-token`), not just the one in the spec's `auth.cookies` list (`auth_token`).
2. **Stored in `access_token` field, runtime reads `auth_token` field** — a generator field-name mismatch. Even if I manually moved the JWT to `auth_token`, the request still failed because #981 sends Authorization header.

Server-side error when the concat blob reaches api2.skool.com (regardless of which header): `parse jwt: invalid token`.

## Cross-API impact

Any cookie-auth CLI whose target domain sets ≥2 cookies. In 2026 that's the **majority** of cookie-auth sites (CSRF, WAF tokens, session ID, login state, anti-bot fingerprints). Candidates:
- **Skool** — confirmed broken in this session (3 cookies stored, 2 are noise)
- **Substack** (see #1102) — composed cookie auth where `substack.lli` matters but other cookies don't
- **Discord** (user-token auth), Reddit, LinkedIn admin endpoints — all set anti-bot/WAF cookies alongside the auth credential

## Proposed fix (Component: generator)

Two changes in the generator's `auth.go` template + `config.go` template:

### 1. Filter extractCookies output by spec.auth.cookies

```go
// auth.go template (generated) — pseudocode
cookies, err := extractCookies(tool, domain, profileDir)
if err != nil { return err }

// NEW: filter by spec's auth.cookies allowlist
allowed := map[string]bool{}
for _, name := range cfg.AuthCookieNames() {  // exposes spec.auth.cookies
    allowed[name] = true
}
kept := []Cookie{}
for _, c := range cookies {
    if allowed[c.Name] {
        kept = append(kept, c)
    }
}
if len(kept) == 0 {
    return fmt.Errorf("no auth_token cookie found for %s; log in first", domain)
}
```

### 2. Store cookies in the field the runtime reads

The generator currently emits TWO fields in config.go (`access_token` and `auth_token`/`SkoolAuthToken`) that conflict. For `auth.type: cookie`, only the spec-declared cookie name should be a config field, and the writer + reader must agree on it.

When `len(cfg.AuthCookieNames()) == 1`: write the value into a TOML field named after the cookie. When > 1 (composed auth): write as a structured map and let the runtime compose the header per the spec's format string.

## Acceptance tests

**Positive 1 (single cookie):**
```yaml
auth: { type: cookie, cookies: [session_id] }
```
Chrome cookie store has `session_id=X; csrf=Y; aws-waf=Z` for domain.
→ `auth login --chrome` saves only `session_id = 'X'` (or `auth.session_id = 'X'`).
→ `auth status` reports `Authenticated`.
→ A real auth-required request reaches the server with `Cookie: session_id=X` (requires #981 also fixed for end-to-end).

**Positive 2 (composed):**
```yaml
auth: { type: composed, format: "Auth {a}:{b}", cookies: [a, b] }
```
Chrome has `a=1; b=2; junk=3` → saves both `a` and `b`, drops `junk`. Runtime composes `Authorization: Auth 1:2`.

**Negative (api_key auth):**
```yaml
auth: { type: api_key, ... }
```
→ `auth login --chrome` either no-ops with a clear message ("this CLI uses api_key auth; set $<API>_API_KEY instead") or is hidden from `auth --help`.

## Related issues

- **#981** [P3] WU-1: cookieAuth in: cookie securityScheme wired as request header — `aligned`, this WU is part 2 of the same cookie-auth gap. #981 fixes the **runtime** header injection; this WU fixes the **storage** layer that feeds it. Both must land for cookie-auth CLIs to work end-to-end.
- **#1101** Windows: auth login --chrome fails (python3 hardcode + pycookiecheat no Windows) — `extends`, same `auth login` command, different platform-specific bug. Not a duplicate.
- **#1102** substack catalog: composed cookie auth needs substack.lli — `aligned` direction (filter by name), but specific to one catalog entry.
- **#1158** Generated config.go AuthHeader() unconditionally overwrites AuthSource — `extends`, same auth code path region.

## Retro context

Skool retro (run 20260513-090632, ai-driven-dev fork) surfaced this on first live dogfood call. The local fix to make Skool work was a 1-line patch to client.go (#981's territory) PLUS manually moving JWT from `access_token` to `auth_token` field in config.toml (this WU's territory). Both changes are needed. The generator should emit a cookie-auth CLI that doesn't require either patch.

Retro doc local path: `~/printing-press/manuscripts/skool/20260513-090632/proofs/20260513-110047-retro-skool-pp-cli.md`

tmchow:
## Intent

Issue: Closes #1292

When a spec declares `auth.cookies: [name]`, `auth login --chrome` was saving every cookie the target domain had set into `access_token`. The runtime then sent the whole blob on every authenticated request (CSRF, WAF, anti-bot fingerprints alongside the real credential), routinely tripping upstream WAFs and shadowing the credential in the Cookie header.

## Approach

Apply the allowlist filter the composed-auth branch already uses to the non-composed cookie-auth branch in `auth_browser.go.tmpl`. After `extractCookies` returns the raw blob: parse, keep only the names declared in `spec.auth.cookies`, and rejoin before `SaveTokens`. A missing required cookie aborts login with a re-login prompt instead of silently saving a bad blob.

Specs that declare no allowlist (legacy cookie-auth CLIs) keep their current behavior unchanged; the filter block is gated on `{{- if .Auth.Cookies}}`.

## Scope

Primary area: generator (`internal/generator/templates/auth_browser.go.tmpl`).

Why this belongs in this repo: machine fix. The bug is in the template emitted into every cookie-auth printed CLI; fixing it here propagates the fix to every future print and to every regen of an existing CLI.

The runtime header-vs-cookie placement (currently sends as `Authorization` instead of `Cookie`) is tracked separately by #981 and not changed here.

## Risk

- Legacy specs (`Auth.Cookies: nil`) hit the existing code path; generated output is byte-for-byte unchanged. Confirmed by the second subtest of `TestGenerate_CookieAuthFiltersAllowlistOnLogin` and by the golden harness (no fixture exercises a non-empty `Auth.Cookies` cookie-auth CLI).
- Composed-auth branch is untouched.
- A user whose Chrome session is missing the required cookie now hits a clear error at login time instead of authenticating with a corrupted blob and 4xx-ing later.

## Output Contract

For cookie-auth CLIs with `auth.cookies` declared, the generated `internal/cli/auth.go` now emits a `requiredCookies := []string{...}` declaration plus a parse-filter-rejoin loop ahead of `SaveTokens`. Generated output for cookie-auth CLIs with empty `Auth.Cookies` is unchanged.

## Verification

- `go test ./...` — 4388 passed in 35 packages
- `scripts/golden.sh verify` — 20/20 passed
- `go build -o ./printing-press ./cmd/printing-press` — clean
- New test `TestGenerate_CookieAuthFiltersAllowlistOnLogin` covers both the allowlisted and legacy branches; the allowlisted subtest also runs `go build ./...` against the generated CLI to catch template compile errors.

## AI / Automation Disclosure

### DIFF
--- a/internal/generator/generator_test.go
+++ b/internal/generator/generator_test.go
@@ -6583,6 +6583,52 @@ func TestGenerate_CookieAuthFiltersAllowlistOnLogin(t *testing.T) {
 		assert.NotContains(t, content, "requiredCookies := []string{")
 		assert.NotContains(t, content, `cookies = strings.Join(kept, "; ")`)
 	})
+
+	t.Run("refresh path also filters by allowlist", func(t *testing.T) {
+		t.Parallel()
+		apiSpec := &spec.APISpec{
+			Name:    "refreshfilterapp",
+			Version: "0.1.0",
+			BaseURL: "https://app.example.com",
+			Auth: spec.AuthConfig{
+				Type:                           "cookie",
+				Header:                         "Cookie",
+				In:                             "cookie",
+				CookieDomain:                   ".example.com",
+				Cookies:                        []string{"session_id"},
+				EnvVars:                        []string{"REFRESHFILTERAPP_COOKIES"},
+				RequiresBrowserSession:         true,
+				BrowserSessionValidationPath:   "/api/items",
+				BrowserSessionValidationMethod: "GET",
+			},
+			Config: spec.ConfigSpec{Format: "toml"},
+			Resources: map[string]spec.Resource{
+				"items": {
+					Description: "Manage items",
+					Endpoints: map[string]spec.Endpoint{
+						"list": {Method: "GET", Path: "/api/items", Description: "List items"},
+					},
+				},
+			},
+		}
+
+		outputDir := filepath.Join(t.TempDir(), "refreshfilterapp-pp-cli")
+		require.NoError(t, New(apiSpec, outputDir).Generate())
+
+		auth, err := os.ReadFile(filepath.Join(outputDir, "internal", "cli", "auth.go"))
+		require.NoError(t, err)
+		content := string(auth)
+
+		require.Contains(t, content, "func refreshStoredBrowserCookies")
+		_, refresh, _ := strings.Cut(content, "func refreshStoredBrowserCookies")
+		// Without the fix, the refresh body has SaveTokens(cookies) with no
+		// prior filter — the second SaveTokens call would then sit before any
+		// allowlist parsing in the function.
+		assert.Contains(t, refresh, `requiredCookies := []string{"session_id"}`)
+		assert.Contains(t, refresh, `cookies = strings.Join(kept, "; ")`)
+
+		runGoCommand(t, outputDir, "build", "./...")
+	})
 }
 
 func TestGenerate_UserAgentOverrideGatedByBrowserTransport(t *testing.T) {
--- a/internal/generator/generator_test.go
+++ b/internal/generator/generator_test.go
@@ -6502,6 +6502,89 @@ func TestGenerate_CookieAuthWindowsCompatibility(t *testing.T) {
 	assert.Contains(t, content, "cookie-scoop-cli")
 }
 
+// TestGenerate_CookieAuthFiltersAllowlistOnLogin pins that `auth login --chrome`
+// stores only the cookies named in spec.auth.cookies when an allowlist is
+// declared, and falls back to the unfiltered blob when the allowlist is
+// empty (legacy specs).
+func TestGenerate_CookieAuthFiltersAllowlistOnLogin(t *testing.T) {
+	t.Parallel()
+
+	t.Run("allowlist declared filters extracted cookies", func(t *testing.T) {
+		t.Parallel()
+		apiSpec := &spec.APISpec{
+			Name:    "filterapp",
+			Version: "0.1.0",
+			BaseURL: "https://app.example.com",
+			Auth: spec.AuthConfig{
+				Type:         "cookie",
+				Header:       "Cookie",
+				In:           "cookie",
+				CookieDomain: ".example.com",
+				Cookies:      []string{"session_id"},
+				EnvVars:      []string{"FILTERAPP_COOKIES"},
+			},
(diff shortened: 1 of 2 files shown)


tmchow:
Addressed Greptile's P1 finding on `refreshStoredBrowserCookies` in cd47dabe: the non-composed cookie-auth branch now applies the same allowlist filter as login. Without this, a refresh on a `cookie + allowlist + requires_browser_session` CLI would overwrite the filtered token with the raw domain blob on the next session refresh. Added `refresh path also filters by allowlist` subtest covering the new path.

PR #1519, opened 2026-05-16
fix(cli): detect integer page paginator and advance numerically in sync

### LINKED ISSUE #1296: Sync template defaults to body-cursor extraction (`cursorParam: "after"`) even when spec declares integer `page` pagination
> Filed by `/printing-press-retro` from a run of **Freshservice** (scorecard 84/100, 100% verify pass).

## Summary

The generator's sync command emits a body-cursor extraction strategy by default — it tries to find a `next_cursor`-shaped field in the response body and uses an `?after=<cursor>` query param to advance. For APIs that paginate by integer `?page=N` and emit no body cursor, this means every sync stops after page 1 (max 100 records), silently truncating data. Freshservice's spec declared `page: {type: integer}` on every list endpoint and got the cursor-based treatment regardless. The fallback in `sync.go.tmpl` (`cursorParam: "after"`) is wrong for ~30-40% of REST APIs.

## Where to look

- **Component:** `comp:spec-parser` (the profiler should detect page-int / Link-header / body-cursor shapes from the spec); the template fix in `internal/generator/templates/sync.go.tmpl` is downstream.
- **Likely area:**
  - `internal/generator/templates/sync.go.tmpl` line ~661 (the `determinePaginationDefaults` template) — current fallback is `cursorParam: "after"`, hardcoded.
  - `internal/spec/` or `internal/openapi/` — wherever `Pagination.CursorParam` gets populated. The profiler likely greps for cursor-shaped field names and doesn't recognize `page: integer` as a distinct paginator that needs a different iteration strategy.
- **Triggered when:** spec declares `page: {in: query, type: integer}` (or similar) on list endpoints and the response has no body cursor field.

## What we observed

Freshservice's spec has `page` as an integer query param on every list endpoint. The generated `sync.go` chose `cursorParam: "after"` and called `extractPaginationFromEnvelope()` on every response. Freshservice's list response is just `{"tickets": [...]}` (no `next_cursor`, no `links.next` body field), so `hasMore` returned `false` after page 1 and the loop terminated. Result: `sync` reported `total: 100` for a tenant with 10,000+ tickets.

The template (`sync.go.tmpl:661`):

```go-template
func determinePaginationDefaults() paginationDefaults {
    return paginationDefaults{
        cursorParam: "{{if .Pagination.CursorParam}}{{.Pagination.CursorParam}}{{else}}after{{end}}",
        limitParam:  "{{if .Pagination.PageSizeParam}}{{.Pagination.PageSizeParam}}{{else}}limit{{end}}",
        limit:       {{if .Pagination.DefaultPageSize}}{{.Pagination.DefaultPageSize}}{{else}}100{{end}},
    }
}
```

`.Pagination.CursorParam` was never populated for Freshservice; the template fell back to `"after"`. Hand-patched sync loop:

```go
if pageSize.cursorParam == "page" && nextCursor == "" && len(items) >= pageSize.limit {
    currentPage, _ := strconv.Atoi(cursor)
    if currentPage < 1 { currentPage = 1 }
    nextCursor = strconv.Itoa(currentPage + 1)
    hasMore = true
}
```

After patch: `sync --max-pages 0` walks page 1 → 2 → 3 → … through 10,000 records, cap hit, clean terminate.

## Suspected root cause

The profiler step that populates `.Pagination.\*` (uncertain which file) doesn't have a branch for the page-int paginator shape. It likely looks for cursor-shaped field names in the response schema (`next`, `cursor`, `after`) and falls through to the `"after"` default when none are found — but the signal it should also look at is whether the *request* declares a `page: integer` query param.

I haven't traced exactly where this lives. Grep for `Pagination.CursorParam` assignments / `paginationDefaults` struct construction to find the profiler.

## Suggested direction

Profile the spec for the three common pagination shapes and emit different iteration loops:

1. **Body cursor (current default).** Spec response declares `next_cursor` / `paging.next.after` / `links.next` body field. Template emits the current envelope-extraction loop.
2. **Page integer.** Spec request declares `page: {type: integer}` query param AND response has no body cursor field. Template emits a numeric counter that advances `?page=<N+1>` until the response returns fewer items than `per_page`.
3. **Link header.** Spec `info.description` or response headers mention RFC 5988, OR a profiler heuristic flags it. Template emits a Link-header parser; falls back to page-int when Link is absent on the first response.

The current loop structure in `sync.go.tmpl` is mostly reusable — the difference between strategies is the "compute next cursor" step. Three small template branches.

## Acceptance criteria

- **positive (page-int):** Generate from a spec with `page: integer` request param and no body cursor → `sync` emits `params["page"] = strconv.Itoa(currentPage + 1)` and terminates when `len(items) < per_page`.
- **positive (link header):** Generate from a spec marked Link-header-paginated → `sync` parses the `Link: <…>; rel="next"` header.
- **negative (regression):** Generate from a spec with `next_cursor` in response → body-cursor extraction is unchanged.
- **regression suite:** Run against existing cursor-based CLIs (e.g. anything using Slack-shaped `?cursor=`) to confirm no behavior change.

## Frequency

~30-40% of REST APIs. Concrete impacted families: GitHub (Link header + page-int), Atlassian Cloud (offset-based, related shape), HubSpot (mixed offset+cursor), all of Freshworks (Freshservice, Freshdesk, Freshchat).

## Complexity

medium

## Scope boundary

Out of scope: GraphQL cursor pagination (different shape entirely), client-side post-filter pagination (return-all + filter), or per-endpoint pagination override (some APIs paginate one endpoint with cursors and another with pages — that's a follow-on after the basic case lands).

## Dependencies

None.

## Related issues

None.

## Artifacts

| Artifact | Link |
|----------|------|
| Retro document | [view](https://litter.catbox.moe/h39hm0.md) |
| Manuscripts (research + proofs) | [zip](https://litter.catbox.moe/3oz5mr.zip) |
| Generated CLI source | [zip](https://litter.catbox.moe/riy166.zip) |

> Note: artifacts hosted on litterbox.catbox.moe with **72h retention** (catbox.moe rejected the IP). Permanent local copies are at `~/printing-press/manuscripts/freshservice/20260512-170855/proofs/`.

---

tmchow:
## Intent

`sync` against APIs that paginate by integer `?page=N` and emit no body cursor (Freshworks family, Atlassian Cloud, HubSpot, large swath of REST APIs) silently truncates after page 1: `extractPageItems` finds no cursor in the response, returns `("", false)`, and the loop breaks. Result: `sync_summary` reports `total: 100` for a tenant with 10,000+ records.

Issue: Closes #1296

## Approach

Two-part change scoped to the positive page-int case from the issue's acceptance criteria. The negative cursor-based regression is covered; Link-header pagination (a different code path in HTTP header land) stays a follow-up.

1. **OpenAPI parser** (`detectPagination`) learns to recognize `page`/`pageNumber`/`page_number`/`page[number]` as a `Type="page"` paginator. Placed AFTER the existing `pagetoken`/`cursor`/`offset` branches so mixed-form specs (`cursor` + `page`, or `offset` + `page`) preserve their existing cursor semantics.
2. **Sync template** adds a runtime fallback: when `cursorParam == "page"` and `extractPageItems` returns an empty cursor but the page is full, atoi the current cursor (defaulting to `1`) and advance by `+1`. Page 1 sends no `?page=` so the API serves its default; subsequent pages send `?page=2`, `?page=3`, etc. Mirrored into the dependent-resource sync loop for parity.
3. `spec.Pagination` doc comment updated to include `page` in the `Type` and `CursorParam` enumerations.

## Scope

Primary area: `comp:generator` (template + parser).

Why this belongs in this repo: every CLI generated against a `page: integer` API regenerates with the fix on the next print. The runtime branch is opt-in by `cursorParam == "page"` so existing cursor-based goldens are byte-stable.

## Risk

**Mixed-form regression.** APIs that declare both `cursor` and `page` (rare but real) keep cursor semantics because the new branch runs only after `pag.Type == ""`. Covered by `TestDetectPaginationCursorBeatsPage` / `TestDetectPaginationOffsetBeatsPage`.

**Runaway page increment.** If an API returns exactly `limit` items on the last page (no body cursor and no `has_more`), the loop fetches one extra page that returns 0 items, then breaks via `len(items) < pageSize.limit`. One wasted request, matches the user's hand-patch shape from the issue, and stays inside the `--max-pages` cap.

**Sticky-cursor detector.** Page-int generates strictly-increasing cursors (`"2"`, `"3"`, …) so `lastNextCursor` comparison cannot false-trip.

## Output Contract

3 golden fixtures (`generate-golden-api`, `generate-sync-walker-api`, `generate-tier-routing-api`) regenerated — every printed CLI's `sync.go` now carries the page-int fallback block. Diff verified identical across all three.

## Verification

- `go build ./...` — clean
- `go vet ./...` — clean
- `go test ./...` — 4410 passed, 35 packages
- `scripts/golden.sh verify` — 20 cases pass after `update`

## AI / Automation Disclosure

### DIFF
--- a/internal/generator/generator_test.go
+++ b/internal/generator/generator_test.go
@@ -2851,6 +2851,66 @@ func TestExtractPageItemsNoCursor(t *testing.T) {
 	runGoCommandRequired(t, outputDir, "test", "-run", "TestExtractPageItems", "./internal/cli")
 }
 
+// TestSyncPageIntPaginationAdvancesAfterFullPage guards #1296: APIs that
+// paginate by integer ?page=N (Freshworks, HubSpot, Atlassian, …) emit
+// no body cursor, so extractPageItems returns ("", false). The runtime
+// must detect cursorParam == "page" and advance the integer counter
+// when the page is full, otherwise sync stops after page 1.
+func TestSyncPageIntPaginationAdvancesAfterFullPage(t *testing.T) {
+	t.Parallel()
+
+	apiSpec := &spec.APISpec{
+		Name:    "freshy",
+		Version: "0.1.0",
+		BaseURL: "https://freshy.example.com",
+		Auth:    spec.AuthConfig{Type: "none"},
+		Config: spec.ConfigSpec{
+			Format: "toml",
+			Path:   "~/.config/freshy-pp-cli/config.toml",
+		},
+		Resources: map[string]spec.Resource{
+			"tickets": {
+				Description: "Tickets",
+				Endpoints: map[string]spec.Endpoint{
+					"list": {
+						Method:      "GET",
+						Path:        "/tickets",
+						Description: "List tickets",
+						Response:    spec.ResponseDef{Type: "array"},
+						Params: []spec.Param{
+							{Name: "page", Type: "integer"},
+							{Name: "per_page", Type: "integer"},
+						},
+						Pagination: &spec.Pagination{
+							Type:        "page",
+							CursorParam: "page",
+							LimitParam:  "per_page",
+						},
+					},
+				},
+			},
+		},
+	}
+
+	outputDir := filepath.Join(t.TempDir(), naming.CLI(apiSpec.Name))
+	gen := New(apiSpec, outputDir)
+	gen.VisionSet = VisionTemplateSet{Store: true, Sync: true}
+	require.NoError(t, gen.Generate())
+
+	syncSrc, err := os.ReadFile(filepath.Join(outputDir, "internal", "cli", "sync.go"))
+	require.NoError(t, err)
+	src := string(syncSrc)
+
+	// Tier 1: emission canary. The page-int fallback branch must ship so
+	// page-int APIs advance numerically when no body cursor is extracted.
+	assert.Contains(t, src, `pageSize.cursorParam == "page"`,
+		"sync.go must emit the page-int paginator fallback branch")
+	assert.Contains(t, src, `strconv.Itoa(currentPage + 1)`,
+		"page-int fallback must increment the integer cursor")
+	assert.Contains(t, src, `cursorParam: "page"`,
+		"determinePaginationDefaults must emit cursorParam \"page\" when the profile selects it")
+}
+
 // TestGeneratedSyncHandlesPascalCaseDotNetShape verifies the generated sync +
 // store paths recognize .NET-shape PascalCase envelopes ("Items"), PKs ("Id"),
 // and field keys (LookupFieldValue PascalCase pass). Parser-side tier-5 PK
--- a/internal/generator/generator_test.go
+++ b/internal/generator/generator_test.go
@@ -2854,61 +2854,84 @@ func TestExtractPageItemsNoCursor(t *testing.T) {
 // TestSyncPageIntPaginationAdvancesAfterFullPage guards #1296: APIs that
 // paginate by integer ?page=N (Freshworks, HubSpot, Atlassian, …) emit
 // no body cursor, so extractPageItems returns ("", false). The runtime
-// must detect cursorParam == "page" and advance the integer counter
-// when the page is full, otherwise sync stops after page 1.
+// must detect the paginator type (not the raw param name) and advance
+// the integer counter when the page is full, otherwise sync stops after
+// page 1. Parametrized over every canonical page-int spelling so all
+// four variants share the same guard regardless of casing.
 func TestSyncPageIntPaginationAdvancesAfterFullPage(t *testing.T) {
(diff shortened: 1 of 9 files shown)


tmchow:
Greptile P1 addressed in 883005ba: plumbed `PaginationProfile.CursorType` (aggregated from `endpoint.Pagination.Type`) through `paginationDefaults.cursorType` and switched both fallback guards to compare on the type, not the param name. The generator test is now parametrized over all four canonical spellings (`page`, `page_number`, `pageNumber`, `page[number]`) so each one independently asserts the guard fires. 4414 tests pass, golden verify clean.

Issue #1520, opened 2026-05-16
Promoted POST/PUT/PATCH commands lack --stdin fallback when body exceeds depth cap


tmchow:
## Summary

Follow-up to #1240. After #1240's depth cap on per-field body flags, promoted POST/PUT/PATCH commands (emitted via `internal/generator/templates/command_promoted.go.tmpl`) can silently drop required fields when the request body recurses past `maxBodyFlagDepth`.

## Where it surfaces

`command_promoted.go.tmpl` uses the same depth-capped helpers as `command_endpoint.go.tmpl`:
- `bodyVarDecls` (line 29)
- `bodyRequiredChecks` (line 48)
- `bodyMapForEndpoint` (line 211)
- `bodyFlagRegs` (line 341)

But unlike `command_endpoint.go.tmpl`, the promoted template emits no `--stdin` flag and no `if !stdinBody { ... }` wrapper around the required-checks block. Result: when a promoted endpoint's body exceeds the cap, the deeper subtree is silently absent from the CLI surface and the API receives an incomplete body. The user has no flag-level signal that fields are missing.

## Why this is reachable now

Pre-#1240, a deeply nested promoted endpoint produced a 40k-line file that OOM-killed the Go compiler before this gap could surface. With the cap in place those endpoints now build, exposing the surface.

## Suggested direction

Two options, in order of scope:

1. **Skip promotion when the body exceeds the cap.** At promotion-decision time, call `bodyExceedsFlagDepth(endpoint)` and refuse to promote endpoints whose body would truncate. The canonical (non-promoted) command already has `--stdin` and remains reachable. Smaller scope; preserves the cap's invariants without expanding the promoted-command template.

2. **Emit `--stdin` scaffolding in the promoted template.** Mirror `command_endpoint.go.tmpl`'s pattern: declare `stdinBody`, wrap required checks in `if !stdinBody`, gate the `--stdin` help text via `bodyExceedsFlagDepth`. Broader scope; preserves promotion for these endpoints.

## Acceptance criteria

- positive: a spec with a deeply nested promoted POST/PUT command does not silently drop required-field validation; either the command is not promoted, or the command exposes `--stdin` with parity to its canonical sibling.
- negative: shallow promoted commands emit identically to today (no regression).

## Frequency

[...]

PR #1522, opened 2026-05-16
fix(cli): cap body-flag recursion depth to prevent compiler OOM

### LINKED ISSUE #1240: POST/PUT command files balloon to >40k lines on deeply nested types; Go compile OOMs
> Filed by `/printing-press-retro` from a run of **Tripletex** (Norwegian SMB accounting/ERP) — scorecard 79/100, 100% verify pass.

## Summary

Generator inlines the entire object hierarchy as Cobra flag declarations in each POST/PUT command file. For commands on deeply nested request body types, a single `.go` file balloons to 15-46 thousand lines, and the cumulative `internal/cli` package exceeds 50 MB / 555k lines. `go build` OOM-kills the Go compiler (`signal: killed`) before it can finish the package.

## Where to look

- **Component:** generator
- **Likely area:** The body-flag emitter inside the POST/PUT command template generator. Recursive expansion has no depth cap; deep types fan out to thousands of flag declarations per command file.
- **Triggered when:** Spec has a POST/PUT body type that recurses through several layers (TravelExpense → Project → Customer → PostalAddress → Country → Region → Customer → PostalAddress → ...).

## What we observed

After successful generation + dedupe (see related issue for the dedupe pattern), the 18 worst-offender files in this run:

```
46011 internal/cli/timesheet_time-clock-put.go
40357 internal/cli/project_post.go
40356 internal/cli/project_put.go
35564 internal/cli/project_orderline-post.go
35563 internal/cli/project_orderline-put.go
25717 internal/cli/timesheet_salary-project-type-specification-post.go
25716 internal/cli/timesheet_salary-project-type-specification-put.go
25687 internal/cli/timesheet_entry-post.go
25686 internal/cli/timesheet_entry-put.go
20979 internal/cli/timesheet_allocated-post.go
...
```

Total `internal/cli` size: 50 MB, 554,604 lines across 276 files.

`go build ./...` exit:

```
tripletex-pp-cli/internal/cli: /opt/homebrew/Cellar/go/1.26.3/libexec/pkg/tool/darwin_arm64/compile: signal: killed
```

Compile process peaked at 2.5 GB RSS before being killed by the OS. Workaround: delete the 18 worst-offender files (drops the package to 68k lines, allows compilation). Deleting them loses CREATE/UPDATE capability for the affected resources — a significant feature loss.

## Suspected root cause

POST/PUT body-flag emitter recursively expands all reachable fields in the request body type. For deeply nested types, this produces tens of thousands of flag declarations per command. The Go compiler can't handle a single function with 46k local variables + flag bindings. Confidence: high — the file size is direct evidence; the emitter has no depth-cap mechanism visible in the output.

## Suggested direction

Cap recursion depth on body-field expansion. Default 3-4 levels. Beyond that, emit a single `--body-stdin` (read JSON body from stdin) flag for the command so users / agents can still call it. Optionally emit a narrowly-scoped `--<resource>-id` shortcut form when the spec exposes a relation.

```go
// Pseudocode for the emitter:
if depth >= maxBodyDepth {
  emitBodyStdinFlag(cmd)
  return
}
for _, field := range bodyType.Fields {
  emitFlag(cmd, field)
  if isObject(field.Type) {
    expandRecursively(cmd, field.Type, depth+1)
  }
}
```

Default `maxBodyDepth = 3` is a reasonable starting point. Shallow APIs (Stripe, Linear) emit identically to today; deep APIs (Tripletex, NetSuite, SAP) get a working `--body-stdin` fallback at the boundary.

## Acceptance criteria

- positive: generate from a spec with a 6-level-deep request body type; the resulting command file is < 5k lines; `go build ./...` succeeds in normal time.
- positive: when depth-cap fires, the command exposes `--body-stdin` with help text explaining the fallback.
- negative: a 1-2 level-deep type emits the same flag set as today (no regression for shallow APIs).
- negative: a CLI generated before the change still builds (no breaking change to existing CLIs).

## Frequency

Every enterprise/ERP API with deep write-side schemas. Estimate: 10-20% of catalog where this manifests as a compile-blocker; more often as "this CLI is unwieldy" (passes build but ships massive files that take 30+ seconds to compile per command).

Three concrete APIs with evidence:

- **Tripletex** — direct evidence: 18 files >15k lines, max 46k.
- **NetSuite SuiteCloud REST** — invoice and journal-entry types nest line-items → accounts → customer → project → department → currency → back-references. Public spec excerpts show 8+ levels of nesting on `customrecord_invoice`.
- **SAP S/4HANA OData** — `SalesOrder` → `Item` → `Product` → `PricingComponents` → `SalesOrderItem` → ... Same recursive-self pattern.

## Complexity

medium

## Scope boundary

Does NOT change response-type rendering (only request body flags). Does NOT remove typed body when shallow.

## Dependencies

None.

## Related issues

None from prior retros (scanned for "compile", "massive", "nested", "body", "recursion", "depth" — closest match #1051 about novel-feature aux tables, but unrelated).

## Artifacts

| Artifact | Link |
|----------|------|
| Retro document | https://files.catbox.moe/lhhuc3.md |
| Manuscripts | Upload failed — local copy at `manuscripts/tripletex/20260512-143756/proofs/tripletex-manuscripts.zip` |
| Generated CLI source code | https://files.catbox.moe/1wcbtu.zip |

---

tmchow:
## Summary

Closes #1240. Caps recursion in the four body-flag emitters at `maxBodyFlagDepth = 3` so deeply nested OpenAPI request bodies stop producing 40k+ line POST/PUT files that OOM the Go compiler. Truncated subtrees stay reachable via the existing `--stdin` flag; the flag's help text is rewritten when truncation fires.

## What changed

- `internal/generator/generator.go` — new `maxBodyFlagDepth` constant, `depth int` parameter threaded through `renderBodyMap`, `renderBodyVarDecls`, `renderBodyFlagRegs`, `renderBodyRequiredChecks`, plus `bodyExceedsFlagDepth` helper exposed to templates.
- `internal/generator/templates/command_endpoint.go.tmpl` — `--stdin` flag's help text now reflects truncation: `"Read request body as JSON from stdin (use this for deeply nested fields not exposed as flags)"`.
- Tests added for the cap (`TestBodyMap_DepthCap_*`, `TestBodyExceedsFlagDepth_*`, `TestGenerateBodyDepthCapPreventsCompilerExplosion` end-to-end with an exact 9-body-var count assertion that catches off-by-one regressions).
- Two existing collision tests updated to acknowledge the cap (`TestGenerateDeduplicatesConvergentNestedBodyPaths`, `TestGenerateDeduplicatesCyclicRefBodyShape` — both now verify the deeper leaves are truncated and the rewritten help text appears).

## Behavior

| Body shape | Before | After |
|---|---|---|
| 1-2 levels of nesting | Per-field flags for every leaf | **Unchanged** (no regression) |
| 3+ levels of nesting | 40k+ line file, compile OOM | Per-field flags through depth 2; deeper fields via `--stdin` |
| Multipart / form / `BodyJSONFallback` | Flat emission (no recursion) | **Unchanged** (cap never fires) |

The cap is enforced consistently in all four emitters: when an object recursion would put us at `depth >= maxBodyFlagDepth`, the subtree is skipped uniformly across map construction, var declarations, flag registration, and required-checks.

## Decisions

- **No new golden fixture.** None of the existing 20 fixtures exercise a deep body; the new behavior is covered by direct unit tests on the helpers (`TestBodyMap_DepthCap_*`) plus an integration test through `Generate()` that asserts the rewritten `--stdin` help text appears in the generated source verbatim. Adding a golden for one isolated branch would add maintenance overhead for behavior already pinned at the unit level.
- **Promoted-command stdin gap deferred to #1520.** `command_promoted.go.tmpl` uses the same depth-capped helpers but has no `--stdin` flag, so a promoted POST/PUT with a deep body now builds (good) but silently drops required-field validation for truncated subtrees (a follow-up gap, not a regression — pre-fix those endpoints couldn't compile at all). Filed as a P3 follow-up rather than expanding this PR's scope.

## Test plan

- [x] `go test ./...` — 4427 passed
- [x] `go vet ./...` — clean
- [x] `golangci-lint run ./...` — clean
- [x] `scripts/golden.sh verify` — 20/20 cases pass (no behavior change on shallow specs)
- [x] Boundary test at exact `maxBodyFlagDepth` (`TestBodyMap_DepthCap_Boundary`)
- [x] Shallow regression guard tightened to assert all sibling leaves still emit
- [x] Multipart branch of `bodyExceedsFlagDepth` covered (`TestBodyExceedsFlagDepth_Multipart`)
- [x] OOM-regression test uses exact body-var count (9) — catches `cap=2` (would be 6) and `cap=4` (would be 12) off-by-one regressions

### DIFF
--- a/internal/generator/body_collision_test.go
+++ b/internal/generator/body_collision_test.go
@@ -6,6 +6,7 @@ import (
 	"go/token"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"testing"
 
@@ -262,6 +263,13 @@ func TestGenerateDeduplicatesNestedTreesCollapsedToSameIdent(t *testing.T) {
 // unique identifiers (bodyProjectCustomerName vs
 // bodyProjectPostalAddressCustomerName) without falling back on the _N
 // suffix.
+//
+// maxBodyFlagDepth truncates the deeper
+// Project.PostalAddress.Customer.name leaf (depth 3); only the shallower
+// Project.Customer.name (depth 2) emits. The dedup behavior is still
+// exercised: were the cap raised, the parent-prefix walker would
+// uniquify both leaves without collision. Truncation is verified via
+// the deeper identifier's absence and the --stdin help text rewrite.
 func TestGenerateDeduplicatesConvergentNestedBodyPaths(t *testing.T) {
 	t.Parallel()
 
@@ -304,12 +312,12 @@ func TestGenerateDeduplicatesConvergentNestedBodyPaths(t *testing.T) {
 		"convergent nested paths must produce distinct Go identifiers")
 	assertNoDuplicates(t, flagBindings,
 		"convergent nested paths must register distinct cobra flag names")
-	require.Len(t, bodyVars, 2,
-		"both convergent name leaves must survive as distinct Go identifiers")
+	require.Len(t, bodyVars, 1,
+		"only the shallower project.customer.name survives the depth cap")
 	assert.Contains(t, bodyVars, "bodyProjectCustomerName",
-		"the direct project.customer.name leaf produces this identifier")
-	assert.Contains(t, bodyVars, "bodyProjectPostalAddressCustomerName",
-		"the project.postalAddress.customer.name leaf produces this identifier via path-prefix joining")
+		"the depth-2 project.customer.name leaf still emits its identifier")
+	assert.NotContains(t, bodyVars, "bodyProjectPostalAddressCustomerName",
+		"the depth-3 project.postalAddress.customer.name leaf is truncated by the cap")
 }
 
 // TestGenerateDeduplicatesCyclicRefBodyShape drives the full OpenAPI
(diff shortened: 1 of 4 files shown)


tmchow:
Addressed in 23420ce4: `bodyExceedsFlagDepth` now walks `flattenCollidingBodyFields(endpoint.Body)` so detection and emission read the same tree. Added `TestBodyExceedsFlagDepth_CollisionFlattenedSubtree` as a regression guard.

PR #1524, opened 2026-05-16
fix(cli): move extra_commands SKILL.md block to its own top-level section

### LINKED ISSUE #1451: generator: extra_commands rendered as Cobra commands in SKILL.md Command Reference but never emitted
**Source:** retro for `shopify-pp-cli` (degraded reprint, press v3.2.1 → v4.6.1, 2026-05-15)
**Priority:** P2 · **Component:** generator (`internal/generator/templates/`)

## Problem

Internal-YAML specs declare hand-written extension commands via `extra_commands`:

```yaml
extra_commands:
  - name: bulk-operations
    description: "Run, poll, and inspect Shopify Admin GraphQL bulk operations."
```

`internal/spec/spec.go` parses this into a struct field. `internal/generator/templates/skill.md.tmpl` renders the rows under SKILL.md's "Command Reference" section as if they were canonical Cobra commands:

```
- `shopify-pp-cli bulk-operations` — Run, poll, and inspect Shopify Admin GraphQL bulk operations.
```

The generator's Cobra emitter never reads the field. No Go code is emitted for `bulk-operations`. `verify-skill` walks SKILL.md, treats Command Reference rows as canonical CLI paths, and fails:

```
[unknown-command] shopify-pp-cli bulk-operations: command path not found in internal/cli/*.go (no matching Use: declaration)
  evidence: Command Reference inline
```

The spec contract says "this command exists." The generator's runtime says "no it doesn't." `verify-skill` correctly flags the disagreement.

## Reproduction

Two catalog specs already trip this:
- `catalog/specs/shopify-2026-04-wrapper.yaml` — `bulk-operations`
- `catalog/specs/producthunt-spec.yaml` — `today`, `recent`, `search`

`printing-press generate --spec catalog/specs/producthunt-spec.yaml --output /tmp/ph --force --validate` then `printing-press verify-skill --dir /tmp/ph` exits 1 with `unknown-command` errors for all three `extra_commands` entries.

## Proposed fix

Two cheap options; recommend (a) for v1.

**(a) Stop rendering `extra_commands` under SKILL.md Command Reference.** `skill.md.tmpl` removes the Command Reference block iteration over `extra_commands`. If the descriptions are useful, render them under a separate "Hand-written extensions" prose section that doesn't look like a Cobra-command listing. `verify-skill`'s walker doesn't try to resolve prose mentions.

**(b) Extend the schema.** Add `endpoint:` or `implementation_hint:` to `extra_commands` entries; emit Cobra stubs from generator when present. More work. Defer until the catalog grows enough hand-written commands to justify schema growth.

## Acceptance

- Positive: regenerate `producthunt` with fix (a); `printing-press verify-skill --dir <cli>` exits 0; no `unknown-command` errors for `today` / `recent` / `search`.
- Negative: a CLI that has both an `extra_commands` entry AND a hand-wired Cobra command matching the same name (e.g., shopify's `bulk-operations` once `internal/cli/bulk_operations.go` is restored) still appears in `--help` and SKILL.md. The hand-written file is what registers the command; the spec entry is no longer needed to surface it.

## Evidence

This session: prior `shopify-pp-cli` shipped with `internal/cli/bulk_operations.go` hand-written. Reprint regenerated; `--force` wiped the file. SKILL.md kept the `bulk-operations` row in Command Reference; verify-skill failed. Restored the hand-written file + wired into `root.go`; verify-skill passed. The bug isn't the wiped file (that's reprint-skill territory) — it's that SKILL.md asserts the command exists when only the spec field, not the generator, knows about it.

Retro doc: `/tmp/printing-press/retro/20260515-183655-shopify-retro.md`

tmchow:
## Intent

\`extra_commands\` declared in internal-YAML specs render under a \`**Hand-written commands**\` subsection nested inside \`## Command Reference\`. The Cobra emitter never reads the field, so those rows are promises the generator does not back. \`verify-skill\` walks \`## Command Reference\` for inline backticks shaped like \`<binary> <cmd>\` and (correctly) fails the unknown-command check on every entry. End-to-end repro from the issue: generating \`catalog/specs/producthunt-spec.yaml\` and running \`printing-press verify-skill --dir <out>\` exits 1 with \`unknown-command\` errors for \`today\`, \`recent\`, and \`search\`.

Issue: Closes #1451

## Approach

Option (a) from the issue. Lift the conditional \`**Hand-written commands**\` block out of \`## Command Reference\` into its own sibling \`## Hand-written Extensions\` top-level section, with an explainer paragraph stating these need separate hand-written wiring. The verify-skill walker's \`_extract_inline_commands\` is scoped to the Command Reference body (regex in \`scripts/verify-skill/verify_skill.py:74\`), so a sibling section is invisible to the unknown-command check.

Kept the binary-prefixed backtick rendering inside the new section so humans still see the canonical invocation; the section heading change alone is what defangs the walker.

## Scope

Primary area: \`comp:generator\` (skill.md.tmpl).

Why this belongs in this repo: every CLI generated from a spec with \`extra_commands\` (shopify, producthunt, future internal-YAML specs that declare hand-written extensions) regenerates with the fix on the next print. The runtime branch is conditional on \`{{- if .ExtraCommands}}\` so CLIs without any extras emit a byte-identical SKILL.md.

## Risk

**Published-library drift.** Two already-published CLIs (\`commerce/shopify\`, \`marketing/producthunt\`) ship the old SKILL.md shape. Filed [mvanhorn/printing-press-library#616](https://github.com/mvanhorn/printing-press-library/issues/616) for the sweep tool to retrofit; anchors and idempotency tests called out in the issue body. Until the sweep runs, fresh prints diverge from existing entries for those two CLIs only.

**Test churn.** \`TestSkillRendersExtraCommands\` is updated to assert the new section heading; \`TestSkillNoExtraCommandsIsBackwardCompatible\` is unchanged (the absent-state still renders nothing). Added a belt-and-suspenders regression that mirrors the Python walker's scoping in Go and asserts no extra_commands path is reachable from inside Command Reference.

## Output Contract

No golden fixture diffs — no golden case declares \`extra_commands\`, so the conditional branch is unreached. Output contract for CLIs with \`extra_commands\`: SKILL.md gains a top-level \`## Hand-written Extensions\` section after \`## Command Reference\` (and before \`## Freshness Contract\` when present).

## Verification

- \`go build ./...\` — clean
- \`go vet ./...\` — clean
- \`go test ./...\` — 4429 passed, 35 packages
- \`golangci-lint run ./...\` — no issues
- \`scripts/golden.sh verify\` — 20 cases pass
- End-to-end: \`./printing-press generate --spec catalog/specs/producthunt-spec.yaml --output /tmp/ph_repro --force --validate\` followed by \`./printing-press verify-skill --dir /tmp/ph_repro\` exits 0 (\"All checks passed\"). Issue reproduces on \`main\` before the patch.

## AI / Automation Disclosure

### DIFF
--- a/internal/generator/skill_test.go
+++ b/internal/generator/skill_test.go
@@ -421,6 +421,15 @@ func TestSkillRendersExtraCommands(t *testing.T) {
 		assert.NotContains(t, afterHeading, name,
 			"extra command %q must not appear under Command Reference; the unknown-command walker would flag it", name)
 	}
+
+	// Markdown nests every `###` under the most recent `##`. Placing
+	// `## Hand-written Extensions` BEFORE `### Finding the right command`
+	// would silently reparent that subsection. Assert the ordering so a
+	// future template edit doesn't regress the structure.
+	findingIdx := strings.Index(content, "### Finding the right command")
+	require.NotEqual(t, -1, findingIdx, "Finding the right command subsection should exist")
+	require.Greater(t, extIdx, findingIdx,
+		"Hand-written Extensions must come after ### Finding the right command so it doesn't reparent that subsection")
 }
 
 // TestSkillFrontmatterMetadataIsClawHubCompliantNestedYAML asserts that the
--- a/internal/generator/skill_test.go
+++ b/internal/generator/skill_test.go
@@ -3,6 +3,7 @@ package generator
 import (
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"testing"
 
@@ -358,11 +359,11 @@ func TestSkillRendersAuthBranchPerType(t *testing.T) {
 }
 
 // TestSkillRendersExtraCommands asserts that hand-written commands declared
-// in spec.ExtraCommands appear in the generated SKILL.md Command Reference,
-// after the spec-driven resources, with binary prefix and optional args.
-// This closes the drift class where SKILL.md silently omitted hand-written
-// commands like `today`, `streak`, `rivals` because the template only iterated
-// .Resources.
+// in spec.ExtraCommands appear in their own `## Hand-written Extensions`
+// section (NOT inside `## Command Reference`), with binary prefix and optional
+// args. The separate section keeps verify-skill's unknown-command walker
+// (scoped to ## Command Reference) from treating extra_commands as canonical
+// Cobra paths — that drift was the root cause of #1451.
 func TestSkillRendersExtraCommands(t *testing.T) {
 	t.Parallel()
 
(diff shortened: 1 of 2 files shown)


tmchow:
Greptile P1 addressed in 897fab3c: moved `## Hand-written Extensions` to after `### Finding the right command` so the subsection stays a child of `## Command Reference` in every combination of ExtraCommands and Freshness Contract presence. Added an ordering assertion to lock the structure. End-to-end re-verified on producthunt: `verify-skill` exits 0 and section order is Command Reference -> Finding the right command -> Hand-written Extensions -> Auth Setup.

PR #1529, opened 2026-05-16
fix(cli): scope-aware sync --param dispatch with --global-param fallback

### LINKED ISSUE #1393: generator: sync --param injects into hierarchical/dependent path requests, breaks parent-scoped endpoints
## Summary

`<cli> sync --full --param workspace=<gid>` correctly scopes flat-list endpoints (`/projects`, `/users`, `/tags`) but ALSO injects the param into hierarchical dependent calls (`/projects/<gid>/tasks?workspace=<gid>`), which the API rejects (Asana: `Must specify exactly one of project, tag, section, user task list, or assignee + workspace`).

This retro: required a two-phase manual workaround — flat-list sync with `--param workspace=...`, then dependent-resource sync with `--resources <list>` and NO global param. Took ~15 minutes to diagnose.

## Root cause

`syncUserParams.applyTo()` in `internal/cli/helpers.go` applies global `--param` indiscriminately to every sync request. No distinction between flat-list and hierarchical-dependent requests.

## Cross-API blast radius

Mid-frequency: APIs where (a) some endpoints require a top-level scope param AND (b) dependent endpoints already scope via path.

- **Asana** (confirmed)
- **Jira**: top-level `/users` requires `accountId`; `/issues/{key}/comment` is path-scoped — same conflict if both are sync'd
- **GitHub**: top-level `/issues` requires `filter`/`state`; nested `/repos/{owner}/{repo}/issues` is path-scoped

## Fallback if unfixed

Agent splits sync into two phases manually. Workable but non-obvious; error message ("Must specify exactly one of...") is opaque to the cause.

## Durable fix

Three options, ordered by effort:

1. **Minimal**: `--param` becomes scope-aware (auto-skips for dependent resources by default). Add `--global-param` to preserve old behavior.
2. **Better error**: When the API returns "Must specify exactly one of..." style errors on a dependent path, emit a sync_warning that names the conflicting param ("workspace=... was injected from --param; remove --param or use --resource-param for flat resources only").
3. **Durable**: Generator auto-detects required top-level scope params from the spec and auto-injects them for flat-list endpoints only — no `--param` workaround needed.

## Acceptance criteria

- **Positive**: `sync --full --param workspace=<gid>` against an API with both flat-list and hierarchical endpoints succeeds end-to-end, populating both layers, without per-resource manual splitting.
- **Positive**: Explicit `--global-param key=value` still applies to every request (preserves power-user override).
- **Negative**: Existing single-phase syncs (flat-only) unchanged.

## Complexity

Medium (scope-aware param dispatch — touches sync.go template, helpers.go, and the dependent-sync dispatch).

## Artifacts

- [Retro doc](https://files.catbox.moe/unp7t0.md)

tmchow:
## Summary

Closes #1393. \`--param\` on the sync command no longer double-injects top-level scope params (workspace, accountId, etc.) onto path-scoped dependent requests. New \`--global-param\` flag preserves the prior apply-everywhere semantic for operators who relied on it.

## The bug

Running \`<cli> sync --full --param workspace=<gid>\` against an API with both flat-list and hierarchical endpoints (Asana, Jira, GitHub) made the dependent path call \`/projects/<gid>/tasks?workspace=<gid>\` — and Asana-style APIs reject that with \"Must specify exactly one of project, tag, section, user task list, or assignee + workspace\". Workaround: manually split sync into two phases (flat-list with \`--param\`, then dependent without). 

## What changed

\`internal/generator/templates/helpers.go.tmpl\`:
- \`syncUserParams.global\` -> \`flatGlobal\` + \`trueGlobal\` + \`perResource\`.
- New \`parseSyncKVFlags(flags, flagName)\` helper deduplicates the \`--param\` / \`--global-param\` parse blocks.
- \`applyTo(resource, params, isDependent bool)\`: flat-list path applies all three maps; dependent path skips flatGlobal.

\`internal/generator/templates/sync.go.tmpl\`:
- New \`--global-param\` flag (repeatable key=value) registered alongside \`--param\` and \`--resource-param\`.
- Two \`applyTo\` call sites pass \`isDependent\` explicitly (flat: false, dependent: true).
- \`--param\` help text updated to flag the new scope-aware behavior.

Golden fixtures regenerated for the five sync-touching cases.

## Behavior

| Flag | Flat-list requests | Dependent path-scoped requests |
|------|-------------------|--------------------------------|
| \`--param key=value\` | applied | **skipped** (new) |
| \`--global-param key=value\` (new) | applied | applied |
| \`--resource-param res:key=value\` | applied to res | applied to res |

Existing single-phase syncs on APIs without dependent resources behave identically.

## Test plan

- [x] \`go test ./...\` — 4443 passed
- [x] \`go vet\` — clean
- [x] \`golangci-lint run\` — clean
- [x] \`scripts/golden.sh verify\` — 20/20 cases pass after intentional regeneration
- [x] New test \`TestGenerateSyncDependentSkipsFlatGlobalParam\` asserts dependent call site passes \`isDependent=true\` and flat call site passes \`isDependent=false\`
- [x] New test \`TestGenerateSyncUserParamsHelperRespectsFlatVsTrueGlobal\` pins the emitted helper signature and the \`if !isDependent\` gate
- [x] Existing \`TestGenerateSyncParamPassthrough\` updated for the three-arg \`parseSyncUserParams\` and the \`isDependent\` argument

## Operator migration

For APIs **without** dependent resources: no change.

For APIs **with** dependent resources where the old behavior was load-bearing (a scope param genuinely needed everywhere): switch from \`--param key=value\` to \`--global-param key=value\`.

### DIFF
--- a/internal/generator/sync_param_passthrough_test.go
+++ b/internal/generator/sync_param_passthrough_test.go
@@ -15,10 +15,11 @@ import (
 )
 
 // TestGenerateSyncParamPassthrough verifies the sync template emits the
-// --param / --resource-param flags, parses them through parseSyncUserParams,
-// and applies them after spec-derived params. Some APIs mark filter params
-// optional in the spec but reject requests without them at runtime; without
-// this passthrough, the only workaround is hand-editing the generated client.
+// --param / --resource-param / --global-param flags, parses them through
+// parseSyncUserParams, and applies them after spec-derived params. Some
+// APIs mark filter params optional in the spec but reject requests
+// without them at runtime; without this passthrough, the only workaround
+// is hand-editing the generated client.
 func TestGenerateSyncParamPassthrough(t *testing.T) {
 	t.Parallel()
 
@@ -37,6 +38,8 @@ func TestGenerateSyncParamPassthrough(t *testing.T) {
 		"sync should expose a repeatable --param flag")
 	assert.Contains(t, syncSrc, `cmd.Flags().StringArrayVar(&resourceParamFlags, "resource-param"`,
 		"sync should expose a repeatable --resource-param flag")
+	assert.Contains(t, syncSrc, `cmd.Flags().StringArrayVar(&globalParamFlags, "global-param"`,
+		"sync should expose a repeatable --global-param flag for the apply-everywhere semantic")
 	assert.Contains(t, syncSrc, "key=value",
 		"--param help text should describe the key=value shape")
 	assert.Contains(t, syncSrc, "resource:key=value",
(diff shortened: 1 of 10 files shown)


tmchow:
Addressed in b9fa61e2: renamed the second parameter of `parseSyncUserParams` from `perResourceFlags` to `resourceParamFlags` to match the call-site name.

PR #1531, opened 2026-05-16
fix(cli): writeThroughCache caches single-object responses keyed by non-id PKs

### LINKED ISSUE #1439: generator: writeThroughCache silently skips single-object responses with non-"id" primary keys
> Filed by `/printing-press-retro` from a run of **PCGS CoinFacts** (`pcgs-pp-cli`)
> (scorecard 79/100, 100% verify pass).

## Summary

`writeThroughCache` in the generated `data_source.go` guards single-object cache writes with `if _, ok := envelope["id"]; ok`. APIs that use a non-`"id"` primary key field (e.g. `CertNo`, `sku`, `invoiceId`) never match this guard. Every lookup command returns HTTP 200 and the user sees correct output, but the response is silently not cached. Every downstream feature that reads from the local cache (smart presets, holdings analysis, offline search) gets no data.

## Where to look

- **Component:** generator
- **Likely area:** Generator template in `internal/generator/` that emits `data_source.go` — specifically the `writeThroughCache` function, single-object detection branch
- **Triggered when:** API returns single-object JSON responses where the top-level key identifying the resource is anything other than `"id"`

## What we observed

In `pcgs-pp-cli`, `internal/cli/data_source.go` (generated):

```go
// original generated code
if _, ok := envelope["id"]; ok {
    _, _, _ = db.UpsertBatch(resourceType, []json.RawMessage{data})
    return
}
```

PCGS cert responses use `"CertNo"` as the primary key. `envelope["id"]` was always false. Every `cert lookup` call succeeded against the live API but wrote nothing to the local SQLite cache. `collection smart` presets, `cert holdings` mintage lookups, and `price compare`'s Numista column all depended on cached data that was never written.

Confirmed via:
- Direct `sqlite3` query after 6 `cert lookup` calls: 0 new rows added
- Tracing: `writeThroughCache` → `"id"` guard never fires → `UpsertBatch` never called
- Fix: replaced guard with `if items == nil && len(envelope) > 0` — lets `UpsertBatch`'s existing `resourceIDFieldOverrides` mechanism handle primary key resolution. All 6 certs cached on next lookup.

The symptom is especially hard to find because `cert lookup` reports success and shows correct output — the broken path is the silent side effect.

## Suspected root cause

**High confidence:** The template hardcodes `"id"` as the sentinel for "this is a single object, attempt to upsert." `UpsertBatch` already has `resourceIDFieldOverrides` (a per-resource-type map of primary key field names) specifically for non-`"id"` APIs. The gate in `writeThroughCache` prevents `UpsertBatch` from running, bypassing the mechanism that already exists to handle this.

## Suggested direction

Replace the `"id"` guard with a shape-based check that fires for any non-decomposed envelope:

```go
// After (proposed template):
// Single object — let UpsertBatch's ID extraction decide.
// UpsertBatch silently skips items with no resolvable primary key,
// so this is safe for any resource.
if items == nil && len(envelope) > 0 {
    _, _, _ = db.UpsertBatch(resourceType, []json.RawMessage{data})
    return
}
```

This change is safe for standard `"id"` APIs: `UpsertBatch` resolves `"id"` correctly via its existing heuristic. It's also safe for "no primary key found" cases — `UpsertBatch` silently skips those, same behavior as before.

## Acceptance criteria

- positive: generate a CLI for an API with a non-`"id"` primary key single-object response (e.g. `certNo`, `sku`, `handle`); call a lookup command; verify the response is stored in local SQLite
- negative: standard `"id"`-keyed API still caches correctly (no regression)

## Frequency

subclass: APIs with non-`"id"` primary key fields in single-object responses

## Complexity

small

## Scope boundary

Does NOT change `UpsertBatch` logic. Does NOT add `resourceIDFieldOverrides` entries (caller responsibility). Only removes the specific `"id"` guard from `writeThroughCache`.

## Dependencies

None. (`resourceIDFieldOverrides` must be populated by the caller for the upsert to resolve the key — that is a separate concern from this guard removal.)

## Related issues

- #1394 — open issue (`related-area`): PK heuristic picks `'name'` for non-id APIs — adjacent territory: both are about the generator not handling non-`"id"` primary keys correctly in different subsystems

## Artifacts

| Artifact | Link |
|----------|------|
| Retro document | https://files.catbox.moe/vn8rdy.md |
| Manuscripts (research + proofs) | https://files.catbox.moe/s2irez.zip |
| Generated CLI source code | https://files.catbox.moe/jpn6oj.zip |

---

tmchow:
## Intent

\`writeThroughCache\` in the generated \`data_source.go\` guards single-object cache writes with \`if _, ok := envelope[\"id\"]; ok\`. APIs whose primary key is named anything other than \`id\` (\`CertNo\`, \`sku\`, \`invoiceId\`, etc.) never match — every lookup returns HTTP 200 with correct stdout while the local SQLite store stays empty. Every downstream feature that reads from the cache (smart presets, holdings analysis, offline search) silently gets no data.

Issue: Closes #1439

## Approach

Two changes in the same template branch:

1. **Drop the hardcoded \`\"id\"\` guard.** Delegate primary-key resolution to \`UpsertBatch\`'s existing \`resourceIDFieldOverrides\` mechanism, which the parser already populates from \`x-resource-id\` and the response-schema inference chain. Matches the reporter's validated fix.
2. **Tighten the single-object branch.** Skip envelopes whose only meaningful content is a known list-wrapper key (\`results\`, \`data\`, \`items\`) with an empty array — \`{\"items\": []}\` must not write a row keyed by the entire envelope. The existing wrapper-array loop above handles non-empty list responses; this guard preserves that invariant for the empty case.

## Scope

Primary area: \`comp:generator\` (\`internal/generator/templates/data_source.go.tmpl\`).

Why this belongs in this repo: every printed CLI that exposes a lookup-style detail endpoint regenerates with the fix on the next print. No spec changes required — the existing \`IDField\` plumbing handles all variants.

## Risk

**False positive avoided.** The first fix attempt triggered \`TestClientBasePathLiveRequest\` (\`{\"items\": []}\` was being upserted as a single object → \"no extractable ID field\" warning). The list-wrapper skip handles that and is locked in by \`TestWriteThroughCacheSkipsEmptyListEnvelope\`.

**Existing \`id\`-keyed APIs unchanged.** The \`UpsertBatch\` path already handled them; this PR removes a hardcoded shortcut that bypassed it for the common case. \`TestWriteThroughCachePopulatesTypedTable\` continues to pass.

## Output Contract

No golden fixture diffs — no golden case exercises a non-id-PK single-object response. Output contract for CLIs whose detail endpoints carry non-\`id\` PKs: \`writeThroughCache\` now upserts the row into the typed table using the spec-declared \`IDField\`. Cache-dependent features (smart presets, offline search) work after first lookup instead of silently returning empty.

## Verification

- \`go build ./...\` — clean
- \`go vet ./...\` — clean
- \`go test ./...\` — 4455 passed, 35 packages
- \`golangci-lint run ./...\` — no issues
- \`scripts/golden.sh verify\` — 20 cases pass
- Inline runtime tests: \`TestWriteThroughCacheNonIDPrimaryKey\` (asserts the CertNo-shaped response lands in the typed table), \`TestWriteThroughCacheSkipsEmptyListEnvelope\` (asserts the empty-list shape never reaches UpsertBatch)

## AI / Automation Disclosure

### DIFF
--- a/internal/generator/generator_test.go
+++ b/internal/generator/generator_test.go
@@ -3742,12 +3742,62 @@ func TestWriteThroughCacheCachesObjectWithNullWrapperFieldName(t *testing.T) {
 		t.Fatalf("certs count after null-wrapper-field response = %d, want 1 (JSON null is not an empty array)", count)
 	}
 }
+
+// TestWriteThroughCacheCachesObjectWithEmptyListWrapperAndOtherFields
+// covers the multi-key case Greptile flagged in round 3: a detail object
+// with real per-row fields PLUS a wrapper-named field that happens to be
+// an empty array. The skip branch must only fire when EVERY top-level
+// key is a list wrapper or known pagination metadata.
+func TestWriteThroughCacheCachesObjectWithEmptyListWrapperAndOtherFields(t *testing.T) {
+	t.Setenv("HOME", t.TempDir())
+
+	writeThroughCache(context.Background(), "certs", json.RawMessage(` + "`" + `{"CertNo":"88888","items":[],"Grade":"PR69"}` + "`" + `))
+
+	db, err := store.Open(defaultDBPath("pcgs-pp-cli"))
+	if err != nil {
+		t.Fatalf("open cache store: %v", err)
+	}
+	defer db.Close()
+
+	var count int
+	if err := db.DB().QueryRow(` + "`" + `SELECT COUNT(*) FROM certs WHERE id = '88888'` + "`" + `).Scan(&count); err != nil {
+		t.Fatalf("query certs: %v", err)
+	}
+	if count != 1 {
+		t.Fatalf("certs count after detail-with-empty-wrapper-field response = %d, want 1 (envelope has real per-row fields and must be cached)", count)
+	}
+}
+
+// TestWriteThroughCacheSkipsListEnvelopeWithPaginationMetadata guards the
+// flip side: a real list envelope with pagination metadata fields
+// (next_cursor, has_more, etc.) plus an empty array must still skip
+// the single-object upsert path. The metadata allowlist is what keeps
+// these envelopes recognized as lists.
+func TestWriteThroughCacheSkipsListEnvelopeWithPaginationMetadata(t *testing.T) {
+	t.Setenv("HOME", t.TempDir())
+
+	writeThroughCache(context.Background(), "certs", json.RawMessage(` + "`" + `{"items":[],"next_cursor":"","has_more":false}` + "`" + `))
+
+	db, err := store.Open(defaultDBPath("pcgs-pp-cli"))
+	if err != nil {
+		t.Fatalf("open cache store: %v", err)
+	}
+	defer db.Close()
+
+	var count int
+	if err := db.DB().QueryRow(` + "`" + `SELECT COUNT(*) FROM certs` + "`" + `).Scan(&count); err != nil {
+		t.Fatalf("query certs: %v", err)
+	}
+	if count != 0 {
+		t.Fatalf("certs count after list-envelope-with-metadata response = %d, want 0 (no real per-row data, must skip)", count)
+	}
+}
 `
 	testPath := filepath.Join(outputDir, "internal", "cli", "write_through_cache_non_id_test.go")
 	require.NoError(t, os.WriteFile(testPath, []byte(inlineTest), 0o644))
 
 	runGoCommandRequired(t, outputDir, "mod", "tidy")
-	runGoCommandRequired(t, outputDir, "test", "-run", "TestWriteThroughCacheNonIDPrimaryKey|TestWriteThroughCacheSkipsEmptyListEnvelope|TestWriteThroughCacheCachesObjectWithListWrapperFieldName|TestWriteThroughCacheCachesObjectWithNullWrapperFieldName", "./internal/cli")
+	runGoCommandRequired(t, outputDir, "test", "-run", "TestWriteThroughCacheNonIDPrimaryKey|TestWriteThroughCacheSkipsEmptyListEnvelope|TestWriteThroughCacheCachesObjectWithListWrapperFieldName|TestWriteThroughCacheCachesObjectWithNullWrapperFieldName|TestWriteThroughCacheCachesObjectWithEmptyListWrapperAndOtherFields|TestWriteThroughCacheSkipsListEnvelopeWithPaginationMetadata", "./internal/cli")
 }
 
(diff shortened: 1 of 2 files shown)


tmchow:
Greptile P1 addressed in f6e43ae4: the list-envelope guard now requires the wrapper-key value to actually unmarshal as a JSON array. A scalar-valued field whose name collides with `data` / `results` / `items` (e.g. `{"id":"tx","data":"<opaque>"}`) is treated as a regular field, not a list envelope, and the row still gets cached. Added `TestWriteThroughCacheCachesObjectWithListWrapperFieldName` as explicit coverage.

tmchow:
Greptile P1 addressed in e062f17e: added `arr != nil` to the wrapper-key unmarshal check so JSON `null` falls through to UpsertBatch instead of being misclassified as an empty list. Added `TestWriteThroughCacheCachesObjectWithNullWrapperFieldName` as explicit coverage.

tmchow:
Mergify dequeued on a flaky pipeline test, not a real bug in this PR.

Failure: `TestDiscoverExampleCheckCommandsIgnoresStderrNoise` exited with `fork/exec /tmp/.../fakebin: text file busy`. That's a Linux race between `os.WriteFile` and `exec.Cmd.Run` on the same path — the kernel still holds a writable fd open when the exec syscall fires and returns `EBUSY`. The test lives in `internal/pipeline/dogfood_test.go` and is unrelated to this PR's diff (only `internal/generator/templates/data_source.go.tmpl` and `internal/generator/generator_test.go` changed).

Local repro: `go test ./internal/pipeline/ -run TestDiscoverExampleCheckCommandsIgnoresStderrNoise -count=1` x5 → all pass. `main` CI history shows the test consistently green.

@mergifyio queue

tmchow:
Greptile round-3 P1 addressed in 63dbe39c. Switched from the symptom-level fix (check that wrapper value is an array) to a structural check: an envelope is treated as list-shaped only when EVERY top-level key is either a list wrapper itself or a known pagination-metadata key (next_cursor, has_more, total, links, meta, response_metadata, etc.). One non-metadata key (`id`, `status`, etc.) signals real per-row data and the row gets cached. Added two regression tests: `TestWriteThroughCacheCachesObjectWithEmptyListWrapperAndOtherFields` for the multi-key detail shape, `TestWriteThroughCacheSkipsListEnvelopeWithPaginationMetadata` to confirm the metadata allowlist still recognizes real list envelopes.

PR #1619, opened 2026-05-18
feat(ci): mirror supply-chain hardening from printing-press-library

tmchow:
## Summary

Mirrors the supply-chain hardening gate that landed in [mvanhorn/printing-press-library#665](https://github.com/mvanhorn/printing-press-library/pull/665) into this generator repo. Same two-layer defense (Greptile rules + deterministic Python scan) adapted to the generator-repo signal set.

## What this PR does

Two layers, applied to PRs touching `.github/workflows/**`:

1. **`verify-supply-chain.yml`** — deterministic Python scan at `.github/scripts/verify-supply-chain/scan.py`, vendored from the published-library repo (source dated 2026-05-17). `scan.py` is verbatim; `signals.py` is adapted to the generator-repo signal set. 15 unit + integration tests pass.

2. **Four `greptile.json` rules** covering the workflow-trust + Go-module env attack shapes plus a judgment-only credential-path read rule for `internal/**/*.go` and `cmd/**/*.go`.

## Scope adaptations (vs. printing-press-library copy)

Signals that don't apply here are **omitted by design**, not stubbed:

| Signal | Adapted? | Why |
|---|---|---|
| R1: `pull_request_target` + PR-head checkout | ✅ same | Workflow trust shape is identical across repos |
| R2: `id-token: write` outside allowlist | ✅ empty allowlist | This repo has no OIDC workflows on origin/main; `release.yml` uses keyed cosign signing. Any future `id-token` grant must be allowlisted explicitly. |
| R3: `replace` directive in library/**/go.mod | ❌ omitted | No `library/**` here — published CLIs live in the other repo where R3 protects them |
| R4: GOPROXY / GOFLAGS / GONOSUMCHECK in workflow env | ✅ same | |
| R5: `postinstall`/`preinstall`/`prepare` in npm/package.json | ❌ omitted | No npm wrapper here |
| R6: `module` directive drift on existing library go.mod | ❌ omitted | No published-CLI module paths here |
| U4 Greptile rule: Go-source credential-path reads | ✅ adapted prose | Re-scoped to `internal/**` and `cmd/**` for the generator's own Go code |

## Current state audit

Performed against fresh `origin/main` as of 2026-05-17:

- ✅ Only `pull_request_target` workflow (`conversation-resolution-check.yml`) is safe — `contents: read | pull-requests: read`, no PR-head checkout, no `id-token`.
- ✅ No `id-token: write` in any workflow.
- ✅ No `GOPROXY` / `GOFLAGS` / `GONOSUMCHECK` overrides.

All new rules are forward-looking gates.

## Single source of truth for incident details

Canonical incident timeline + primary-source citations live in [mvanhorn/printing-press-library/docs/solutions/security/2026-05-supply-chain-hardening.md](https://github.com/mvanhorn/printing-press-library/blob/main/docs/solutions/security/2026-05-supply-chain-hardening.md). This repo's AGENTS.md cross-links there rather than duplicating.

## Rollout posture

Runs informationally on landing. Promote to required check via branch protection after a one-week green window on the published-library PR confirms no false-positive miscalibration. Staged deployment per the published-library plan's KD6.

## Test plan

- [x] 15 unit + integration tests pass: `cd .github/scripts/verify-supply-chain && python3 -m unittest scan_test`
- [x] `signals.py` adaptation review — confirm R3/R5/R6 absence is intentional and the empty `ID_TOKEN_ALLOWLIST` is correct for current state
- [ ] Sandbox PR with each attack shape (R1, R2, R4) produces the expected red workflow + Greptile P0 — to exercise during the informational window.


tmchow:
Thanks — the scanner-self-bypass concern is real and was already addressed by the time this review posted; the review fired against an earlier commit. Two commits land the fix:

- [`02d3d225`](https://github.com/mvanhorn/cli-printing-press/pull/1619/commits/02d3d22515ddb35ec73ce63a3e90efa8b18edcb0) — rewrote the workflow to check out the **base** branch from the base repo, then fetch the PR head SHA as a detached ref and pass it to `scan.py` via `--head-ref`. The `scan.py` executing the gate is now the trusted base version.
- [`2fc39dbe`](https://github.com/mvanhorn/cli-printing-press/pull/1619/commits/2fc39dbe96ae366b1fcc84cf71957f823644e952) — bootstrap fallback: when base has no scanner yet (the PR introducing it), restore the scanner from the PR head SHA via `git restore --source`. No-op after merge.

Could you re-review the current HEAD (`2fc39dbe`)? If the bootstrap-restore approach still leaves an exploit window worth closing before required-check promotion, happy to iterate.

@greptileai review

tmchow:
The compact YAML inline-list form (`on: [pull_request_target]`) is already fixed in [\`f22e3d4\`](https://github.com/mvanhorn/cli-printing-press/pull/1619/commits/f22e3d44) — that commit landed after this review fired (reviewed commit was \`2fc39dbe\`, the bootstrap fix).

The fix adds a second regex `_PR_TARGET_TRIGGER_FLOW` (`^\\s*on\\s*:\\s*\\[[^\\]\\n]*\\bpull_request_target\\b`) unified via `_has_pr_target_trigger()`. Tests include `test_flow_sequence_trigger_blocks` covering the exact `on: [pull_request_target, push]` shape. The mirror of the same fix landed simultaneously in [printing-press-library#665](https://github.com/mvanhorn/printing-press-library/pull/665).

Greptile should re-review the new HEAD. Re-triggering manually if needed.

@greptileai review

tmchow:
Greptile is right that SHA-pinning third-party actions is the supply-chain-hardening best practice. Two reasons I'm deferring it rather than pinning here:

1. **Repo-wide consistency**: every existing workflow in both this repo and printing-press-library uses floating major-version tags — `actions/checkout@v6`, `actions/setup-go@v6`, `actions/cache@v4`, `goreleaser/goreleaser-action@v7`, `sigstore/cosign-installer@v4.1.0`, etc. Pinning the single new workflow to a SHA while every neighbor floats creates inconsistency without meaningfully improving the security posture — an attacker compromising `actions/checkout@v6` reaches every CI run, not just this one.

2. **Scope**: SHA-pinning is a repo-wide convention change that belongs in a dedicated sweep PR with a coordinated dependabot config (since SHA-pinned actions need a different update bot than tag-pinned ones). Bundling it into this PR mixes concerns and complicates the rollback story.

I've noted this in my follow-up list — happy to land a dedicated SHA-pinning sweep PR for both repos as a follow-up. Closing this thread as "valid finding, out-of-scope for this PR".

For the rest: the bootstrap path is one-time (no-op after this PR merges; base will have scan.py thereafter), and the scanner-from-base posture handles all subsequent PRs correctly. Sounds like the 4/5 is acceptable for landing as the informational gate.

Also pushed a mirror of the printing-press-library scan.py signature change (rename-aware `changed_files()` returning `old_path`) for cross-repo parity — no behavioral change here since R3/R5/R6 are still omitted.

tmchow:
@greptileai review

PR #1625, opened 2026-05-18
fix(cli): dogfood prefers bundled <dir>/spec.yaml over caller --spec

### LINKED ISSUE #1620: scorer: dogfood should prefer bundled <cli-dir>/spec.yaml over caller-passed --spec
> Filed from a tenderned printed-CLI polish run. Initial dogfood reported `Path Validity: 0/2` and `Verdict: FAIL` because the caller passed the upstream OpenAPI spec (`.manuscripts/<run-id>/research/tenderned-spec.yaml` — documents only the 1 auth'd XML endpoint TenderNed officially publishes) instead of the CLI's own internal `spec.yaml` (which enumerates the 7 endpoints the CLI actually implements). Polish discovered this on re-run with `--spec $CLI_DIR/spec.yaml` — Path Validity jumped to 10/10, scorecard 69 → 84.

## Symptom

`printing-press dogfood --dir <cli-dir> --spec <some-spec.yaml>` treats `--spec` as opaque caller input. When the caller picks the wrong spec from a multi-spec manuscripts directory (which is what happens by default for browser-sniff / crowd-sniff / hand-authored-on-top-of-official runs), dogfood:

1. Compares CLI's paths against the wrong spec
2. Reports `Path Validity: 0/N` even though the CLI is correctly built against its own internal spec
3. Cascades into `Auth Protocol: MISMATCH`, `Type Fidelity: 3/5`, and other downstream score regressions
4. Drives polish's verdict to `further_polish_recommended: yes` for an issue that doesn't exist

The CLI's own bundled `spec.yaml` at `<cli-dir>/spec.yaml` is the authoritative spec for what was actually generated (per `publish package`'s contract). Dogfood doesn't prefer it.

## Cross-CLI evidence

Multi-spec runs are common, not exotic:
- **Every browser-sniff-enriched run** has ≥2 specs: the original + the sniffed spec.
- **Every crowd-sniff-enriched run** has ≥2 specs: the original + the crowd-derived spec.
- **Every hand-authored-on-top-of-partial-official run** (like tenderned) has ≥2 specs: the partial official + the hand-authored full spec.

Catalog instances I can name with evidence:
- `factor75/` — browser-sniff retro [#1598](https://github.com/mvanhorn/cli-printing-press/issues/1598) explicitly mentions the browser-sniffed + original spec coexistence in its hand-authored sync.
- `recipe-goat/` — browser-sniff per its prior retros.
- `tenderned/` — this run (hand-authored 7-endpoint internal YAML on top of the official 1-endpoint Swagger).

The score-regression symptom recurs anywhere the caller picks a smaller / different spec than the CLI was actually generated from.

## Proposed fix

In `printing-press dogfood`'s spec resolution (`internal/dogfood/...` is the likely location; trace from where the `--spec` flag is parsed):

1. **Prefer `<dir>/spec.yaml` over caller-passed `--spec`** when the bundled file exists. The bundled spec is authoritative by definition — it was placed there by `publish package` as the canonical spec for the printed CLI.
2. **Print a single-line header at run start** naming the spec actually used:
   ```
   dogfood: using spec <abs-path> (bundled)
   ```
   or, when the bundled is absent and `--spec` is honored:
   ```
   dogfood: using spec <abs-path> (--spec)
   ```
3. **Warn when both are present and differ:**
   ```
   dogfood: caller --spec=<path> overridden by bundled <dir>/spec.yaml
   ```
   So the caller learns and updates their invocation.

## Acceptance criteria

- **Positive:** generate a CLI from an internal-format spec with N endpoints. Verify `<cli-dir>/spec.yaml` exists. Run `printing-press dogfood --dir <cli-dir> --spec <some-other-spec.yaml>` and assert (a) the spec actually loaded is `<cli-dir>/spec.yaml`, (b) the run header reports it as such, and (c) Path Validity scores against the bundled N endpoints (not the caller's smaller set).
- **Negative:** when `<cli-dir>/spec.yaml` is missing (legacy / orphan / pre-publish-package run), fall through to `--spec` exactly as today. No regression on existing callers.

## Scope boundary

- Don't change `publish validate`'s spec resolution. It uses its own contract and isn't affected by this finding.
- Don't change generator behavior or `publish package`'s spec-bundling step (the bundled `spec.yaml` is already correct; just need dogfood to use it).
- Don't change polish's invocation. Polish calls dogfood with the spec it discovers from manuscripts; this fix lets dogfood ignore that input when bundled is present, which is the right semantic.

---

🤖 Filed from `/printing-press-retro tenderned`. Full retro doc: `/tmp/printing-press/retro/20260518-094500-tenderned-retro.md` (local).

markvandeven:
## Summary

Fixes #1620. `printing-press dogfood --dir X --spec Y` now prefers the spec archived at `<X>/spec.{json,yaml,yml}` (placed there by `publish package`) over the caller-passed `--spec`. The bundled spec is authoritative for the printed CLI; preferring it eliminates false-negative Path Validity / Auth Protocol regressions on multi-spec manuscripts runs (browser-sniff, crowd-sniff, hand-authored-on-top-of-official) where the caller reaches into the manuscripts dir and picks the upstream spec instead.

Falls through to caller's `--spec` unchanged when no bundled spec exists — no regression for legacy / pre-`publish package` CLI dirs.

## Changes

- `internal/pipeline/dogfood.go`
  - New `resolveDogfoodSpec(dir, callerSpec)` helper picks `<dir>/spec.{json,yaml,yml}` (search order mirrors `findArchivedSpec`) over caller's `--spec`.
  - `RunDogfood` emits two stderr lines matching the proposal exactly:
    - `dogfood: using spec <abs-path> (bundled)` or `(--spec)` whenever a spec is in play.
    - `dogfood: caller --spec=<path> overridden by bundled <abs-path>` when bundled wins over a differing caller path.
  - New `DogfoodReport.SpecSource` field (`"bundled"` / `"caller"` / `""`) so JSON consumers can see the resolution programmatically.
- `internal/pipeline/dogfood_test.go` — 7 new tests:
  - 5 resolver unit tests: bundled-over-caller, no-override-when-caller-equals-bundled, fallthrough-when-no-bundle, empty-when-neither-present, spec.json-preferred-over-spec.yaml.
  - 2 end-to-end `RunDogfood` tests: bundled overrides smaller caller spec (Path Validity 2/2 not 1/2); fallthrough preserved when bundled absent.

## Scope respected (per #1620)

- Did not touch `publish validate`, generator, `publish package`'s bundling step, or polish.
- Polish keeps calling dogfood with whatever spec it found in manuscripts; this fix simply lets dogfood ignore that input when the bundled is present.

## Test plan

- [x] `go test ./internal/pipeline/... ./internal/cli/...` — all green
- [x] New resolver unit tests cover acceptance criteria (positive + negative) from #1620


markvandeven:
**Heads-up on a broader behavior change surfaced by the golden fixture refresh**

The original framing of #1620 was \"prefer bundled over caller's \`--spec\`\". My implementation goes one step further: when caller passes **no** \`--spec\` and a bundled spec exists, dogfood now auto-discovers it and runs the full check matrix against it. Previously, the empty-\`--spec\` branch skipped auth_check and browser_session_check entirely.

This surfaced in the `generate-golden-api` golden, whose command runs `printing-press dogfood --dir X --json` with no \`--spec\`. Diff in the refreshed `dogfood.json`:

\`\`\`diff
-    \"detail\": \"spec not provided; auth protocol check skipped\",
-    \"generated_format\": \"\",
+    \"detail\": \"spec not provided or no bot/bearer/basic scheme detected\",
+    \"generated_format\": \"unknown\",
\`\`\`

I believe this is the right semantic — if the spec is sitting right there in the CLI dir, dogfood should use it regardless of whether the caller passed it explicitly. The previous behavior was a degraded mode that masked auth-check signal for the common \`dogfood --dir X\` invocation. Happy to narrow the resolver to only fire when caller passes \`--spec\` if reviewers prefer the literal proposal scope — let me know.

Issue #1598, opened 2026-05-17
[retro/factor75] 3 findings — personalization-param drop in hand-authored sync; looksLikeJWT no length floor; auth login --chrome can't reach Auth0-SPA-in-memory tokens


npwalker:
> Filed from a live dogfooding session of the **factor75** printed CLI against a real user account. The trigger was `deliveries pending` misclassifying the user's already-customized week as "auto-preselected"; root-cause work surfaced three independent machine-side gaps that can compound on any sniffed BFF CLI.
>
> A matching printed-CLI fix shipped in the same change (Factor75 syncer now forwards `servings` + `delivery-option` to `/gw/my-deliveries/menu`). The three WUs below are systemic — they protect against the next CLI making the same mistake.

This retro packages three findings into one issue because they share a debugging arc (a single broken-cart symptom) but live in different components and have independent acceptance criteria. They can be picked up separately.

---

### Symptom

Factor75's `internal/syncer/syncer.go` calls `GET /gw/my-deliveries/menu` with 5 query params (`week`, `country`, `locale`, `subscription`, `product-sku`). The website (live browser-sniff capture in this session) calls the same endpoint with 11 params, adding `servings`, `delivery-option`, `postcode`, `preference`, `customerPlanId`, and `include-future-feedback`.

The gateway responds to the 5-param shape with a generic plan preselect (3 distinct meals × qty 6 = 18 to fill a 15-meal plan). It responds to the 11-param shape with the user's actual cart (11 distinct meals × qty 1–3 = 15 exactly). The two responses have the **same JSON structure** — only `meals[*].selection.quantity` differs. Structural validation never flags it.

Downstream effects compound: `deliveries pending` classifies the week as `auto-preselected` (correctly given the data, wrong against reality), `favorites suggest` builds its recency-exclusion window from a wrong "last ordered" picture, `history` analytics has been silently inflating counts for the three preselect meals for as long as the CLI has been running.

### Root cause

Hand-authored sync code under `internal/syncer/` is exempt from `cobratree`'s mechanical endpoint surface checks and from the scorer's `auth_protocol`/`reimplementation_check`. The agent who wrote the syncer picked the params that looked obviously needed (sub ID, locale, SKU) and dropped the rest. There is no skill checklist gate or scorer rule that flags param-cardinality drop against the sniff capture for the same endpoint.

### Cross-API evidence (3 concrete APIs)

[...]

PR #1645, opened 2026-05-19
feat(cli): detect Auth0 SPA in-memory auth + emit CDP extractor

npwalker:
## Intent

Auth0 SPA SDK v2+ deployments running the default `cacheLocation: memory` keep the access token in JS heap memory — never in cookies, never in localStorage. The existing `auth login --chrome` cookie-jar extractor has no path to those tokens, so generated CLIs for sites like Factor75 / HelloFresh / Green Chef / EveryPlate (shared BFF) have no automatable login flow. PR #1641 ships the `auth set-token` manual-paste escape hatch (WU-3a). This PR adds the *automated* extraction path.

Issue: Refs #1598 (WU-3b)

## Approach

Two pieces:

1. **Sniff-time detection** — `internal/browsersniff/auth0_spa.go` flags a capture as `auth0_spa_in_memory` when it has `/oauth/token` calls that return `access_token` in the body and no JWT-shaped `Set-Cookie` is observed on the same response. Sets `Auth.Subtype = \"auth0_spa_in_memory\"` on the generated spec via a new `x-auth-subtype` extension (round-trips through `internal/openapi/parser.go`, documented in `docs/SPEC-EXTENSIONS.md`). Falls through silently for any other shape — preserves current behavior.
2. **Runtime CDP extraction** — `auth_browser.go.tmpl` gains an `{{- else if \$hasAuth0SPA}}` branch on `newAuthLoginCmd` that accepts `--auth0-spa --debug-port=N --origin=URL --timeout=Ns` and uses `chromedp` + `cdproto` to attach to a running Chrome (`--remote-debugging-port`), install a `Fetch.enable` interceptor, capture the first outbound `Authorization` header from a matching origin, validate the bearer via `cliutil.LooksLikeJWT` (the helper shipped in #1602, now serving its second runtime caller), and `SaveTokens(\"\", \"\", token, \"\", time.Time{})`. Falls back to the cookie-jar path when the spec doesn't declare the subtype.

`chromedp` + `cdproto` are emitted into the *printed CLI*'s `go.mod` only when the subtype is set — the Printing Press itself takes no new dep.

Honors `cliutil.IsVerifyEnv()` and `cliutil.IsDogfoodEnv()`: the CDP path short-circuits in both, per AGENTS.md side-effect-command rules. The verifier and dogfood matrix never launch a real Chrome.

## Scope

Primary area: `internal/browsersniff/` (detector), `internal/spec/` + `internal/openapi/` (Subtype + x-* extension), `internal/generator/` + `auth_browser.go.tmpl` (CDP code path).

Why this belongs in this repo: every site running Auth0 SPA SDK v2 (and growing fast — SDK v2 deprecated localStorage as XSS-unsafe in late 2023) now generates a CLI with a working auth-import path. Without this, a printed CLI for any of these sites is stuck on manual config-toml editing or the WU-3a paste escape hatch.

## Risk

- **Merge-risk with PR #1641 (WU-3a).** Both PRs add a `cliutil` import to `auth_browser.go.tmpl` under different conditions: #1641 gates on `not (eq .Auth.Type \"none\")`, this PR gates on `\$hasAuth0SPA`. The union reduces to the former (Auth0-SPA is always bearer_token, so non-none). Resolution at merge time is trivial — keep #1641's gate. Functions added by the two PRs (`newAuthSetTokenCmd` from #1641; `captureAuth0SPABearer` from this PR) sit in separate sections and don't collide.
- **chromedp + cdproto are concrete pins** (`v0.15.1` and `v0.0.0-20260321001828-e3e3800016bc`). CI builds will need network to pull them. Pseudo-version for `cdproto` verified via a scratch module locally.
- **Live CDP integration is not exercised in tests** — by design. The verifier short-circuit means we can't run a real Chrome from the test suite. The build-and-help test (`TestGenerateAuth0SPAEmitsCDPLoginCmd`) is the best automatable proxy: builds the generated CLI, runs `auth login --chrome --auth0-spa --help`, asserts the flag is wired correctly.
- **Detector path-suffix matching** on `/oauth/token` will also fire for non-Auth0 OAuth providers that surface access_token in-body without a JWT cookie. That's intentional — the CDP path works for any in-memory bearer regardless of vendor, so flagging \"Auth0 SPA-like\" is correct.

## Output Contract

Existing cookie-auth golden cases unchanged (the new code path is gated on subtype). One new generator test fixture for the Auth0-SPA emission case. `x-auth-subtype: auth0_spa_in_memory` is the spec-side wire shape; round-trips documented in `docs/SPEC-EXTENSIONS.md`.

## Verification

- `go test ./...` — green across `spec` / `openapi` / `browsersniff` / `generator` / `pipeline` / `cli`.
- `scripts/golden.sh verify` — 20/20 pass at branch base. Branch is one commit off `b5b1dd59`; no golden changes in this commit.
- `go vet ./...` — clean.
- Acceptance criteria:
  - Positive — oauth/token + access_token body + no JWT-cookie → subtype set AND CDP path emitted: `TestDetectAuth0SPAInMemory_Positive`, `TestSpecgenDetectAuth_AppliesAuth0SPASubtype`, `TestGenerateAuth0SPAEmitsCDPLoginCmd`.
  - Negative — spec without subtype: `TestGenerateBearerTokenWithoutSubtypeUsesSimpleTemplate`, `TestDetectAuth0SPAInMemory_JWTCookiePresent`, `TestDetectAuth0SPAInMemory_NoTokenCall`.

## AI / Automation Disclosure


tmchow:
CI is red on `go-lint`: https://github.com/mvanhorn/cli-printing-press/actions/runs/26075317786/job/76665026198

tmchow:
@mergifyio refresh

PR #1574, opened 2026-05-17
fix(cli): handle binary-only response endpoints (Accept + base64 envelope)

andreasvainio:
## Intent

Generated CLIs cannot download binary content. The HTTP client hardcodes
`Accept: application/json` when no per-endpoint override is set, so any endpoint
whose only 2xx content type is non-JSON (e.g. `application/octet-stream`
PDF/file downloads) is rejected server-side with HTTP 406. Even with the right
Accept, every response was run through the JSON sanitizer and returned as
`json.RawMessage`, so there was no usable binary path. This is a generic
generator gap — it affects any spec with a binary-only success response.

Issue: N/A (surfaced via a printed Tripletex CLI, where all 9 `application/octet-stream`
GET endpoints — `/document/{id}/content`, `/invoice/{invoiceId}/pdf`, etc. — returned 406).

## Approach

Three minimal changes that reuse existing plumbing:

- **openapi parser**: detect success responses whose media types are all
  non-JSON / non-`*/*` / non-`+json` and pin `Accept` to that type via the
  existing per-endpoint header-override path. New `upsertHeaderOverride` and a
  merge in `applyHeaderOverrides` keep the Accept override stable when
  configured per-endpoint headers also apply. JSON / `*/*` endpoints are never
  touched.
- **client.go.tmpl**: a content-type-gated base64 envelope
  (`{"_pp_binary":true,…,"data":"<b64>"}`) for genuinely binary success bodies
  so they survive the `json.RawMessage` contract. JSON, `*/*`, XML and every
  `text/*` type (including `text/html`, so `response_format: html` CLIs are
  unaffected) pass through unchanged. The error path is byte-identical
  (still sanitized + truncated).
- **command_promoted.go.tmpl**: promoted (top-level) GET commands now pass
  per-endpoint header overrides, matching typed commands. Without this the
  Accept override never reached the client for promoted commands.

## Scope

Primary area: openapi parser + client/promoted-command templates (generated-CLI HTTP behavior).

Why this belongs in this repo: the symptom appeared in a printed CLI, but the
defect is general Printing Press behavior — the emitted client and the parser's
response handling. Every spec with a binary-only success response is affected;
the fix is in the generator and its templates, not a printed-CLI patch.

## Risk

Generated `client.go` changes for every CLI (adds the binary envelope branch +
helpers). JSON / `*/*` / text / HTML responses are unchanged (gated on response
content type); error responses are byte-identical. No auth, catalog, publish,
[...]

PR #1662, opened 2026-05-19
fix(mcp): correct code-orchestration write-body shapes (double-marshal, query/body split, bare-array)

andreasvainio:
## Intent

Writes through the generated MCP code-orchestration surface (`<api>_execute`) sent request bodies in shapes strict-mapping APIs reject, so every POST/PUT/PATCH failed while all reads worked. Surfaced against Tripletex (HTTP 422 "Request mapping failed" / "Invalid json"); the defect is generic to the engine.

Issue: N/A (found in a printed CLI; root cause is general Printing Press behavior)

> Stacked on #1574 (still open). The first three commits in this diff (`df48d625`, `6c8634f7`, `825cdbea`) belong to #1574 and will drop out of this diff once #1574 merges. The four commits this PR adds are `5aff3949`, `a728af24`, `46d05ebc`, `42781365`.

## Approach

Three independent code-orchestration write-path defects, same emission family:

1. **Double-marshal** (`5aff3949`): the executor pre-marshaled `params` to `[]byte` and handed that to the client, which marshals once more — `json.Marshal([]byte)` yields a base64 JSON *string*. Now hands the client the structured map via `codeOrchWriteBody`.
2. **Query/body split** (`a728af24`): spec-declared `in:query` params on write methods were dumped into the JSON body. `codeOrchEndpoint.QueryParams` + `codeOrchSplitQuery` route them to the query string, reusing the existing `mcpParamBindings` location data.
3. **Bare top-level array body** (`46d05ebc`): array-root request bodies have no object properties to flatten, so nothing was emitted and the params object was sent. `mapRequestBody` now detects an array-root JSON body and sets `spec.Endpoint.BodyIsArray`; both orchestration emission paths (code-orch + intents) send the agent-supplied `params["body"]` array, with safe-degrade to the params object (no silent wrong-shape / partial write).
4. **Durable regression test** (`42781365`): the write-body contracts are now emitted from the generator (gated on `IsCodeOrchestration()`), so regens carry the guard.

## Scope

Primary area: printing-press generator (`internal/openapi` parser, `internal/spec` model, `internal/generator` templates + emission).

Why this belongs in this repo: the symptom appeared in a printed CLI, but the root cause is the generic code-orchestration/intents emission building wrong request shapes for every printed CLI. Fixing the generator fixes all of them; hand-patching a printed CLI would not survive a regen.

## Risk

Changes generated MCP code-orch + intents output. Mitigations: object write-path byte-identical (only the double-marshal indirection removed); query split scoped to write methods (GET/DELETE unchanged); array path gated on `BodyIsArray` so non-array endpoints are byte-identical; test emission gated so intents-only CLIs still compile. `scripts/golden.sh` updated (1 fixture: `generate-mcp-api/.../code_orch.go`, the verified cumulative union of the three fixes).

## Output Contract

`codeOrchEndpoint` gains `QueryParams []string` and `BodyIsArray bool`. New `codeOrchWriteBody`, `codeOrchArrayBody`, `codeOrchSplitQuery`. `intentEndpointMeta` gains `bodyIsArray bool`. New emitted file `internal/mcp/code_orch_writebody_test.go` when orchestration == code. `spec.Endpoint` gains `BodyIsArray`. Golden corpus updated accordingly.

## Verification

- `go build ./...`, `go vet ./...`, `gofmt -l` — clean
- `go test ./internal/openapi/ ./internal/spec/ ./internal/generator/` — pass
- `scripts/golden.sh` — 17/17 pass (one fixture intentionally refreshed)
- Real-spec proof: regenerated an 815-endpoint Tripletex CLI; parity preserved (codeOrchEndpoints=815, no regression); array-root booking endpoint correctly flagged `BodyIsArray`; Linux cross-compile clean

## AI / Automation Disclosure


tmchow:

Conflict footprint:
- internal/openapi/parser.go
- internal/spec/spec.go
- internal/generator/templates/ (client, command_promoted, mcp_code_orch)
- golden fixtures under testdata/golden/expected/ (regenerate after the source merge)

PR #1673, opened 2026-05-19
fix(cli): redact shipcheck api key in json

tmchow:
shipcheck `--json` no longer echoes raw `--api-key` values in each leg's `command` field. The subprocess invocation still receives the original credential, but the JSON envelope now renders both `--api-key value` and `--api-key=value` forms as `<redacted>` so logs and CI artifacts do not leak secrets.

Tested with `go test ./internal/cli` and `go test ./...`.

---

GPT--5


tmchow:
@mergifyio queue

tmchow:
@mergifyio queue default

tmchow:
@mergifyio refresh

PR #1674, opened 2026-05-19
fix(cli): create llm prompt temp files privately

tmchow:
Long LLM prompts are now written through private OS temp files instead of predictable world-readable files. This keeps prompt contents from being exposed via `llm-prompt-<timestamp>.md` names while preserving the existing Claude long-prompt handoff and cleanup behavior. The regression test installs a fake `claude` binary and verifies the prompt file mode is `600` on Unix-like platforms.

Verification:
- `go fmt ./...`
- `go test ./...`
- `go build -o ./printing-press ./cmd/printing-press`
- `golangci-lint run ./...`

---

GPT--5

### DIFF
--- a/internal/llm/llm.go
+++ b/internal/llm/llm.go
@@ -4,9 +4,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
-	"path/filepath"
 	"strings"
-	"time"
 )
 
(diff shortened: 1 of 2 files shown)


tmchow:
@mergifyio refresh

PR #1777, opened 2026-05-22
fix(cli): accept root mcp in OpenAPI parser

### LINKED ISSUE #1110: OpenAPI parser accepts root mcp: alongside canonical x-mcp:
## Summary

The main printing-press SKILL.md's "Pre-Generation MCP Enrichment" section documents the enrichment block under the `mcp:` key. That works for internal YAML specs. For OpenAPI specs, `docs/SPEC-EXTENSIONS.md` requires the **`x-mcp:`** prefix (the OpenAPI extension convention).

When an agent follows the SKILL.md example on an OpenAPI spec, the parser silently ignores `mcp:` at the root, generation proceeds with the default endpoint-mirror surface, and the agent only discovers the gap when the >50-tool warning still appears verbatim after enrichment was "added."

## Reproducing

Add this to an OpenAPI 3.0 spec root level (verbatim from the SKILL.md example):

```yaml
mcp:
  transport: [stdio, http]
  orchestration: code
  endpoint_tools: hidden
```

Run `printing-press generate --spec <spec> --output <dir>` against a >50-endpoint API. Observe the warning `spec exposes N MCP endpoint tools (>50 threshold). Consider enriching the spec's mcp: block` even though the block IS present.

Rename `mcp:` → `x-mcp:` and re-run. Warning disappears; Cloudflare pattern emitted.

## Proposed fix

In `internal/openapi/`, where root-level extensions are read into `APISpec.MCP`:

- accept `x-mcp` first (canonical)
- fall back to `mcp` and emit a one-line warning: `accepted root-level 'mcp:' for backwards compatibility — rename to 'x-mcp:' per OpenAPI extension convention`
- treat both as equivalent semantically

This widens input tolerance without changing the canonical schema. The warning gives agents a clear migration message.

Alternative (less preferred): branch the SKILL.md MCP-enrichment example on spec type — show `mcp:` for internal YAML and `x-mcp:` for OpenAPI. This is documentation-only and doesn't catch agents that work from memory or older skill copies.

## Acceptance

- positive: an OpenAPI spec with `mcp: { transport: [stdio, http], orchestration: code, endpoint_tools: hidden }` at root produces the same generated CLI as the canonical `x-mcp:` form, AND prints the migration warning
- negative: an OpenAPI spec with neither key still emits the existing "consider enriching" advisory at the >50-tool threshold

## Scope boundary

Does NOT change canonical `x-mcp:` semantics. Internal YAML's `mcp:` continues to work as today. Only input-side tolerance widens for OpenAPI sources.

## Frequency

Every OpenAPI generation where the user (or agent) follows the SKILL.md MCP-enrichment example. Catalog APIs likely to hit this on first attempt: Stripe (OpenAPI), GitHub (OpenAPI), Notion (OpenAPI), Linear (OpenAPI). Many.

## Evidence

dataforseo retro: first generate against the official OpenAPI spec with `mcp:` block at root produced the warning `warning: spec exposes 554 MCP endpoint tools (>50 threshold). Consider enriching the spec's mcp: block` despite the spec containing exactly such a block. Renaming `mcp:` → `x-mcp:` and re-running produced a clean generate with no warning.

## Component

`openapi-parser` — `internal/openapi/` root-level extensions reader

## Artifacts

- Retro document: https://files.catbox.moe/t3g2vw.md
- Manuscripts: https://files.catbox.moe/9bordi.zip
- CLI source: https://files.catbox.moe/876ayb.zip

tmchow:
## Summary

- Accept root-level `mcp:` in OpenAPI specs as a backwards-compatible fallback when canonical `x-mcp` is absent.
- Preserve canonical `x-mcp` precedence, including `info.x-mcp`, without merging fields from legacy `mcp:`.
- Document the fallback and add parser coverage for warning, precedence, and no-merge behavior.

Closes #1110

## Verification

- `go fmt ./...`
- `go test ./internal/openapi -run 'TestParse(MCPExtension|MCPFallback)' -count=1`
- `go test ./internal/spec -run MCP -count=1`
- `go test ./...`
- `golangci-lint run ./...`
- `scripts/golden.sh verify`
- `git diff --check`

### DIFF
--- a/docs/SPEC-EXTENSIONS.md
+++ b/docs/SPEC-EXTENSIONS.md
@@ -209,6 +209,10 @@ Rules:
   (including `transport: [stdio]`) bypasses the default and is honored as-is.
 - May be declared at the OpenAPI root or under `info`. Root takes precedence
   when both are present.
+- For backwards compatibility, root-level `mcp:` is also accepted when
+  canonical `x-mcp` is absent. The parser emits a warning asking authors to
+  rename it to `x-mcp`. `x-mcp` and `mcp` are never merged; any canonical
+  `x-mcp` declaration, including under `info`, wins as a complete config.
 - Shape mirrors the internal YAML `mcp:` block field-for-field: `transport`,
   `addr`, `intents`, `endpoint_tools`, `orchestration`,
   `orchestration_threshold`.
--- a/internal/openapi/parser.go
+++ b/internal/openapi/parser.go
@@ -44,6 +44,7 @@ const (
 	extensionTierRouting           = "x-tier-routing"
 	extensionTier                  = "x-tier"
 	extensionMCP                   = "x-mcp"
+	extensionLegacyMCP             = "mcp"
 	extensionSyncWalker            = "x-pp-sync-walker"
 	extensionAPIName               = "x-api-name"
 	extensionDisplayName           = "x-display-name"
@@ -422,7 +423,7 @@ func parseWithLocation(data []byte, lenient bool, location *url.URL, authPrefere
 		return nil, err
 	}
 
-	mcpConfig, err := parseTypedExtension[spec.MCPConfig](doc, extensionMCP)
+	mcpConfig, err := parseMCPExtension(doc)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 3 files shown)


tmchow:
@mergifyio queue

PR #1831, opened 2026-05-22
fix(cli): support GeoJSON FeatureCollection in sync extraction

### LINKED ISSUE #1658: generator: add `features`/`Features` to `pageItemKeys` for GeoJSON FeatureCollection sync
## Summary

`sync.go.tmpl`'s `pageItemKeys` list doesn't include `features`, so GeoJSON `FeatureCollection` responses get stored as a single wrapper row instead of splitting into N feature rows. Every printed CLI that targets a geo / hazard / mapping API hits this silently.

## Symptom

During USGS Earthquakes generation, `sync --resources events --param "format=geojson"` reported `{"event":"sync_complete","resource":"events","total":1,…}` and `SELECT COUNT(*) FROM resources` returned `1` — with the entire FeatureCollection wrapped object stored as a single JSON blob. The 200 actual earthquake features were unreachable to `search`, `sql`, and any local-store query.

After hand-patching `internal/cli/sync.go` to add `"features"` and `"Features"` to `pageItemKeys`, the next sync correctly returned `total":200` with 200 individual feature rows.

## Root cause

`internal/generator/templates/sync.go.tmpl` (around line 1568):

```go
var pageItemKeys = []string{
    "data", "results", "items", "records", "nodes", "entries",
    "Data", "Results", "Items", "Records", "Nodes", "Entries",
}
```

`extractPageItems` walks this list to find the array-of-objects key inside a wrapper response. The list covers REST + .NET PascalCase conventions but omits the canonical GeoJSON wrapper key `features` / `Features`.

The spec's `response_path: features` field is honored by *read* commands but not by sync's `extractPageItems` call path.

## Cross-API evidence

This pattern recurs across every geo / hazard / mapping API:

1. **USGS Earthquakes** (this run) — `/fdsnws/event/1/query?format=geojson` returns `FeatureCollection` with `features[]`. Documented at https://earthquake.usgs.gov/fdsnws/event/1/
2. **NOAA Weather API** — `/gridpoints/{office}/{x},{y}/forecast` and `/alerts/active` both return `FeatureCollection` with `features[]`. Documented at https://www.weather.gov/documentation/services-web-api
3. **OpenStreetMap Nominatim / Mapbox / ArcGIS Feature Service** — every GeoJSON-conformant API returns the same OGC-standard `FeatureCollection` shape.

## Proposed fix

Append `"features"` to the lowercase row and `"Features"` to the PascalCase row of `pageItemKeys` in `internal/generator/templates/sync.go.tmpl`. The existing `extractPageItems` already verifies the value is an array of objects, so a non-geo API with a scalar `features` field (boolean / string / single object) won't false-match — it falls through to the existing ambiguity-scan path.

## Tests

- **Positive:** Generate a CLI from a spec whose list-endpoint response is `{"type":"FeatureCollection","features":[{...},{...}]}`; assert sync stores N rows in the `resources` table, one per feature.
- **Negative:** Generate from a spec whose response is `{"features":true}` or `{"features":"flag-list"}`; assert the extractor does NOT false-positive — falls through to the existing ambiguity-scan path.

## Workaround until fixed

Hand-patch `internal/cli/sync.go` in every regen of any geo API:

```go
var pageItemKeys = []string{
    "data", "results", "items", "records", "nodes", "entries", "features",
    "Data", "Results", "Items", "Records", "Nodes", "Entries", "Features",
}
```

The patch is overwritten on the next `generate`, so this lands as a permanent printed-CLI fix.

## Related issues

- #1234 — `generator: workflow archive --json interleaves NDJSON sync-events with a JSON doc` — `extends`. Same template (`sync.go.tmpl`), different gap — wrapper-key detection vs streaming-output handling.
- #1421 — `Sync template doesn't auto-inject spec-declared field-selector parameters` — `related-area`. Both findings are about sync's relationship to spec-declared content the template doesn't respect.

## Artifacts

This finding emerged from the `/printing-press usgs-earthquakes` run on 2026-05-19. Full retro at `manuscripts/usgs-earthquakes/20260519-002226/proofs/20260519-015355-retro-usgs-earthquakes-pp-cli.md` (will be uploaded to catbox.moe via the standard retro artifact path).

## Acceptance criteria

- [ ] `internal/generator/templates/sync.go.tmpl`: `pageItemKeys` includes `"features"` and `"Features"`.
- [ ] Positive regression: generate a CLI from a fixture spec returning `{"features":[...]}`, assert sync stores N rows.
- [ ] Negative regression: generate from a fixture returning `{"features":true}`, assert the extractor falls through cleanly.
- [ ] Existing template tests pass (no regression on REST/PascalCase envelopes).

SATYAM-PRATIBHAN:
## Intent

This PR adds support for GeoJSON `FeatureCollection` responses in the sync extraction logic. Previously, GeoJSON APIs (like USGS Earthquakes or NOAA Weather) would often be stored as a single JSON blob instead of individual feature rows because the extraction logic didn't recognize the `"features"` key and was confused by sibling arrays like `"bbox"`.

Issue: #1658 

Fixes:  #1658 

## Approach

I added `"features"` and `"Features"` to the `pageItemKeys` priority list in the sync template. This allows the extractor's "fast path" to find the correct items array even when the response contains other top-level arrays (like GeoJSON's standard `bbox`), which previously caused the "ambiguity scan" fallback to fail.

## Scope

Primary area: `internal/generator/templates/sync.go.tmpl`

Why this belongs in this repo: This is a systemic fix for the sync template (the "machine"). Every printed CLI targeting a GeoJSON-conformant API was silently hitting this extraction failure, resulting in unsearchable single-row resource tables.

## Risk

N/A

## Output Contract

This change modifies `internal/generator/templates/sync.go.tmpl`, which affects the generated `internal/cli/sync.go` file in every printed CLI. The `pageItemKeys` list in the generated code now includes `"features"` and `"Features"`.

## Verification

The following commands were run to verify the fix and ensure no regressions:

1. **Unit Tests**: `go test -v ./internal/generator -run TestSync`
   - Verified that the new Tier 1 emission canaries pass.
   - Verified that the new Tier 2 behavioral tests (`TestExtractPageItemsGeoJSON`, `TestExtractPageItemsGeoJSONPascal`, and `TestExtractPageItemsNonArrayFeaturesFallback`) pass.
2. **Golden Tests**: `bash scripts/golden.sh verify`
   - Confirmed that the intentional drift in 3 generated `sync.go` fixtures is stable and matches the expected template change.

## AI / Automation Disclosure

### DIFF
--- a/internal/generator/generator_test.go
+++ b/internal/generator/generator_test.go
@@ -3186,6 +3186,10 @@ func TestSyncExtractPaginationNestedCursor(t *testing.T) {
 		"paginationWrapperKeys must include response_metadata (Slack's envelope)")
 	assert.Contains(t, src, `"pagination"`,
 		"paginationWrapperKeys must include pagination (MongoDB Atlas-style)")
+	assert.Contains(t, src, `"features"`,
+		"pageItemKeys must include features (GeoJSON support)")
+	assert.Contains(t, src, `"Features"`,
+		"pageItemKeys must include Features (PascalCase GeoJSON support)")
 
 	// Tier 2: behavioral test, written into the generated tree as a
 	// same-package _test.go so it can call the unexported helper.
@@ -3195,6 +3199,43 @@ import (
 	"encoding/json"
 	"testing"
 )
+
+func TestExtractPageItemsGeoJSON(t *testing.T) {
+	body := []byte(` + "`" + `{
+		"type": "FeatureCollection",
+		"bbox": [-180, -90, 180, 90],
+		"features": [{"id":"1"},{"id":"2"}]
+	}` + "`" + `)
+	items, _, _ := extractPageItems(json.RawMessage(body), "cursor")
+	if len(items) != 2 {
+		t.Fatalf("GeoJSON: want 2 items, got %d", len(items))
+	}
+}
+
+func TestExtractPageItemsGeoJSONPascal(t *testing.T) {
+	body := []byte(` + "`" + `{
+		"Type": "FeatureCollection",
+		"Bbox": [-180, -90, 180, 90],
+		"Features": [{"Id":"1"},{"Id":"2"}]
+	}` + "`" + `)
+	items, _, _ := extractPageItems(json.RawMessage(body), "cursor")
+	if len(items) != 2 {
+		t.Fatalf("GeoJSON Pascal: want 2 items, got %d", len(items))
+	}
+}
+
+func TestExtractPageItemsNonArrayFeaturesFallback(t *testing.T) {
+	// features: true should fall through to ambiguity scan or return nil.
+	// In this case, there are NO arrays, so it returns nil.
+	body := []byte(` + "`" + `{
+		"features": true,
+		"other": "val"
+	}` + "`" + `)
+	items, _, _ := extractPageItems(json.RawMessage(body), "cursor")
+	if len(items) != 0 {
+		t.Fatalf("Non-array features: want 0 items, got %d", len(items))
+	}
+}
 
 func TestExtractPageItemsSlackEnvelope(t *testing.T) {
 	body := []byte(` + "`" + `{
--- a/internal/generator/generator_test.go
+++ b/internal/generator/generator_test.go
@@ -3237,6 +3237,25 @@ func TestExtractPageItemsNonArrayFeaturesFallback(t *testing.T) {
 	}
 }
 
+func TestExtractPageItemsFeaturesPrecedenceConflict(t *testing.T) {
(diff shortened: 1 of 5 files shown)


SATYAM-PRATIBHAN:
hey @tmchow , could you take a look at this PR?

tmchow:
> hey @tmchow , could you take a look at this PR?

You don't need to ping me to review PRs. They will be triaged.

SATYAM-PRATIBHAN:
> > hey @tmchow , could you take a look at this PR?
> 
> You don't need to ping me to review PRs. They will be triaged.

got it!

Issue #1642, opened 2026-05-19
generator: html_extract has no mode for schema.org JSON-LD; SSR content sites need hand-written extractors


bobeglz:
> Filed by `/printing-press-retro` from a run of **Rappi (Mexico)** (scorecard 81/100, 100% verify pass).

## Summary

The generator's `html_extract` template offers three modes: `page`, `links`, and `embedded-json` (Next.js `__NEXT_DATA__` blob). It has no mode for schema.org JSON-LD blocks (`<script type="application/ld+json">`), which are the most common structured-data carrier on SSR content sites (food/recipe/product/news). When the generator targets a site like Rappi that embeds typed `Restaurant`, `ItemList`, `OpeningHoursSpecification`, `AggregateRating`, etc., the existing modes either drop the structured data (links: returns noisy anchor text) or skip it entirely (embedded-json: looks for `__NEXT_DATA__`, not application/ld+json). The agent has to hand-write a JSON-LD parser at `internal/source/<api>/jsonld.go` (~300 lines on Rappi) for every such CLI.

## Where to look

- **Component:** generator
- **Likely area:** `internal/spec/spec.go` (HTMLExtract struct + Validate), `internal/generator/templates/html_extract.go.tmpl`
- **Triggered when:** spec has `response_format: html` against a site whose SSR pages embed schema.org JSON-LD blocks (most "no public API" SSR catalog sites)

## What we observed

Rappi's restaurant detail pages render structured `Restaurant` data via JSON-LD:

```html
<script type="application/ld+json">
{
  "@type": "Restaurant",
  "@id": "https://www.rappi.com.mx/restaurantes/10000295-el-farolito",
  "name": "El Farolito",
  "servesCuisine": ["Tacos", "Mexicana"],
  "address": {"@type": "PostalAddress", "streetAddress": "..."},
  "geo": {"@type": "GeoCoordinates", "latitude": 19.406346, "longitude": -99.172858},
  "aggregateRating": {"ratingCount": 64, "ratingValue": 4.1, "bestRating": 5},
  "openingHoursSpecification": [...]
}
</script>
```

The existing modes can't extract this:
- `mode: links` returns a noisy mix of nav anchors, promo text, and detail links — verifiable on the rappi-pp-cli output for `restaurants list-city --json` before novel commands took over.
- `mode: page` extracts `<title>`, `<meta>`, and `<a>` links from the document — doesn't read JSON-LD.
[...]

PR #1864, opened 2026-05-23
fix(generator): support embedded-json selector shapes

ItsOtherMauridian:
## Summary

Addresses mvanhorn/cli-printing-press#1642 by making the generated `embedded-json` extractor understand the selector shapes that browser-sniff reachability defaults already emit for SSR state blobs.

Specifically, the generated runtime now supports this limited selector grammar:

- `tag`
- `tag#id`
- `tag.class`
- `tag[attr="value"]`

This covers schema.org JSON-LD selectors such as `script[type="application/ld+json"]` and existing `script.state-view` defaults without introducing a broad CSS selector engine.

## Test-first behavior

Before the implementation change, the new JSON-LD generated-runtime case failed with:

```text
Error: embedded-json: no element matching selector "script[type=\"application/ld+json\"]" found in page
```

After the implementation change, the generated CLI extracts both:

- schema.org JSON-LD via `script[type="application/ld+json"]`
- state-view JSON via `script.state-view`

## Test plan

Ran locally:

```bash
go test ./internal/spec ./internal/browsersniff -count=1
go test ./internal/generator -run 'TestGenerateHTMLExtraction(EmbeddedJSONMode|PerModeGating)' -count=1
```

Result:

```text
ok  github.com/mvanhorn/cli-printing-press/v4/internal/spec          0.093s
ok  github.com/mvanhorn/cli-printing-press/v4/internal/browsersniff  0.631s
ok  github.com/mvanhorn/cli-printing-press/v4/internal/generator     9.489s
```

I also ran the focused embedded-json test alone:

```text
ok  github.com/mvanhorn/cli-printing-press/v4/internal/generator  3.205s
```

Note: I attempted the full `go test ./internal/generator -count=1`, but stopped that local run after it produced no output for several minutes. No failure output was produced from that interrupted run.

## Transparency

This PR was prepared with AI assistance. I reviewed the generated changes and ran the local tests listed above.

### DIFF
--- a/internal/generator/generator_test.go
+++ b/internal/generator/generator_test.go
@@ -2477,6 +2477,19 @@ func TestGenerateHTMLExtractionEmbeddedJSONMode(t *testing.T) {
 			_, _ = w.Write([]byte(`<html><body>
 				<script id="ARTICLE_DATA" type="application/json">{"items":[{"slug":"a"},{"slug":"b"}]}</script>
 			</body></html>`))
+		case "/jsonld":
+			// Schema.org JSON-LD is emitted by browser-sniff reachability defaults as
+			// script[type="application/ld+json"]. The generated runtime must understand
+			// that attribute-selector shape, not only tag#id.
+			_, _ = w.Write([]byte(`<html><body>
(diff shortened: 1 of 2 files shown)


ItsOtherMauridian:
Closing this in favor of #1877, which covers the same #1642 selector-runtime fix and also rejects unsupported attribute-selector shapes explicitly. Thanks.

PR #1906, opened 2026-05-23
fix(cli): distinguish verify-skill prose invocations

### LINKED ISSUE #1556: scorer: verify-skill command detection flags natural-English prose mentions of CLI name
> Filed by `/printing-press-retro` from a run of **exchangerate-api** (scorecard 81/100, verify 90.9%).

## Summary

verify-skill's command-detection scanner flags natural-English prose as unknown commands. In this run it reported:

```
[unknown-command] exchangerate-api-pp-cli wraps: command path not found in internal/cli/*.go (no matching Use: declaration)
  evidence: bash recipe (SKILL.md)
```

The matched source line is plain prose, not a command invocation:

> exchangerate-api-pp-cli wraps the official ExchangeRate-API with typed commands for every endpoint…

The scanner greps for `<cli-name> <next-token>` patterns and treats every match as a candidate command, regardless of markdown context (code fence vs. inline code vs. plain paragraph) or whether the next token has a command shape (subcommand / `--flag`) vs. an English verb.

## Where it lives

- **Component:** scorer
- **Likely area:** verify-skill's command-detection regex/heuristic in `internal/cli/verify_skill_bundled.py` (or equivalent), parallel to `FLAG_DECL_RE` from #1255 and the recipe-parser fix from #1316.

## What we observed

Workaround for this CLI: rewrite `narrative.value_prop` to NOT start with the CLI name + verb. Original:

```text
exchangerate-api-pp-cli wraps the official ExchangeRate-API with typed commands…
```

Reworded to:

```text
This CLI exposes every ExchangeRate-API endpoint (latest, pair, codes, quota, …)
```

The reworded form clears verify-skill, but the original is more natural prose. Future CLIs will face the same friction every time the agent writes "<cli-name> wraps/handles/provides/queries/exposes/connects/manages…" in narrative or README intro paragraphs.

## Cross-CLI evidence (Step B)

Every CLI has narrative.value_prop or README intro prose where it's natural to write "<cli-name> <verb>". Direct evidence is limited to this run (one occurrence I caught and worked around), but the verb-following-CLI-name pattern is ubiquitous in English exposition.

- **exchangerate-api** (this run): direct evidence, flagged "wraps".
- **Architectural:** every Printing Press CLI ships a SKILL.md and README.md authored by an agent; that prose naturally opens with the CLI name. The longer the catalog grows, the more these false positives accumulate.

The same shape of scanner bug (regex misses legitimate vs flags illegitimate patterns) already has two siblings open: #1255 (flag-declaration regex variants) and #1316 (recipe parser keeps =value suffixes).

## Durable fix

Make verify-skill's command detection context-aware. Require ONE of these to fire a warning:

1. The match is inside a fenced code block (```` ``` ```` or `~~~`)
2. The match is inside an inline code span (backticks)
3. The next token is a known shape that looks like a CLI invocation: starts with `--` (a flag), OR is a subcommand the CLI declares (would resolve in the cobra tree), OR is followed by a known flag pattern

Any `<cli-name> <word>` outside code context where `<word>` isn't a real subcommand should be treated as prose and dropped (or warned at a much lower severity level — `info`, not `error`).

This is a strict improvement: it does not weaken wrong-command detection for the real cases (which all live inside code fences in this run's narrative). The fix only suppresses prose context.

## Acceptance criteria

- **positive 1:** SKILL.md with `cli-name latest USD` inside a fenced bash block where `latest` is a real subcommand → no warning.
- **positive 2:** SKILL.md with `` `cli-name latest USD` `` in an inline code span → no warning.
- **positive 3:** SKILL.md with `cli-name wraps the API and provides...` in a plain paragraph → no warning (currently fires).
- **negative 1:** SKILL.md with a real wrong-command invocation inside a code fence (e.g., `cli-name fakecommand --foo` where `fakecommand` is not declared) → WARN (preserve existing behavior).
- **negative 2:** SKILL.md with `cli-name sync --base USD` in plain prose where `--base` is not declared on `sync` → WARN (the real false-positive case caught in this run: gibberish flag in prose-shape).

## Scope boundary

- Only the prose-vs-code disambiguation.
- Does NOT change which commands or flags are considered "known".
- Does NOT relax wrong-command detection inside actual code context.

## Related issues

- #1255 — adjacent verify-skill scanner bug (flag-declaration regex variants).
- #1316 — adjacent verify-skill scanner bug (recipe parser keeps `=value` suffix on tokens).

## Artifacts

Full retro: `manuscripts/exchangerate-api/20260517-025552/proofs/20260517-040800-retro-exchangerate-api-pp-cli.md` (local).

tmchow:
## Intent

Closes #1556.

`verify-skill` should ignore natural prose like `<cli> wraps the API...` while still catching prose that is shaped like a real CLI invocation with flags.

## Approach

The verifier now has a shared invocation extractor for fenced bash recipes and plain prose mentions that include long flags. Plain prose is stripped of markdown code spans before scanning, and multiple CLI mentions on one line are split so flags stay attributed to the right command.

Bash recipe extraction remains the only source for positional argument validation. Flag checks and unknown-command checks consume the broader invocation set.

## Scope

Primary area: verifier and scorer support code.

Why this belongs in this repo: the false positive comes from the Printing Press `verify-skill` scanner, not from a single printed CLI.

## Risk

The risk is verifier classification drift. Regression tests cover natural prose, invocation-shaped prose flags, positional-arg separation, multiple CLI mentions on one line, and script-to-bundled sync.

## Output Contract

N/A

## Verification

- `python3 -m py_compile scripts/verify-skill/verify_skill.py internal/cli/verify_skill_bundled.py`
- `go test ./internal/cli -run 'TestVerifySkill_(UnknownCommandIgnoresNaturalLanguageCliProse|ProseInvocationWithFlagIsChecked|ProseInvocationDoesNotAffectPositionalArgs|ProseInvocationSplitsMultipleCliMentions)' -count=1`
- `go test ./internal/cli -run 'TestVerifySkill_(UnknownCommandIgnoresNaturalLanguageCliProse|ProseInvocationWithFlagIsChecked|ProseInvocationDoesNotAffectPositionalArgs|ProseInvocationSplitsMultipleCliMentions|ScriptInSync)' -count=1`
- `go test ./internal/cli -count=1`
- `go test ./internal/cli -run TestVerifySkillScriptInSync -count=1`
- `go fmt ./...`
- `go test ./...`
- `golangci-lint run ./...`

### DIFF
--- a/internal/cli/verify_skill_bundled.py
+++ b/internal/cli/verify_skill_bundled.py
@@ -68,6 +68,8 @@ def read_utf8(path: Path) -> str:
 COMMON_FLAGS = {"help", "version"}
 
 CODEBLOCK_BASH = re.compile(r"```bash\n(.*?)\n```", re.DOTALL)
+FENCED_CODE = re.compile(r"```.*?```|~~~.*?~~~", re.DOTALL)
+INLINE_CODE = re.compile(r"`[^`\n]*`")
 COMMAND_REFERENCE_SECTION_RE = re.compile(
     r"^##\s+Command\s+Reference\s*$(.*?)(?=^##\s+|\Z)",
     re.DOTALL | re.MULTILINE | re.IGNORECASE,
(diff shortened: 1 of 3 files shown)


tmchow:
@mergifyio queue

PR #1925, opened 2026-05-23
docs(skills): add cache enrichment guidance

### LINKED ISSUE #1558: skill: Pre-Generation Cache Enrichment block missing; agents leave cache.enabled unset and ship without auto-refresh
> Filed by `/printing-press-retro` from a scorecard-walk follow-up on
> **Blu-ray.com**. cache_freshness scored 5/10 on the original run because
> the agent didn't know to add a `cache:` block to the spec — the SKILL
> doesn't surface that the framework already has the templates ready to
> emit when opted in.

## Summary

The framework ships two production-ready templates that score the user a
+5 cache_freshness bonus AND give the printed CLI a genuinely useful
auto-refresh behavior (re-sync the local catalog before serving a stale
read): `cliutil_freshness.go.tmpl` and `auto_refresh.go.tmpl` under
`internal/generator/templates/`. The generator emits both — but only
when `spec.Cache.Enabled == true`. The SKILL doesn't tell agents this.
Most agents (Claude, Codex) leave `Cache.Enabled` unset and the templates
silently don't emit. The CLI scores 5/10 on the cache_freshness dim AND
ships without a useful feature, despite the framework having shipped the
exact code to provide it.

## Where to look

- **Component:** skill (primary fix); generator (secondary if changing the default)
- **Likely area:**
  - `skills/printing-press/SKILL.md` Phase 2 — the spec-author guidance
    blocks (Pre-Generation Auth Enrichment, Pre-Generation MCP Enrichment,
    etc.). A new sibling block "Pre-Generation Cache Enrichment" would
    surface this.
  - The CacheConfig struct in `internal/spec/spec.go` for the schema reference.
  - Emit gate at `internal/generator/generator.go:1537` (`g.VisionSet.Store && (g.Spec.Cache.Enabled || g.Spec.Share.Enabled)`).
- **Triggered when:** any CLI with `VisionSet.Store == true` (i.e., the
  CLI has a local store at all) and `spec.Cache.Enabled` unset. Per the
  scoreCacheFreshness rubric (`scorecard.go:825`), the missing
  `autoRefreshIfStale` + `EnsureFresh` symbols cost 5/10 on the dim.

## What we observed

On the blu-ray run, the absorb manifest required a local catalog (sitemap
sync → SQLite FTS5), so `VisionSet.Store == true` triggered the framework's
store scaffolding. But the spec didn't carry a `cache:` block, so the
generator's emit gate skipped `cliutil_freshness.go` and `auto_refresh.go`.
Result:

- `scoreCacheFreshness` returns 5 (got 3 for schema-version gate, 2 for
  doctor cache section, 0 for the missing auto-refresh wiring).
- No auto-refresh on commands: `search`, `editions`, `drift`, `history`
  all happily serve stale local data without checking sync_state. A user
  who hadn't run `sync` in a month would see month-old data with no
  warning.

To fix on the CLI side: we added a `cache:` block to spec.yaml and
side-channel-rendered the two missing files into the library (without
running the full regen, to avoid re-emitting the dead generic sync.go).
The dim jumped from 5 → 10, and the CLI now auto-refreshes if the local
catalog is older than the configured threshold.

But the SKILL never told the agent to do this. Codex/Claude/any agent
writing a spec for a stateful CLI today is going to leave `Cache.Enabled`
unset by default, and ship a CLI that scores 5/10 on the dim AND lacks
auto-refresh.

## Suspected root cause

This is a SKILL gap, not a generator bug. The generator's gate is
deliberate (small/stateless CLIs don't need auto-refresh; emitting the
helpers blindly would be dead code there). But the SKILL's spec-authoring
phases (Auth Enrichment, MCP Enrichment, public-param Enrichment) don't
include a Cache Enrichment block telling the agent: "If your absorb
manifest declares a local store / catalog / sync resource, add
`cache: { enabled: true }` to the spec."

Confidence: high. The mechanism is well-understood; the only question is
whether the right fix is a SKILL addition, a generator default flip
(emit when VisionSet.Store == true and let the agent opt OUT), or both.

## Suggested direction

Two paths; (a) is the SKILL fix, (b) is the generator-default flip. Pick (a)
as the durable fix; (b) is a stronger option but a behavior change for
existing CLIs.

(a) **Add a "Pre-Generation Cache Enrichment" block to SKILL.md Phase 2.**
Mirror the existing "Pre-Generation Auth Enrichment" / "Pre-Generation
MCP Enrichment" / "Pre-Generation Public Parameter Name Enrichment"
shape. Concrete content:

> **Pre-Generation Cache Enrichment**: If the absorb manifest declares any
> local store / catalog / sync resource, add a `cache:` block to the spec
> BEFORE generation:
> ```yaml
> cache:
>   enabled: true
>   stale_after: 168h   # or domain-appropriate, e.g. 1h for news-feed APIs
>   commands:
>     - name: <novel-read-command>
>       resources: [<resource-name>]
> ```
> This emits `internal/cliutil/freshness.go` + `internal/cli/auto_refresh.go`
> which auto-refresh stale local data before serving read commands. Without
> this block, the cache_freshness dim caps at 5/10 and read commands serve
> stale data without warning.

(b) **(Stronger but breaking) Flip the generator default.** Emit the
auto-refresh helpers whenever `VisionSet.Store == true`, regardless of
`Cache.Enabled`. The `Cache` struct's defaults would supply the stale-after
threshold (24h or similar). Existing CLIs that explicitly want NO
auto-refresh add `cache: { enabled: false }`.

Path (a) is one SKILL edit; path (b) is a one-line generator change plus
schema-default plumbing. I'd recommend (a) unless the maintainers want to
make stateful-CLI agents' lives easier across the board.

## Acceptance criteria

- positive: a fresh CLI run where the absorb manifest declares a local
  catalog produces a spec.yaml with a `cache:` block, the generator emits
  the two files, and `cache_freshness` scores 10/10.
- negative: a CLI with no local store (read-through wrappers around a
  pure-stateless API) still skips the emit (no dead helpers shipping where
  they have nothing to refresh).

## Frequency

every API with a local store / sync resource. Concrete examples from the
library: blu-ray (sitemap sync, FTS5), pcgs (per-cert lookup cache, hydrated
on demand), and presumably most CLIs the press has emitted since the
auto-refresh templates landed. Conservative count: any CLI in
`~/printing-press/library/` whose spec sets `vision.store: true` (the
default for catalog-shaped APIs) is missing this enrichment.

## Complexity

small. One SKILL block addition.

## Scope boundary

Does NOT change the CacheConfig schema. Does NOT auto-enable cache for
all CLIs (that's the optional path b). Does NOT touch the rubric inside
`scoreCacheFreshness` (the rubric is correct; the SKILL just doesn't
surface how to satisfy it).

## Dependencies

None.

## Related issues

- #1131 — `[P3] Cache freshness helper not emitted; scorer audit needed before generator fix` — `related-area`. Adjacent to this finding — they both observe that the cache-freshness helpers don't reach printed CLIs. #1131 questions the rubric; this issue says the rubric is correct, the SKILL just needs to tell agents the satisfying input.

## Artifacts

| Artifact | Link |
|----------|------|
| Retro document | https://files.catbox.moe/5updn7.md |
| Manuscripts (research + proofs) | https://files.catbox.moe/i7psqi.zip |
| Generated CLI source code | https://files.catbox.moe/rffoxy.zip |

---

tmchow:
## Summary

Stateful generated CLIs now get explicit Printing Press skill guidance for opting into cache freshness before generation. Agents are told to add `cache.enabled` only when a resolved sync-backed read path can refresh from upstream, and to avoid auto-refresh for stateless wrappers, per-user working state, and expensive or quota-sensitive refresh paths.

The guidance also separates the base cache config from optional `cache.commands` coverage so hand-authored store-reading commands do not look mandatory for generated resource commands.

Closes #1558

## Verification

- `go test ./...`
- `golangci-lint run ./...`
- `git diff --check origin/main`

### DIFF
--- a/skills/printing-press/SKILL.md
+++ b/skills/printing-press/SKILL.md
@@ -1812,6 +1812,10 @@ refresh APIs unless the refresh path is cheap, bounded, and clearly valuable;
 those CLIs should rely on manual `sync` plus the generated `doctor` cache report
 instead of surprising users with pre-read upstream calls.
 
+Catalog-mode runs skip this step: keep the built-in catalog entry's cache
+settings unchanged. Do not pass a flag or patch generated files after the fact;
+cache freshness must come from the spec that drives generation.
+
 For internal YAML specs, add the cache block before the final `generate`
 invocation only when at least one generated syncable resource read command will
 be covered automatically, or `cache.commands` will register a real hand-authored
@@ -1833,6 +1837,7 @@ resource must be declared in `resources:` and classified as syncable.
 ```yaml
 cache:
   enabled: true
+  stale_after: 168h        # choose a domain-appropriate default
   commands:
     - name: <novel-read-command>
       resources: [<resource-name>]
--- a/skills/printing-press/SKILL.md
+++ b/skills/printing-press/SKILL.md
@@ -1791,6 +1791,59 @@ spec, or `verify-skill canonical-sections` can drift.
 Catalog-mode runs skip this step: keep the built-in catalog entry's category
 unchanged, even if Phase 1 research would classify the API differently.
 
+### Pre-Generation Cache Enrichment
+
+Before generating, decide whether the spec should opt into generator-owned cache
+freshness. The generator already has the freshness helpers and auto-refresh hook,
+but it emits them only when the spec declares `cache.enabled: true` and the CLI
+has a real sync path. Stateful catalog-shaped CLIs otherwise serve local data
+exactly as it was last synced, which caps the cache freshness score and can leave
+agents reading stale SQLite rows without a warning.
+
+Enable cache freshness only when the resolved spec, profiler output, or absorb
+manifest shows at least one covered read path backed by a syncable resource that
+`sync` can refresh from the upstream API before serving. Do not enable it from
+Phase 1 research notes or scorecard goals alone. Leave it disabled for stateless
+read-through wrappers and for local stores that are primarily per-user working
+state, such as carts, drafts, or other session-owned data where a pre-read
+refresh could replace the user's local state with a different snapshot. Also
+leave it disabled for quota-metered, paid, rate-limited, or expensive bulk
+refresh APIs unless the refresh path is cheap, bounded, and clearly valuable;
+those CLIs should rely on manual `sync` plus the generated `doctor` cache report
+instead of surprising users with pre-read upstream calls.
+
+For internal YAML specs, add the cache block before the final `generate`
+invocation only when at least one generated syncable resource read command will
+be covered automatically, or `cache.commands` will register a real hand-authored
+store-reading command:
+
+```yaml
+cache:
+  enabled: true
+  stale_after: 168h        # choose a domain-appropriate default
+  refresh_timeout: 30s     # optional; blank uses the generated runtime default
+```
+
+Generated resource list/get/search commands are covered automatically from the
+syncable resources profile. Use `cache.commands` only for hand-authored novel
+commands that read the local store and are not generated resource commands. The
+command `name` is the Cobra path without the binary name, and every listed
+resource must be declared in `resources:` and classified as syncable.
+
+```yaml
+cache:
+  enabled: true
+  commands:
+    - name: <novel-read-command>
+      resources: [<resource-name>]
+```
+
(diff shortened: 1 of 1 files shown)


tmchow:
@mergifyio queue

PR #1927, opened 2026-05-23
fix(cli): align MCP tool design scoring threshold

### LINKED ISSUE #1560: scorer: mcp_tool_design threshold (10 endpoints) contradicts SKILL Pre-Generation MCP Enrichment (<30 = endpoint-mirror fine); 20-tool silent-penalty window
> Filed by `/printing-press-retro` from a run of **exchangerate-api** (scorecard 81/100, Grade A — `mcp_tool_design 5/10` was one of the dimensions capping the total).

## Summary

`scoreMCPToolDesign` (`internal/pipeline/mcp_scoring.go:144`) engages and penalizes endpoint-mirror surfaces at **≥10 endpoint tools**:

```go
const toolDesignMinEndpoints = 10
...
if s.endpointTools < toolDesignMinEndpoints {
    return 0, false  // unscored
}
...
return 5, true  // endpoint-mirror only
```

The SKILL's **Pre-Generation MCP Enrichment** section (`SKILL.md`) explicitly says the opposite for the 10–29 endpoint range:

| Total tools | SKILL action |
|-------------|--------------|
| **<30** | **Skip — default endpoint-mirror surface is fine.** |
| 30–50 | Ask the user; suggest `mcp.transport: [stdio, http]` and possibly `mcp.intents`. |
| >50 | Default to recommending the Cloudflare pattern (transport + code orchestration + hidden endpoint tools). |

This produces a **20-tool window (10–29)** where the SKILL tells the agent "endpoint mirror is fine; don't enrich" and the scorer awards 5/10 anyway. Every CLI in that window loses 5pt on `mcp_tool_design` despite following the SKILL's explicit guidance.

## Direct evidence

**exchangerate-api** (this run): 10 `mcplib.NewTool(` registrations in `internal/mcp/tools.go` (7 spec-derived + 3 framework). SKILL Pre-Generation MCP Enrichment said "skip enrichment for <30 tools" — followed correctly. Scorer awarded 5/10.

## Cross-CLI Step B

Any 10–29 endpoint CLI in the catalog hits this. Concrete catalog evidence:

- **exchangerate-api** (this run, 10 tools): 5/10
- **pcgs** (likely 10+ tools given its endpoint shape; needs verification): 5/10
- **Most small-to-medium APIs in the published library** (catalog has many APIs in the 10-29 endpoint range)

The catalog's larger APIs (>50 endpoints) are correctly enriched per the SKILL; the small APIs are correctly skipped per the SKILL. **The 10–29 range is where SKILL and scorer disagree, and the SKILL is the document agents follow.**

## Two paths (one issue, two valid fix surfaces)

**Path A (preferred): scorer threshold change.** Raise `toolDesignMinEndpoints` from 10 to 30, aligning with the SKILL. The 10–29 range becomes unscored (`(0, false)`), removing the 5-point silent penalty for following SKILL guidance.

**Path B (alternative): SKILL change.** Lower the SKILL's "default endpoint-mirror surface is fine" threshold from 30 to 10, telling agents to start adding intents at 10 tools. This is a much bigger ask — it commits agents to writing intent tools (hand-authored Go) for every modest API, plus mandates spec edits.

Path A is the right call: code orchestration / intents are valuable patterns but they cost real implementation work, and the 5-point silent penalty doesn't justify mandating that work on tiny surfaces. The SKILL's guidance is sound; the scorer threshold should match.

## Counter-check (Step C)

Raising `toolDesignMinEndpoints` to 30 doesn't hurt larger APIs — they still get scored. Doesn't hurt smaller APIs — they correctly stay unscored. The only "loss" is the scorer no longer pressures 10-29 endpoint CLIs to add intent groupings, which the SKILL has already declared not-worth-the-cost at that scale.

## Acceptance criteria

- **positive 1:** Generate a CLI with 15 endpoint tools (no intents.go, no code_orch.go). `scoreMCPToolDesign` returns `(0, false)` — dimension is omitted from the denominator. No silent 5pt penalty.
- **positive 2:** Generate a CLI with 35 endpoint tools (no intents). `scoreMCPToolDesign` returns 5/10 (current behavior preserved for the range where SKILL also recommends action).
- **positive 3:** Generate a CLI with 15 endpoint tools + intents.go covering ≥30% of the surface. `scoreMCPToolDesign` returns 10/10 (still rewards correct adoption when present).
- **negative:** Generate a CLI with 60 endpoint tools (no intents). `scoreMCPToolDesign` returns 5/10 (anti-pattern at scale still penalized — `scoreMCPSurfaceStrategy` does the heavy lifting here too).

## Scope boundary

Only the `toolDesignMinEndpoints` constant. Does not change `surfaceStrategyLargeThreshold` (= 50) which correctly penalizes the very-large endpoint-mirror anti-pattern.

## Related issues

None directly. The SKILL Pre-Generation MCP Enrichment section is recent (post-most-retros); this is the first time the threshold-mismatch shape is being raised.

## Artifacts

Full retro: `manuscripts/exchangerate-api/20260517-025552/proofs/20260517-040800-retro-exchangerate-api-pp-cli.md` (local).

tmchow:
## Summary

Small endpoint-mirror CLIs no longer lose `mcp_tool_design` points for following the Printing Press SKILL guidance that says surfaces below 30 tools do not need MCP enrichment.

This raises the scorer's tool-design threshold to match that guidance, while still rewarding explicit intent or code-orchestration adoption whenever a spec declares it. The golden scorecard now treats the 10-tool fixture as unscored for `mcp_tool_design`, so the dimension leaves the denominator instead of silently applying a 5-point penalty.

## Validation

- `go test ./...`
- `golangci-lint run ./...`
- `scripts/golden.sh verify`

## Review Notes

Accepted for later: one headless review pass raised the broader question of whether `mcp_tool_design` should threshold on the full runtime MCP surface including cobratree-walked tools. Issue #1560 explicitly scoped this fix to `toolDesignMinEndpoints` and its 15/35/60 endpoint acceptance cases, so this PR preserves the existing scorer input shape and only aligns the threshold contract.

Closes #1560

---

Codex


tmchow:
@mergifyio queue

PR #2085, opened 2026-05-24
feat(cli): generator-wide self-learning CLI loop

mvanhorn:
## Summary

Lifts the prediction-goat self-learning loop into the generator so every printed CLI gets `teach`, `recall`, `learnings` (+ `teach-pattern`, `teach-lookup`, `learnings forget`) by default. Plan: [`docs/plans/2026-05-23-002-feat-generator-wide-self-learning-cli-plan.md`](docs/plans/2026-05-23-002-feat-generator-wide-self-learning-cli-plan.md). Authoring guide: [`docs/SPEC-LEARN-AUTHORING.md`](docs/SPEC-LEARN-AUTHORING.md).

This is Phase 1 only (generator templates + plumbing). Phase 2 (library sweep tool) and Phase 3 (full library sweep across 168 CLIs) are explicit follow-up phases with checkpoint gates.

## What ships

- `learn:` spec block on `APISpec` (`TickerPatterns`, `Stopwords`, `EntityLookupSeeds`) with parse-time validation
- Generator-reserved `internal/learn/` namespace + reimplementation_check exemption
- Three-subpackage `internal/learn/` lift from prediction-goat (entities, lookups, patterns) with the recipes -> patterns rename and the `DefaultPredictionGoatConfig()` singleton dismantled in favor of Config-per-call
- **`StoreSchemaVersion` gated on `Learn.Enabled`**: non-learn CLIs stay at v2 with no schema delta and no rollback impact; learn-enabled CLIs advance to v3 for the additive `search_learnings`, `search_patterns`, `entity_lookups`, `teach_log_metadata`, `search_learnings_fts` tables. The v2->v3 migration test is wrapped in the same gate.
- **Spec validation rejects `learn.enabled: true` without `VisionSet.Store=true`** (`learn.enabled requires VisionSet.Store=true; the learn package depends on internal/store`), so the broken combo fails fast at generate time instead of producing uncompilable code.
- Six Cobra commands wired conditionally in `root.go.tmpl` with MCP `read-only` annotations on `recall` + `learnings list`; PersistentPreRunE skip-list machinery; `--no-learn` / `<PREFIX>_NO_LEARN` opt-out.
- **`shouldSkipLearnHook` is ancestor-aware** via `cmd.CommandPath()` segment matching, so `auth login`, `completion bash`, `help <anything>`, etc. correctly skip `runLearnInitOnce` instead of opening/migrating/seeding the local store on framework commands.
- Spec -> runtime config init shim (`learn_init.go.tmpl`) that translates `spec.Learn` into a `*entities.Config` and seeds `entity_lookups` at startup
- SKILL.md / README.md / AGENTS.md template updates with the four-branch automatic-learning protocol
- **Scorecard +0.5 tier-1 credit** keyed off `internal/learn/doc.go` (the artifact the generator actually emits), with the unrelated `import.go` presence credit preserved alongside it
- CI: `scripts/verify-learn-purity.sh` + `.github/workflows/verify-learn-purity.yml` block domain identifiers (kalshi, polymarket, etc) from leaking into generator templates. **Scope expanded** to scan `teach*.go.tmpl` and `learn/*.go.tmpl` in addition to `learn*.go.tmpl` and `learn_*/*.go.tmpl` (32 templates scanned, up from 17), so the gate also covers the emitted learn CLI surface.
- **Domain-neutral `teach-pattern` example** (`"items in {entity}"` / `GROUP-{entity:category}` / `entity-kind "category"`) so help text shipped into every learn-enabled CLI doesn't inherit prediction-market vocabulary
- Golden fixture `generate-learn-loop-api/` locks the deterministic emission shape

## Verification

- `go test ./...`: every package passes across 34 packages
- `scripts/golden.sh verify`: 24/24 cases (1 new)
- `scripts/verify-learn-purity.sh`: 32 learn + teach templates clean
- `go vet ./...`, `go fmt ./...`, `golangci-lint run ./...`: clean
- Golden case emits + the emitted CLI compiles + its internal tests pass
- Manual: generated a fresh non-learn API spec, confirmed `internal/store/store.go` contains `StoreSchemaVersion = 2` and zero learn-table DDL
- Manual: generated a fresh learn-enabled API spec, confirmed it compiles and the `teach-pattern --help` example reads domain-neutral

## Phase 1 checkpoint

Per the plan: if the shape needs more time to bake, stop here. No library impact yet. Phase 2 (sweep tool) and Phase 3 (full sweep) are gated on Phase 1 review + a Phase 1.5 single-CLI publish-flow dry-run that hasn't run yet.

## Test plan

- [ ] CI green (`golden`, `full-golden`, `build-and-test` were the three that needed the gate fix to flip)
- [ ] Spot-check a learn-enabled fresh print produces a working CLI with all 6 commands
- [ ] Confirm AGENTS.md changes match repo conventions


internal/generator/templates/store.go.tmpl
```diff
@@ -42,7 +42,7 @@ func IsUUID(s string) bool {
 // shape — adding columns, dropping indexes, changing FTS5 tokenizers —
 // so an older binary refuses to open a newer database rather than silently
 // producing wrong results against a schema it cannot read.
-const StoreSchemaVersion = 2
+const StoreSchemaVersion = 3
```
tmchow: This version bump needs to be gated with the schema change. Right now non-learn CLIs emit `StoreSchemaVersion = 3` while the learn tables below are gated by `Learn.Enabled`, so a re-generated non-learn CLI can stamp an existing DB to v3 without adding any schema it actually needs. That breaks rollback to the current v2 binary, which will refuse to open the DB. Please make the constant conditional, e.g. v3 only when `.Learn.Enabled`, and update the schema-version tests/goldens accordingly.

internal/generator/learn_store_test.go
```diff
@@ -0,0 +1,148 @@
...
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/mvanhorn/cli-printing-press/v4/internal/spec"
+)
+
+// TestGenerateStoreSchemaVersionBumpedToV3 verifies the generator stamps
```
tmchow: This test is locking in the behavior that worries me architecturally. The disabled case should assert that the generated store stays at v2 and emits no learn schema; only the learn-enabled case should advance to v3. For this kind of generator-wide feature, `Learn.Enabled=false` needs to mean no on-disk migration or rollback impact.

internal/generator/generator.go
```diff
@@ -1693,6 +1705,40 @@ func (g *Generator) renderOptionalSupportFiles() error {
 		}
 	}
 
+	// Emit the self-learning loop (entities + normalize + match +
+	// recall + teach + teach_log + preseed + lookups + patterns) only
+	// when the spec opts in via Learn.Enabled. The schema migrations
+	// for the supporting tables are emitted from store.go.tmpl under
+	// the same gate; see internal/generator/templates/learn/doc.go.tmpl
+	// for the package-level design and the schema-adaptation note.
+	if g.Spec.Learn.Enabled {
```
tmchow, line 1844: This gate is not enough by itself because every emitted learn CLI imports/uses `internal/store`, but `internal/store` is emitted only when `VisionSet.Store` is true. A spec with `learn.enabled: true` and no store vision will generate uncompilable code. Please either make learn force store emission, reject that combo at validation/planning time with a clear error, or decouple learn from the normal store package.

internal/generator/templates/root.go.tmpl
```diff
@@ -319,6 +327,16 @@ Run '{{.Name}}-pp-cli doctor' to verify auth and connectivity.{{backtick}},
...
 			flags.freshnessMeta = autoRefreshIfStale(cmd.Context(), flags, resources)
 		}
+{{- end}}
+{{- if .Learn.Enabled}}
+		// Seed entity_lookups from spec.Learn.EntityLookupSeeds once per
+		// process. Skipped for framework commands that should never
+		// touch the local store (auth, doctor, help, etc.) and for
+		// --no-learn invocations so deterministic agent flows don't
+		// race a background seed.
+		if !noLearnActive(flags) && !shouldSkipLearnHook(cmd.Name()) {
```
tmchow: This hook is live today, but it skips based only on `cmd.Name()`. Nested framework commands like `auth login` or `completion bash` miss the top-level skip entries (`auth`, `completion`) and will still open/migrate/seed the learn store. That is an ambient side effect on commands the comment says should never touch local store. Please make this command-path/ancestor aware, or better, seed lazily only from the learn commands that need it.

internal/generator/templates/teach.go.tmpl
```diff
@@ -0,0 +1,960 @@
...
+	var strategy string
+	var entityKind string
+	var dbPath string
+
+	cmd := &cobra.Command{
+		Use:   "teach-pattern",
+		Short: "Install a manual generalization pattern (query_template, resource_template, entity_kind)",
+		Long: `Adds one row to search_patterns. The recall path uses patterns to
+substitute live query entities into a resource template, so a pattern
+with query_template="odds {entity} wins world cup" and
```
tmchow: This example is still prediction/sports-market flavored and will ship into every learn-enabled CLI help surface. The purity workflow does not catch it because it only scans `learn*.go.tmpl` and `learn_*/*.go.tmpl`, not `teach.go.tmpl`. Please make the example domain-neutral and extend the purity gate to cover the generated learn CLI surface too.

internal/pipeline/scorecard.go
```diff
@@ -1150,7 +1150,11 @@ func scoreVision(dir string) int {
 	if fileExists(filepath.Join(cliDir, "tail.go")) {
 		tier1 += 0.5
 	}
-	if fileExists(filepath.Join(cliDir, "import.go")) {
+	if fileExists(filepath.Join(dir, "internal", "learn", "doc.go")) {
+		tier1 += 0.5
+	}
+	// time, so the credit stays modest (+0.5).
+	if fileExists(filepath.Join(dir, "internal", "learn", "learn.go")) {
```
tmchow: This check appears stale: `renderLearnFiles` emits `internal/learn/doc.go`, `normalize.go`, `recall.go`, `teach.go`, etc., but not `internal/learn/learn.go`. As written, this credit never fires even for a learn-enabled generated CLI. Please check for an artifact that is actually emitted, or collapse this into the existing `doc.go` presence check.

tmchow (changes requested):
I support the direction and the value of the self-learning loop, especially as an opt-in generated primitive for search/discovery-shaped CLIs. I am requesting changes because the current implementation does not yet preserve the generator contract for disabled/non-store CLIs and it introduces ambient store side effects that are too risky for a foundational generator change.

Please fix the blocking issues inline before merge: disabled CLIs should have no schema/runtime delta, learn-enabled CLIs need an explicit store contract, startup seeding should not touch framework commands unexpectedly, and the learn surface/purity gate should be domain-neutral. Also please refresh golden output and the PR description after those fixes; live golden/full-golden/build-and-test are currently red.

tmchow:
I checked the related printing-press-library PRs (#826 and #827) against this review. They confirm the Phase 2 sequencing risk: #826 embeds the current learn templates/schema/root wiring from this PR, and #827 is a findings-only pilot that should retry only after this generator contract is settled.

For this PR’s follow-up, please make the downstream contract explicit in the PR body/plan before library sweep work continues:

- learn-enabled schema/store behavior: what version is emitted, and whether no-learn/no-store CLIs remain unchanged
- init strategy: ambient root hook vs lazy/explicit teach/recall initialization, with the same behavior fresh prints and swept CLIs should share
- domain-neutrality: teach.go/CLI help text is part of the generated surface and should be covered by the purity gate
- template sync point: library PR #826 should rebase/refresh from the accepted generator commit rather than the current disputed snapshot

I requested changes on library PR #826 for that reason, but left #827 as non-blocking because it only records pilot findings and ships no retrofitted CLI diffs.

internal/generator/templates/store.go.tmpl
mvanhorn: Fixed in 8f147fe4. Gated `const StoreSchemaVersion = {{if .Learn.Enabled}}3{{else}}2{{end}}`, wrapped `TestMigrate_V2ToV3IsAdditive` in the matching gate, and refreshed the sync-walker (non-learn) golden so it stays at v2. Tested with a fresh non-learn print: `internal/store/store.go` now contains `StoreSchemaVersion = 2` and zero learn-table DDL.

internal/generator/learn_store_test.go
mvanhorn: Fixed in eaf028ae. Split into `TestGenerateStoreSchemaVersion_DisabledStaysV2` (asserts v2 and zero learn-table DDL) and `TestGenerateStoreSchemaVersion_EnabledAdvancesToV3WithLearnTables` (asserts v3 plus all 5 learn-table creates). The old uniform-bump test is removed.

internal/generator/generator.go
mvanhorn: Fixed in 5b6fdc16. `prepareOutput()` now returns `learn.enabled requires VisionSet.Store=true; the learn package depends on internal/store` if the combo is set, after `constrainVisionTemplates` finalizes the vision set. Covered by `TestGenerateLearnEnabledRequiresStoreVision` in `learn_store_test.go`.

internal/generator/templates/root.go.tmpl
mvanhorn: Fixed in dfaf45fe. Call site now passes `cmd.CommandPath()`; `shouldSkipLearnHook` iterates `strings.Fields(commandPath)` and matches any segment against `learnHookSkipList`. So `auth login`, `completion bash`, `help <anything>`, etc. correctly skip. New `TestSkipLearnHook_FrameworkCommandsDoNotCreateDefaultDB` asserts no default DB file is created when running `auth setup`, `completion bash`, `version`, `help items`, `doctor` — the observable side-effect you flagged.

internal/generator/templates/teach.go.tmpl
mvanhorn: Fixed in b79653fb. Example is now `"items in {entity}"` / `GROUP-{entity:category}` / `entity-kind "category"` (no domain leakage). `scripts/verify-learn-purity.sh` now scans `teach*.go.tmpl` and `learn/*.go.tmpl` in addition to the prior globs (32 templates scanned vs 17 before), so the gate covers the emitted learn CLI surface.

internal/pipeline/scorecard.go
mvanhorn: Fixed in 00cce054. The check now points at `internal/learn/doc.go` (what `renderLearnFiles` actually emits), the dead `learn.go` check is gone, and the orphaned comment fragment is replaced with a clean one. Also restored the unrelated `import.go` presence credit that was accidentally dropped in the earlier web-edit. Covered by `TestScoreVision_LearnLoopGoldenFixtureAddsHalfPoint` (runs scoreVision against the real `generate-learn-loop-api/learn-loop-example/` golden fixture) and `TestScoreVision_ImportPresenceAddsHalfPoint`.

tmchow (approved):

tmchow:
@mergifyio queue

PR #2105, opened 2026-05-25
fix(cli): route GraphQL reads around verify gate

### LINKED ISSUE #1955: generator: GraphQL transport rides verify-gated POST — generated sync noops under PRINTING_PRESS_VERIFY
**Component:** generator · **Priority:** P2 · Surfaced by the first GraphQL CLI retro (dice-fm).

## Summary
The generated GraphQL transport's `Query`/`Mutate`/`PaginatedQuery` ride the verify-gated `Post`, so the generated GraphQL `sync` command short-circuits to the `__pp_verify_synthetic__` noop under `PRINTING_PRESS_VERIFY=1` and silently reports "synced 0".

## Root cause
- `internal/generator/templates/graphql_client.go.tmpl:170` — `Client.Query` calls `c.Post(graphqlEndpointPath, req)`, which routes through the verify-gated `do()`. Mutating verbs short-circuit under `PRINTING_PRESS_VERIFY=1` (without `LIVE_HTTP=1`).
- `internal/generator/templates/graphql_sync.go.tmpl:307` — generated GraphQL sync calls `c.Query(ctx, def.Query, variables)`, so it receives the synthetic envelope, parses zero nodes, and exits 0 with 0 records.
- The generated per-endpoint commands already avoid this by using `PostQueryWithParams` (→ `doRead`, ungated). The GraphQL transport methods do not.

GraphQL queries are reads riding a POST verb — exactly the case `doRead` exists for.

## Durable fix
In `graphql_client.go.tmpl`, route `Query`, read-`Mutate`, and `PaginatedQuery` through the ungated read path (`doRead` / the `PostQueryWithParams` plumbing in `client.go.tmpl`) instead of `Post`. Keep the gated path only for genuine GraphQL mutations if/when the generator emits them.

## Acceptance
- positive: generate a GraphQL CLI; under `PRINTING_PRESS_VERIFY=1` (no `LIVE_HTTP`), a `Query`-backed read reaches the mock transport and parses nodes (sync lands rows).
- negative: a genuine GraphQL mutation still short-circuits under verify.

## Evidence
dice-fm's hand-authored `dice_query.go` deliberately used `PostQueryWithParams` instead of the generated `client.Query` to dodge this; the original generated `sync.go` used `c.Query` and would have shipped a verify-green-but-noop sync.

Evidence: n=1 (dice-fm) in catalog/library — but the defect is deterministic in the shared GraphQL transport template, provable by inspection, and the fix is narrow and safe. Full retro: `~/printing-press/manuscripts/dice-fm/20260511-193451/proofs/` (local).

tmchow:
## Summary

Generated GraphQL read queries now reach the real transport under `PRINTING_PRESS_VERIFY=1` instead of being mistaken for mutating POSTs and returning the synthetic noop envelope. This fixes GraphQL sync paths that could otherwise report a clean zero-row sync without ever exercising the mock or live read response.

The GraphQL mutation helper remains on the verify-gated POST path. Cost-throttled GraphQL clients keep their budget/update/retry behavior for both reads and mutations, with only the read transport using the read-only POST helper.

Closes #1955

## Verification

- `go test ./internal/generator ./internal/pipeline`
- `scripts/golden.sh verify`
- `golangci-lint run ./...`

Note: `go test ./...` was also run once and hit a transient 5s timeout in `internal/pipeline TestRunLiveDogfoodProcessRetriesTransientAuth401`; rerunning that exact test passed, and the focused `internal/pipeline` package run passed afterward.

### DIFF
--- a/internal/generator/cost_throttling_test.go
+++ b/internal/generator/cost_throttling_test.go
@@ -155,6 +155,16 @@ func TestGenerateCostThrottlingPrimitives(t *testing.T) {
 		"graphql.go must parse extensions.cost via updateThrottleFromBody")
 	assert.Contains(t, graphqlGo, "throttleStatus",
 		"graphql.go cost parser must read throttleStatus")
+	queryWithThrottleBody := generatedFunctionBody(t, graphqlGo, "func (c *Client) queryWithThrottle(")
+	assert.Contains(t, queryWithThrottleBody, "c.PostQueryWithParams(ctx, graphqlEndpointPath",
+		"throttle-aware GraphQL reads must use the read-only POST helper")
+	assert.NotContains(t, queryWithThrottleBody, "c.Post(ctx, graphqlEndpointPath",
+		"throttle-aware GraphQL reads must not use the verify-gated POST helper")
+	mutateWithThrottleBody := generatedFunctionBody(t, graphqlGo, "func (c *Client) mutateWithThrottle(")
+	assert.Contains(t, mutateWithThrottleBody, "c.Post(ctx, graphqlEndpointPath",
+		"throttle-aware GraphQL mutations must keep the verify-gated POST helper")
+	assert.Contains(t, mutateWithThrottleBody, "c.graphqlWithThrottle",
+		"throttle-aware GraphQL mutations must preserve cost-budget handling")
 
 	// root.go must register --throttle-mode and validate its value.
 	rootGoBytes, err := os.ReadFile(filepath.Join(outputDir, "internal", "cli", "root.go"))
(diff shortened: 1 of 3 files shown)


tmchow:
Accepted for later: local review flagged that typed MCP endpoint tools for GraphQL operations may still need a dedicated GraphQL semantics pass. This PR stays scoped to issue #1955: generated GraphQL client reads and sync use the read-only POST helper under verify mode, while genuine mutations remain on the gated POST path. The runtime Cobra mirror still exposes store-population commands such as sync to agents, so this does not block the issue fix.

PR #2150, opened 2026-05-25
feat(skills): capture reachability 4xx body hints

### LINKED ISSUE #987: [P2] WU-2: Phase 1.9 reachability gate captures 4xx response body content
# Phase 1.9 reachability gate captures 4xx response body content

## Problem

Phase 1.9 of the main skill (`skills/printing-press/SKILL.md`) probes `verify_path` with a single GET request to confirm the host is reachable and credentials authenticate. The gate currently records the HTTP status code only — not the response body.

Many credentialed APIs encode their real access model (tier limits, allowed-set restrictions, scope-limited capabilities, quota state) in the body of 4xx error responses rather than in GET endpoint availability:

- A vendor's GET endpoints all return 200, but their PUT/POST returns 403 with `{"error":"...","allowed":["A","B","C"]}` listing the subset of capabilities the user's tier permits.
- Tier upgrades or rate limits surface in 429 response bodies with structured "next allowed time" or "current plan" fields.
- Scope-limited tokens succeed on read endpoints and fail on write endpoints with body content naming the missing scope.

When the absorb manifest in Phase 1.5 is built on the assumption that all spec-described capabilities are available to the user's tier, and the actual restriction surfaces only later (during Phase 5 live dogfood), the run wastes a generation cycle on features the user's tier can't access.

Concrete example: odds-api.io's `/bookmakers/selected/select` PUT returns 403 with body `{"error":"...","allowed":["A","B","C","D","E"]}` listing the per-tier-allowed-set of bookmakers. The GET probe Phase 1.9 currently runs returns 200 with no hint of this restriction. Phase 1.5 locks in a manifest assuming any bookmaker can be subscribed to; the assumption breaks at Phase 5 when the live PUT surfaces the real allowed set, and 4+ Phase 3 novel commands have already been built around the wrong assumption.

## Cross-API evidence

The "tier info in 4xx body" pattern is broad across credentialed SaaS APIs:

- **Stripe** — POST customers in livemode with a test-mode key returns 403 with body describing the mode mismatch. `unit_amount` limits surface in 400 body content with `detail` field.
- **Twilio** — POST Messages on trial accounts returns 400 with body indicating "trial accounts can only send to verified numbers" and naming the missing verification.
- **Plaid** — item-create POST in sandbox vs development surfaces tier-locked products in 403 body listing supported products.
- **Discord** — POST guilds.channels with insufficient bot permissions returns 403 with body listing the missing permission flag.

## Proposed fix

Edit Phase 1.9 in `skills/printing-press/SKILL.md` to:

1. **Always capture and inspect 4xx response bodies from the existing GET `verify_path` probe.** Read the body, heuristic-scan for tier-restriction keywords (`tier`, `allowed`, `permitted`, `subscription`, `quota`, `plan`, `scope`, `limit`), and surface any matched lines verbatim into the Phase 1 research brief's `## Reachability Risk` section.

2. **Optional second tier (opt-in via spec extension)**: when the spec declares an endpoint with `x-pp-safe-probe: true` (or equivalent), Phase 1.9 may also probe that endpoint. This is opt-in only because mutation probes against arbitrary APIs would create real resources, send real messages, or charge real money — only the spec author can safely declare a mutation endpoint as idempotent/probe-safe.

**Explicitly do NOT** probe arbitrary mutation endpoints. The skill instruction must call out the danger: a generic "probe any PUT/POST" rule would break safety on non-idempotent surfaces (account creation, payment capture, message send). Body capture on GET is universally safe; mutation probes require per-endpoint opt-in.

## Acceptance criteria

- **Positive**: when Phase 1.9 probes a `verify_path` that returns 4xx with body matching tier-keyword heuristics, matched lines appear under `## Reachability Risk` → "Tier/permission hints from 4xx body" in the Phase 1 brief.
- **Positive**: when a spec declares `x-pp-safe-probe: true` on an endpoint AND the user provided a key, that endpoint is probed and its body captured.
- **Negative**: when `verify_path` returns 200, no tier-hint section appears.
- **Negative**: endpoints without an explicit `x-pp-safe-probe: true` are never mutation-probed even when a key is available.

## Priority rationale

P2 — affects every credentialed API where the real access model lives in mutation-response bodies. Fallback is "agent remembers to curl 4xx bodies during research" which is inconsistent in practice. Incremental fix; safety-bounded (GET-only by default; opt-in mutation probes).

## Related issues

- #925 (related-area: Phase 1.9 LAN-only / mDNS-discovered API carve-out — same gate, different unhandled case)
- #895 (related-area: doctor distinguishes HTTP 401 vs 403 — same "4xx is signal, not just failure" pattern at a different phase)

tmchow:
## Summary

Phase 1.9 now preserves 4xx response bodies during the reachability probe, extracts bounded tier or permission hints into the research brief, and keeps mutation probes disabled unless a spec explicitly marks one endpoint as `x-pp-safe-probe: true`.

Closes #987

## Verification

- `bash .github/scripts/validate-skill-docs.sh`
- `git diff --check`

### DIFF
--- a/docs/SPEC-EXTENSIONS.md
+++ b/docs/SPEC-EXTENSIONS.md
@@ -39,6 +39,7 @@ in the same change as any new `Extensions["x-*"]` lookup in that file.
 | `x-resource-id` | path item | `Endpoint.IDField` | No |
 | `x-critical` | path item | `Endpoint.Critical` | No |
 | `x-tier` | path item or operation | `Endpoint.Tier` | No |
+| `x-pp-safe-probe` | operation | Phase 1.9 reachability guidance | No |
 | `x-pp-sync-walker` | operation | `Endpoint.Walker` | No |
 
 ## `info` Extensions
@@ -1006,6 +1007,35 @@ paths:
   /premium/search:
     get:
       x-tier: paid
+      responses:
+        "200": {description: ok}
+```
+
+### `x-pp-safe-probe`
+
+Marks a mutation endpoint as explicitly safe for the Phase 1.9 reachability gate
+to call once as an optional second probe after the low-risk GET/body capture.
+This extension is consumed by Printing Press skill guidance rather than the Go
+OpenAPI parser; it documents author intent for agents reviewing a resolved spec.
+
+Parsed field: Phase 1.9 reachability guidance
+
+Rules:
+- Optional.
+- Must be on an operation, not the root, `info`, or path item.
+- Accepts native boolean `true` only.
+- Use only for idempotent or otherwise harmless operations for the real account
+  being used.
+- Absence means mutation probing is not allowed; agents must stop after the
+  GET/body reachability capture.
+
+Example:
+
+```yaml
+paths:
+  /webhooks/test:
+    post:
+      x-pp-safe-probe: true
       responses:
         "200": {description: ok}
 ```
--- a/docs/SPEC-EXTENSIONS.md
+++ b/docs/SPEC-EXTENSIONS.md
@@ -1018,7 +1018,7 @@ to call once as an optional second probe after the low-risk GET/body capture.
 This extension is consumed by Printing Press skill guidance rather than the Go
 OpenAPI parser; it documents author intent for agents reviewing a resolved spec.
 
-Parsed field: Phase 1.9 reachability guidance
+Parsed field: none; consumed by skill guidance only
 
 Rules:
 - Optional.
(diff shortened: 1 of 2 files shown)


cathrynlavery:
Addressed the latest Greptile findings in `8907565a`.

Fixes:
- Added cleanup for the `mktemp` reachability response body file via `trap 'rm -f "$body_file"' EXIT`.
- Removed the ineffective `curl -S` flag while stderr is redirected.
- Made the WebFetch fallback explicitly preserve and scan 4xx body text for the same tier/permission hints.
- Registered/documented `x-pp-safe-probe` in `docs/SPEC-EXTENSIONS.md` as a Phase 1.9 guidance extension.

Verification:
- `git diff --check`

@greptile-apps please re-review current head.
@coderabbitai review

PR #2161, opened 2026-05-25
fix(cli): accept nested internal body schemas

### LINKED ISSUE #1267: spec-parser: internal YAML body field can't express nested-object POST shapes (GraphQL, Power BI, Slack)
## Summary
The internal YAML spec format's `body:` field is typed as `[]Param` (a flat list of `name + type + description`). This works for simple flat-body POSTs (most REST APIs) but cannot express common richer shapes:

- **GraphQL** (`{query: string, variables: object}`)
- **Power BI executeQueries** (`{queries: [{query: string}], serializerSettings: {includeNulls: bool}, impersonatedUserName: string}`)
- **Slack chat.postMessage** (`{channel, text, blocks: [...]}`)
- **Stripe POST /v1/customers** (`{name, address: {line1, city, ...}, metadata: {key: value}}`)

When an agent hand-writes an internal YAML spec for an API without an OpenAPI, this limitation forces dropping the endpoint from the spec and hand-coding the entire POST command (Go struct types, JSON encoding, validation). The Power BI CLI's `dax run` command is ~270 lines of hand-written code that the generator could have scaffolded if the spec could express the body shape.

First attempt that triggered the error:

```yaml
endpoints:
  execute_queries:
    method: POST
    path: /datasets/{datasetId}/executeQueries
    body:
      schema:
        type: object
        properties:
          queries:
            type: array
            items:
              type: object
              properties:
                query:
                  type: string
```

Parser response:

```
Error: parsing spec ...powerbi-spec.yaml: yaml: unmarshal errors:
  line 175: cannot unmarshal !!map into []spec.Param
  line 211: cannot unmarshal !!map into []spec.Param
```

The agent's first instinct (OpenAPI-style nested) is rejected; the only working shape is the flat `[]Param`, which doesn't fit.

## Why it matters
Affects any rich-body POST/PUT endpoint in any API without an OpenAPI spec (which is the case the internal YAML format exists for). Three named APIs with evidence:
- **GitHub GraphQL** — POST /graphql body is `{query, variables}` — every GraphQL API has this shape.
- **Slack chat.postMessage** — POST with `blocks` and `attachments` arrays of typed objects.
- **Stripe POST /v1/customers** — nested `address` object and `metadata` map.

For all three, agents either hand-build the entire command or skip the endpoint.

## Proposed fix
Extend the spec parser at `internal/spec/` to accept `body:` as either of:

1. The current `[]Param` flat-list shape (for simple bodies — preserves backward compatibility).
2. A new object-with-properties shape that maps to JSON Schema-lite:

```yaml
body:
  properties:
    queries:
      type: array
      items:
        type: object
        properties:
          query: { type: string, required: true }
    serializerSettings:
      type: object
      properties:
        includeNulls: { type: bool }
```

When the body is the object shape, the generator emits a typed Go struct + JSON encoding in the generated command, plus flags for the top-level scalar fields and a `--body-file` flag for the user to supply the nested portion as JSON. This is the same pattern OpenAPI-sourced commands already produce — the work is making it available for hand-crafted specs.

## Acceptance criteria
- **Positive:** A fixture spec declaring a GraphQL-shaped body (`{query: string, variables: object}`) parses successfully and the generated command emits a typed struct with `--query` and `--variables-file` flags.
- **Positive:** Existing `body: [{name, type}]` flat-list syntax still works unchanged (backward compat).
- **Positive:** A nested body with array-of-objects (`{queries: [{query: string}]}`) emits a slice of typed structs.
- **Negative:** Malformed body shapes produce a clear parser error pointing to the failing line, not the current "cannot unmarshal !!map into []spec.Param" YAML internal message.

## Scope boundary
Does NOT include full OpenAPI-equivalence with `$ref`s and discriminators — just the common nested-object and array-of-objects shapes. Composition via `$ref` can be a future increment.

## Component
`comp:spec-parser` — `internal/spec/`. Plus generator template work for POST/PUT command emission.

## Related issues
- #1240 — POST/PUT files balloon to >40k lines for deeply-nested types and OOM the Go compiler. Different direction (output too big) but same general territory (request body schemas). Worth keeping the fix here from regressing #1240's compile-time problem.

## Source
Retro from a Power BI CLI generation, Run 20260512-201900. Cost ~270 lines of hand-written `internal/cli/dax.go` that scaffolding could have absorbed most of.

tmchow:
## Summary

Internal YAML specs can now describe JSON-schema-lite request bodies with `body.properties` or `body.schema.properties`, including nested object and array shapes copied from API docs. Printed CLIs expose reachable scalar leaves as typed body flags, and generated MCP tools plus `tools-manifest.json` now publish the same nested body inputs so agent calls build the same request payload.

The legacy list-style `body:` format remains supported, and the Printing Press skill reference now documents the object-schema form.

Closes #1267

## Verification

- `go test ./internal/spec -run TestParseBodyObjectSchemaMalformed -count=1`
- `go test ./internal/generator -run 'TestGenerateInternalYAMLBodyObjectSchema|TestGenerateRenamesObjectBodyFieldCollidingWithPathParam' -count=1`
- `go test ./internal/pipeline -run TestWriteToolsManifest_NestedBodyParamsMatchPublicSurface -count=1`
- `go test ./internal/spec ./internal/generator ./internal/pipeline`
- `scripts/golden.sh verify`
- `go test ./...`
- `git diff --check`
- `golangci-lint run ./internal/spec/...`


tmchow:
@greptileai

PR #2318, opened 2026-05-26
feat(generator): emit atomic OAuth refresh-token rotation package

shardin2:
## Summary
- New `spec.Auth.AtomicRefreshTokens bool` flag gates emission of an `internal/tokens/` package with atomic-rename `tokens.json` persistence, interprocess flock, NDJSON audit log, macOS Keychain backend, and `ClientIDSuffix` integrity check. Targets providers whose refresh_token rotates on every use (Withings, Strava, Fitbit pattern), where naive `os.WriteFile` persistence can lose rotations between request-send and response-receive and forces weekly re-OAuth.
- Six new templates (`tokens_{audit,keychain,store,refresh,store_test,refresh_test}.go.tmpl`) plus an emission predicate `hasAtomicRefreshTokens()` wired into `renderOptionalSupportFiles` in `generator.go`.
- `doctor.go.tmpl` gains a `hasUsableTokensFile()` helper and a 3-source auth check (`config.toml AuthHeader` → `tokens.json` bearer → env-derived OAuth inputs) gated on the same spec field. Previously, env-driven and `tokens.json`-backed setups produced a false `not configured` verdict.

## Background
Verified during a real Withings CLI migration where multiple legitimate auth setups all produced `auth: not configured` from `doctor` while five successful token rotations proved auth worked. The hand-authored `internal/tokens/` lived as a `.printing-press-patches.json` preservation entry; this PR upstreams it as a structural emission so the fix is regenerated instead of preservation-dependent.

## Test plan
- [x] `go build ./...` from repo root — PASS
- [x] `pp generate --spec <spec-with-atomic_refresh_tokens:true> --output <tmpdir>` — emits `internal/tokens/*` + `internal/cli/doctor.go` with the env-aware switch
- [x] Post-generation shipcheck (`go mod tidy`, `govulncheck`, `go vet`, `go build`, runnable binary, `--help`, `version`, `doctor`) — all PASS
- [x] Diff vs prior hand-authored library (`internal/tokens/`): comments-only differences + the `DO NOT EDIT` generated-header line + a defensible `url`→`tokenURL` rename to avoid colliding with the imported `net/url` package
- [x] `grep` of emitted `doctor.go` matches `hasUsableTokensFile`, `case hasUsableTokensFile()`, `"oauth2:tokens.json"`, `configured (login pending)` — 3-source switch landed
- [ ] CI on this PR

## Backward compatibility
- The flag defaults to `false`. Specs that don't set `atomic_refresh_tokens: true` emit no `internal/tokens/` directory and no behavior changes.
- `doctor.go.tmpl` auth-check is `{{- if .Auth.AtomicRefreshTokens}} ... {{- else}} ... {{- end}}` — the existing branch is preserved verbatim for specs without the flag set.
- `os` import in `doctor.go.tmpl` was already gated on several upstream conditions; this PR adds `.Auth.AtomicRefreshTokens` to that OR-list so the import only appears when actually needed.


tmchow:
Please don't submit PRs without an issue filed first. It helps ability to correctly review and triage things.

shardin2:
It sounds like drive-by PRs without issues create more cleanup than they save. You're probably looking at this thinking somebody dumped half-finished code on you and expected you to do the triage work for them.

Both findings from Greptile are correct. The Withings-specific assumptions shouldn't have been generalized into the upstream templates, and the keychain `Clear()` error-swallowing is a real bug regardless of provider.

Closing this. Would it be ridiculous to file an issue first describing what I was trying to solve (atomic refresh-token rotation for providers whose refresh_token rotates on every use), and let you decide whether it's a fit before any code lands?

Either way, sorry for the noise.

PR #2333, opened 2026-05-26
fix(cli): make agentcookie soft + stop publishing plans

### LINKED ISSUE #2320: generator: private agentcookie dependency breaks cookie-auth CLI builds
**Component:** generator · **Priority:** P1 · Filed from a `/printing-press` retro (streeteasy, cookie-auth, HAR-sourced).

## What happened
`generate` for an `auth.type: cookie` CLI emitted `require github.com/mvanhorn/agentcookie v0.14.0-beta.1` in `go.mod` and `import ".../agentcookie/pkg/agentcookiesecret"` in `internal/config/config.go`. That module is not publicly accessible — `git ls-remote https://github.com/mvanhorn/agentcookie` returns "Repository not found" (and the same over the SSH `insteadOf` rewrite). So `go mod tidy` and the `go build` quality gate fail for **any** cookie/composed-auth CLI on a machine without access to that repo.

The failure is opaque: it surfaces as cascading test-dependency resolution errors across `mark3labs/mcp-go` and `modernc.org/sqlite`, all blocked on fetching the agentcookie go.mod.

## Root cause
The cookie-auth template path unconditionally wires the agentcookie secrets-bus integration into `config.go` + `go.mod`, depending on a private/unpublished module. The integration is *optional* at runtime (it's a secrets-bus override layered on top of the normal env/config cookie read), but it's a hard build-time dependency.

## Frequency
Every cookie / composed / session-handshake-auth CLI (auth-type-gated, deterministic from the template) — independent of API shape or spec source.

## Durable fix
The generator must not emit a `require`/`import` on a non-public module by default. Either make `agentcookie` public, gate the secrets-bus import behind a build tag / spec opt-in flag, or drop it from the default cookie-auth emission so a clean cookie-auth CLI builds against public modules only.

## Acceptance criteria
- **positive:** generate a cookie-auth CLI on a machine without agentcookie access → `go mod tidy && go build ./...` passes.
- **negative:** with the integration opted in and the dep available, the agentcookie secrets-bus path is still emitted.

## Evidence
`generate` quality gate failed repeatedly with `github.com/mvanhorn/agentcookie@v0.14.0-beta.1: reading .../go.mod ...: Repository not found`. De-vendoring the optional secrets-bus block from `config.go` + removing the `go.mod` require unblocked the build (env/config cookie path unaffected).

## Related
#981 (cookie securityScheme header-vs-cookie wiring) — adjacent cookie-auth area, distinct defect.

mvanhorn:
## Summary

Two unrelated-but-coincident fixes. Newly printed CLIs fail to install because they require a private repo:

1. **agentcookie integration is now soft.** PR #1972 added `require github.com/mvanhorn/agentcookie v0.14.0-beta.1` to every newly-printed CLI's `go.mod` whenever the CLI had non-cookie auth. The `agentcookie` module is private, so `go install <any-pp-cli>@latest` now fails `go mod download` for anyone without read access. This branch strips the require, removes the typed import from the generated config loader, and replaces it with a stdlib-only stat for `.agentcookie-managed` in the same directory as the config file. When the marker is present, `auth_source` upgrades to `"agentcookie"`; when absent, the CLI behaves exactly as it did pre-#1972 (no import, no network call, no error).

2. **Plan documents are no longer tracked.** `docs/plans/` had ~200 plans committed and pushed to this public repo, frequently describing in-progress or unreleased work. `/docs/plans/` is now in `.gitignore`; the previously-tracked files are removed from the index (working-tree copies preserved). Historical plans remain visible in git history per explicit direction; no force-push.

## Why this is a `fix`, not a `feat` or breaking change

The contract every printed CLI promises is "installs cleanly via `go install`." That contract is being restored, not changed. Per AGENTS.md's "Breaking changes" criteria, removing emitted go.mod content does not qualify as `!` when the removal undoes a regression rather than altering a documented surface.

## What ships

| Commit | Concern |
|--------|---------|
| `chore(cli): untrack docs/plans from public repo` | gitignore + `git rm --cached` + AGENTS.md rule + presence test |
| `fix(cli): drop private agentcookie require from generated CLIs` | template diffs + helper cleanup + golden regen |
| `feat(cli): soft-detect agentcookie-managed credentials via on-disk marker` | stdlib marker stat + doctor line + README paragraph + per-fork test |
| `test(cli): guard against private-module requires in generated go.mod` | structural CI guard with extensible prefix list |

## Test plan

- [x] `go test ./...` — 5868 tests pass (was 5858; +3 from soft-detection test, +7 from private-dep guard test)
- [x] `scripts/golden.sh verify` — 25 cases pass after intentional golden regen for the template diffs
- [x] `golangci-lint run ./...` — no issues
- [ ] After merge: confirm release-please opens a `v4.15.0 -> v4.15.1` PR with the regression callout
- [ ] After release: `go install <a representative pp-cli>@latest` from a shell with no GitHub auth completes without error

Fixes #2320


tmchow:
@Greptileai review

tmchow:
@Greptileai review

tmchow:
@mergify queue

tmchow:
@Mergifyio queue

PR #2352, opened 2026-05-26
fix(cli): honor docs batch filter flags client-side

### LINKED ISSUE #2092: docs-generated CLIs surface filter flags the API ignores (e.g. funding_rate/batch --symbols is a server-side no-op)
## Summary

When generating from `--spec-source docs`, the press emits CLI flags from a documented endpoint's parameter list even when the endpoint ignores those parameters server-side. The flag looks functional but is a silent no-op — misleading to both humans and agents.

## Concrete case

Bitunix `GET /api/v1/futures/market/funding_rate/batch` is a "return everything" endpoint. The generator produced a `--symbols` flag on `market get-funding-rate-batch`, wired as a `symbols=` query param. The endpoint ignores it (verified live):

- no param → 720 rows
- `?symbols=BTCUSDT` → 720 rows
- `?symbol=BTCUSDT` → 720 rows

So `get-funding-rate-batch --symbols BTCUSDT` returns all ~720 pairs. The singular `get-funding-rate` (takes a symbol) is the real per-symbol path.

## Impact

- Misleading flag: users/agents reasonably expect `--symbols` to filter, and it silently doesn't.
- Agents burn context on the full list and may trust a "filtered" view that was never filtered.

## Suggested directions (any of)

1. Don't emit a flag for a parameter the endpoint doesn't actually honor (hard to detect from docs alone).
2. When a documented filter param has no server effect, generate **client-side filtering** for it (fetch, then filter the response by the documented key) so the flag is honest. This generalizes well — many "batch/all" list endpoints across exchange APIs return everything and rely on client filtering.
3. If neither is feasible automatically, mark such flags / note in `--help` that filtering is client-side / best-effort.

## Workaround in our CLI

We patched `bitunix-pp-cli` to filter client-side: stop sending the ignored param and filter the response by `symbol`, handling both the bare-array and `{code,data,msg}` envelope response shapes. Recorded as a local printing-press patch; happy to drop it if the generator grows this.

Related: #1987 (also `funding_rate/batch`, but a different failure — path-segment normalization).

tmchow:
## Summary

Docs-derived batch commands now keep simple filter flags honest when the upstream endpoint accepts but ignores the corresponding query parameter. For batch/all GET endpoints whose filter flag maps to a response item field, generated CLIs still send the documented query param and then narrow the returned JSON locally, including common `{data:[...]}` envelopes.

The filtering runs before client-side `--limit` truncation, fails open when the actual response shape does not contain the expected field, and is limited to docs-derived batch list endpoints so ordinary list endpoints keep trusting server-side filter semantics.

## Tests

- `go test ./internal/generator -run 'TestEndpointClientSideFilters|TestGeneratedDocsClientSideFilterNarrowsIgnoredBatchParam|TestGeneratedPromotedDocsClientSideFilterRunsBeforeLimit' -count=1`
- `scripts/golden.sh verify`
- `go test ./...`
- pre-push hook: `golangci-lint` passed with 0 issues

Closes #2092

---

GPT--5_Codex

### DIFF
--- a/internal/generator/client_side_filter_test.go
+++ b/internal/generator/client_side_filter_test.go
@@ -1,0 +1,318 @@
+package generator
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/mvanhorn/cli-printing-press/v4/internal/spec"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestEndpointClientSideFilters(t *testing.T) {
+	t.Parallel()
+
+	apiSpec := minimalSpec("filterdocs")
+	apiSpec.SpecSource = "docs"
+	apiSpec.Types = map[string]spec.TypeDef{
+		"FundingRate": {
+			Fields: []spec.TypeField{
+				{Name: "symbol", Type: "string"},
+				{Name: "query", Type: "string"},
+				{Name: "rate", Type: "string"},
+			},
+		},
+	}
+	endpoint := spec.Endpoint{
+		Method: "GET",
+		Path:   "/funding_rate/batch",
+		Params: []spec.Param{
+			{Name: "symbols", Type: "string"},
+			{Name: "limit", Type: "integer"},
+			{Name: "query", Type: "string"},
+		},
+		Response: spec.ResponseDef{Type: "array", Item: "FundingRate"},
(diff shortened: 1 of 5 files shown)


tmchow:
@greptileai review

tmchow:
@greptileai review

PR #2356, opened 2026-05-26
fix(cli): promote global scope query params

### LINKED ISSUE #1846: generator + spec-parser: spec staleness probe + PP-required tenant-filter pattern (per-tenant SaaS APIs)
## Summary

Two related findings from the CIPP CLI v2 reprint dogfood that share a root cause — the generator has no visibility into how the live API actually behaves at request time, only what the spec claims:

1. **Spec staleness drift goes undetected at generate-time.** CIPP-API has migrated from per-endpoint Azure Functions to a unified `ListGraphRequest` proxy. The OpenAPI spec the CLI is generated from still describes the older surface, so 14/83 endpoints (17%) are 404/deprecated/5xx/timeout on live CIPP. The CLI is correctly built from the spec but ships with a known-broken surface and no warning.

2. **PP-required tenant filter pattern not detected.** CIPP uses a global `TenantFilter` query parameter on 58/83 endpoints. The HTTP-level spec marks it optional, but the Printing Press has a hard requirement: bare multi-tenant queries blow past the 1MB MCP tool result limit on tenants with many users/mailboxes/devices, so every agent invocation MUST scope to one tenant. The spec doesn't express this; the generator has no way to know.

## What we observed

Dogfood persistently reported 39/332 failures across two full live-API runs. The same number recurred regardless of auth patches, base_url corrections, or env-var setup. The user pushed back and asked whether there was a fundamental misunderstanding of the API. Reframe: there were two.

**Finding 1 evidence** — probed all 83 spec endpoints against `cipptipwb.azurewebsites.net/api` with a valid bearer token and `TenantFilter=AllTenants`:

```
Probing 83 endpoints...
  data: 28          (returns real data)
  tiny: 40          (returns [] or {})
  5xx: 4            (ExecBECCheck, ListMailboxMobileDevices, ListSignIns, ListUserPhoto)
  timeout: 4        (ListInactiveAccounts, ListIntunePolicy, ListMFAUsers, ListSpamfilter)
  404: 3            (ListMFAUsersAllTenants, ListMailboxStatistics, ListPhishPolicies)
  400: 2            (ListExternalTenantInfo, ListSites)
  http-403: 1
  deprecated: 1     (ListUsers — explicitly says "use ListGraphRequest")
```

The `ListUsers` deprecated response is the smoking gun:

```json
[{"Message": "This function has been deprecated for all users, please use ListGraphRequest instead"}]
```

This is the same shape any API undergoing endpoint consolidation will produce. Stripe deprecations, Twilio v2 migrations, Notion API v3 — common pattern.

**Finding 2 evidence** — the user's reframe was decisive:

> "If the tools work with the tenant filter we need to ENFORCE the tenant filter for every tool call so that claude knows how to work with it. ... Our workflows will almost never be 'alltenant' work queries, but will always be tenant specific."

Re-probing the same endpoints with `TenantFilter=<specific-tenant-guid>` instead of `AllTenants` shifted the breakage profile substantially: timeouts disappeared, the "deprecated" ListUsers returned data, several "broken" endpoints worked. The spec described AllTenants as one valid value; the Printing Press needed it to be the exceptional case, not the default.

## Root cause

The generator's contract with reality is mediated entirely by the spec. When the spec is wrong about endpoint health (Finding 1) or about which params are *operationally* required even if HTTP-optional (Finding 2), the generator has no signal.

**Finding 1**: No live-endpoint validation step exists at generate-time. The generator emits endpoint commands from the spec without confirming the endpoints respond.

**Finding 2**: The spec-parser doesn't detect cross-endpoint query-param frequency. A param that appears on 70%+ of endpoints with the same name is signal that it's a global tenant/account/workspace scoping parameter that the operator needs to set once and have applied everywhere. The current generator treats each per-endpoint declaration in isolation.

## Cross-API check

**Finding 1 — spec staleness drift:**
- CIPP-API (verified): KelvinTegelaar/CIPP-API is moving to ListGraphRequest; spec describes older per-endpoint surface
- Stripe: deprecations move quickly; v2 endpoints replace v1
- Notion: API v3 surface differs significantly from v2 spec authoring
- Twilio: per-product API migrations (Conversations → Messaging, Programmable SMS → Messaging Service)

Any API with active deprecation will produce this drift. Frequency: any spec-based CLI generated from a vendor spec that pre-dates the live API. Probably most non-trivial APIs over time.

**Finding 2 — PP-required global scope param:**
- CIPP (verified): `TenantFilter` on 58/83 = 70% of endpoints
- Auth0 Management API: `tenant_id` on most endpoints
- AWS APIs: `Region` as a global scope param
- Stripe Connect: `Stripe-Account` header on most endpoints
- Microsoft Graph: `$select` / `$filter` aren't quite the same shape but `consistencyLevel` header has the global-param feel
- Any multi-tenant SaaS API: tenant identifier appears on most endpoints

Frequency: any per-tenant SaaS API. Probably 20%+ of catalog entries.

## Durable fix

**Finding 1 — spec staleness probe at generate-time:**

Add a `cli-printing-press probe-spec` subcommand (or fold into the existing reachability gate) that:
1. Reads a sample (say 5–10 random) endpoints from the resolved spec.
2. Hits each with the operator's bearer token + minimum required params (or with `TenantFilter=AllTenants` for CIPP-style, or with `--probe-fixtures path/to/fixtures.json` for custom).
3. Classifies each response: 200-data / 200-empty / 200-deprecated / 4xx / 5xx / timeout.
4. Reports a drift score: `<N>/<sample> endpoints respond differently than the spec describes`.
5. If drift > threshold (e.g., 25%), emit a warning at generate-time (or hard-fail with `--strict-spec-freshness`).
6. Optionally write `Known Gaps: <list>` into the generated README's `## Known Gaps` block based on drift findings.

This is similar in spirit to `Phase 1.9 Reachability Gate` but checks endpoint-level health, not just base-URL reachability.

**Finding 2 — global tenant-filter pattern detection in spec-parser:**

In `internal/openapi/parser.go` (or wherever the parser walks param declarations), add a cross-endpoint pass:
1. For each query parameter name `N`, count how many endpoints declare it.
2. When `N` appears on ≥50% (configurable threshold) of endpoints with the same case-insensitive name, mark it as a `global_scope_param`.
3. The generator emits this param:
   - As `required: true` on every per-endpoint flag (regardless of HTTP-level required)
   - With an env-var default: `<API>_<PARAM_UPPER>` (e.g., `CIPP_TENANT_FILTER` or `CIPP_TENANT`)
   - At the root persistent-flag level too, so the operator sees it as a top-level concept in `--help`
4. Surface this decision in the generated README's auth/usage section so the operator understands the enforcement.

CIPP's `TenantFilter` would be auto-detected. AWS APIs' `Region` would be auto-detected. Stripe Connect's `Stripe-Account` header would be auto-detected (extend the heuristic to headers when prevalence is high).

## Acceptance criteria

**Finding 1:**
- **Positive test**: generate a CLI from a spec describing 50 endpoints; 40 of those endpoints respond 200, 10 return 404 on the live API. The generate output includes a `Known Gaps` block listing the 10 dead endpoints in the README, and the shipcheck report has a `spec_freshness` field with the drift score.
- **Negative test**: generate from a fresh-vs-live-aligned spec (e.g., catalog entry that ships its own probe fixtures). No `Known Gaps` block emitted; drift score = 0.

**Finding 2:**
- **Positive test**: generate a CLI from a CIPP-shaped spec where `TenantFilter` appears on >50% of endpoints. The emitted commands have `--tenant-filter` marked required; CIPP_TENANT_FILTER (or CIPP_TENANT) env var is documented; bare `<cli> list-users` returns Cobra's native required-flag error.
- **Negative test**: generate from a spec where a query param appears on only 1 endpoint. No required-promotion, no env-var binding. Single-endpoint params stay as-emitted.

## Related issues

- #1158 — Generated config.go AuthHeader() (auth template area, different bug)
- #1383 — openapi-parser: relative-only servers URL (parser-level spec handling, different specific gap)
- #1408 — generator + spec-parser: per-endpoint requires_role declaration (similar shape: cross-endpoint annotation that should drive emit)
- #1435 — openapi-parser: prefer API canonical env-var (related env-var binding concern)
- #1652 — generator: mutation commands should auto write-back response to local store (different but related: generator-level workflow knowledge)

## Source

Retro doc: see `manuscripts/cipp/20260522-082135/proofs/20260522-102445-retro-cipp-pp-cli.md` in user's local printing-press install (5th finding, filed after re-probing the CIPP endpoint surface revealed the persistent dogfood-failure floor was structural, not a CLI bug).

User insight that surfaced Finding 2 verbatim:

> "If the tools work with the tenant filter we need to ENFORCE the tenant filter for every tool call so that claude knows how to work with it. Our workflows will almost never be 'alltenant' work queries, but will always be tenant specific."

This run's hand-applied workaround (sets the floor for what the durable fix should look like):
1. Modified spec to mark every `TenantFilter` parameter `required: true` (58 of them), regenerated. Cobra now rejects bare commands with "required flag tenant-filter not set" — Layer 1 enforcement.
2. Hand-wrote `internal/cli/cipp_root.go` that walks subcommands at wire time and pre-populates `--tenant-filter` from the `CIPP_TENANT` env var — Layer 2 default propagation.
3. Documented the 14 broken endpoints in README's `## Known Gaps`.

All three workarounds are CIPP-specific. The durable fixes above generalize them.

🤖 Filed via /printing-press-retro (5th finding, follow-up to the original 4)

tmchow:
## Summary

Promotes high-prevalence tenant/workspace/account-style query parameters into required, env-backed flags instead of filtering them out as generic globals. Printed CLIs now preserve CIPP-style `TenantFilter` as a command flag, default it from `CIPP_TENANT_FILTER`, and keep rejecting unscoped calls when neither the flag nor env var is present.

The existing noisy-global filter still drops non-scope parameters such as ubiquitous `limit` values, so broad optional pagination knobs do not become required.

Closes #1846

## Verification

- `go test ./internal/openapi -run 'TestPromoteGlobalScopeQueryParams' -count=1`
- `go test ./internal/generator -run 'TestGenerateGlobalScopeParamUsesEnvDefault' -count=1`
- `go test ./...`
- `go fmt ./...`
- `golangci-lint run ./...`
- `scripts/golden.sh verify`


tmchow:
@greptileai review

tmchow:
@greptileai review

PR #2359, opened 2026-05-26
fix(cli): emit happy args annotations

### LINKED ISSUE #2180: generator: emit pp:happy-args annotation on promoted commands from spec field
## Summary

The Printing Press runtime supports a `pp:happy-args` Cobra annotation for live-dogfood happy-path fixtures (`internal/pipeline/runtime_commands.go:happyArgsAnnotation`), but there is no spec field to declare those fixtures and no template emission for them on promoted-endpoint commands. The result: promoted commands whose endpoints require at-least-one-of conditional query params ship with broken live-dogfood happy-path coverage and no in-spec way to fix it. Half-built feature: runtime parser exists, propagation doesn't.

## Affected pattern

Any spec resource whose single promoted endpoint requires conditional query params produces this failure. Concrete subclasses:

- Annotation/referent lookup endpoints where the API requires one of `{song_id, web_page_id, created_by_id}` or similar parent-key alternatives
- Search endpoints where `q` (or equivalent) is required but not declared `positional`
- Single-endpoint resources for lookup-only flows (URL → canonical resource, slug → ID) where the input flag is required but not type-checked
- Auth-flow primitives (register/login/refresh) whose body or query params are conditionally required

The dogfood matrix probes the bare command first (no args) and classifies the resulting 4xx as `http_4xx`. There is no in-spec way to declare "use this fixture for the happy-path probe," so the failure ships uncorrectable in the current pipeline.

## Surfaced during

A reprint run against the Genius API — Phase 5 live dogfood: 6/96 failures, all `http_4xx`, all on promoted endpoints requiring conditional query params:

- `referents` happy_path + json_fidelity → HTTP 422 (`/referents` requires `song_id`, `web_page_id`, or `created_by_id`)
- `web-pages` happy_path + json_fidelity → HTTP 404 (`/web_pages/lookup` requires one of `--raw-annotatable-url`/`--canonical-url`/`--og-url`)
- `sync` happy_path + json_fidelity → collateral (default sync walks the `referents` resource and inherits the 422)

The prior version of the CLI shipped with the same six failures — this is not a regression, it is a persistent matrix gap with no in-spec fix surface. Hand-patching the generated `promoted_*.go` to add the annotation surfaces as `TEMPLATED-WITH-ADDITIONS` at the next `regen-merge` and halts promote, so the per-CLI workaround is also not durable.

## Proposed fix

1. Add `HappyArgs string` to `internal/spec/spec.go` `Endpoint` struct, marshaled as `happy_args` in YAML. Tie through OpenAPI's `x-happy-args` vendor extension via `internal/openapi/parser.go`.
2. Update the promoted-command template in `internal/generator/templates/` to emit the annotation when populated:
   ```go
   Annotations: map[string]string{"mcp:read-only": "true", "pp:happy-args": "<value>"}
   ```
3. Optionally extend to non-promoted typed endpoints if they are also subject to dogfood matrix probing.

Spec YAML shape:

```yaml
resources:
  referents:
    endpoints:
      list:
        path: "/referents"
        happy_args: "--song-id=378195"
```

The runtime annotation parser already handles `<positional>=value;--flag=value` semicolon-separated tokens — no runtime changes needed.

## Acceptance

- **Positive:** add a `happy_args` line to any endpoint requiring conditional input, regenerate, run `cli-printing-press dogfood --live`; the corresponding happy_path no longer 4xx-fails.
- **Negative:** regenerate a CLI without `happy_args` populated; the generated `promoted_*.go` continues to emit only `mcp:read-only`, no behavior change. No guard needed — `happy_args` is opt-in metadata.

## Complexity

Small. Spec field + parser pass-through + one template literal. Runtime already exists.

## Related issues

- #1209 — generator: resource parents shouldn't be Hidden:true without a promoted-command alias *(adjacent: same template surface, different specific fix)*
- #1289 — Required-body-flag commands error on empty invocation instead of printing help *(adjacent: same matrix-coverage failure shape, different endpoint class — body-required vs query-conditional)*

tmchow:
## Summary

Specs can now declare live-dogfood happy-path fixtures for endpoints that cannot be probed with generic synthesized args. `happy_args` in internal specs and `x-happy-args` in OpenAPI now flow into generated Cobra `pp:happy-args` annotations on promoted commands and normal typed endpoint commands, while endpoints without fixtures keep the existing annotation shape.

Closes #2180

## Verification

- `go test ./internal/openapi -run 'TestParseHappyArgsExtension|TestParseDataSourceStrategyExtension'`
- `go test ./internal/generator -run 'TestHappyArgsAnnotationEmittedFromSpec|TestRequiredFlagCommands_HelpFallbackGatedToRequiredInput'`
- `scripts/golden.sh verify`
- `go test ./...`
- `golangci-lint run ./...`


tmchow:
@greptileai review

PR #2382, opened 2026-05-26
docs(skills): document reliable HAR body capture

### LINKED ISSUE #2141: skill: document reliable Chrome HAR body capture (disable-cache hard-reload / Firefox) in browser-sniff-capture.md
**Component:** `skill` — `skills/printing-press/references/browser-sniff-capture.md` (manual-HAR guidance)
**Priority:** P3
**Source:** retro of a `/printing-press` run that relied on user-provided HAR fixtures.

## What happened
Across two large user-provided Chrome HAR exports ("Save all as HAR with content"), nearly every site HTML page had an **empty response body** — they were `206`/cached responses with no stored `content.text`. Only ~1 usable body landed per export, forcing three capture rounds and leaving per-page-type fixtures incomplete.

## Root cause
`browser-sniff-capture.md` documents the manual-HAR path but does not warn that Chrome's HAR-with-content export omits bodies served from disk cache (the `206` partial responses), nor recommend the reliable capture method.

## Durable fix
Add a one-paragraph note to the manual-HAR section: to guarantee response bodies are captured, check **Disable cache** in DevTools → Network and **hard-reload** each page (so responses are full `200`s, not cached `206`s) before "Save all as HAR with content"; or use Firefox, whose HAR export reliably includes bodies. Symptom to call out: `206` entries with empty `content.text`.

## Why this is worth a machine fix
Recurs on **every** Chrome HAR-based discovery run, independent of the target API; the cost is multiple wasted capture rounds. Cheap doc fix.

tmchow:
## Summary
Manual HAR guidance now tells agents how to recover when Chrome exports cached `206` responses with empty bodies. Agents can point users to Disable cache plus hard reload, or ask for Firefox when Chrome still omits `response.content.text`, instead of repeating failed capture rounds.

Replaces #2344 after GitHub did not create the required pull_request workflow runs for its rebased head.

Closes #2141.

## Tests
- `go test ./internal/spec ./internal/pipeline`

---

GPT--5-Codex

### DIFF
--- a/internal/pipeline/contracts_test.go
+++ b/internal/pipeline/contracts_test.go
@@ -354,6 +354,15 @@ func TestBrowserSniffEscalates200ChallengeShells(t *testing.T) {
 	assert.Contains(t, capture, "HTTP 200 challenge shells or truncations")
 }
 
+func TestBrowserSniffManualHARGuidesReliableBodyCapture(t *testing.T) {
+	capture := readContractFile(t, filepath.Join("..", "..", "skills", "printing-press", "references", "browser-sniff-capture.md"))
+
+	assert.Contains(t, capture, "Chrome can export page responses from disk cache as `206` partial-content entries with empty `response.content.text`")
+	assert.Contains(t, capture, "check **Disable cache**, then hard-reload each page while DevTools stays open before exporting the HAR")
+	assert.Contains(t, capture, "ask for a Firefox HAR export instead")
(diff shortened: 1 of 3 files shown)


tmchow:
Closing duplicate test PR. GitHub Actions also did not create pull_request workflow runs here, so keeping #2344 as the active PR.

PR #2443, opened 2026-05-28
fix(cli): prevent parse panics and propagate injectAddCommands errors

cropsgg:
## Intent

Issue:

## Approach

fix(cli): 
- harden parseStmt and injectAddCommands against panic and silent data loss parseStmt (ast_inject.go)
: - Replace unchecked Decls[0] type assertion and Body.List[0] array access with   a bounds-safe loop that returns an explicit error when the parsed function has   an empty body. Mirrors the already-correct pattern in regenmerge/Prettiest
. - Empty string and whitespace-only inputs now return an error instead of panicking. - Add TestParseStmt_EmptyStringReturnsError and TestParseStmt_WhitespaceOnlyReturnsError   to pin the empty-body guard against regression.  injectAddCommands (regenmerge/apply.go): 
- parseStmtViaDST errors were silently swallowed with continue, leaving the host   file with fewer AddCommand registrations than intended 
— a build-invisible   regression. Now captured via closure var (injectErr) and returned to the caller
. - Add TestInjectAddCommandsMalformedCallPropagatesError to cover the new path.

Now shorten this for a title
## Scope

Primary area:

Why this belongs in this repo:

## Catalog Justification

Embedded catalog fit:

Distinct blueprint pattern:

Closest existing entries checked:

Source provenance:

Auth and tenant assumptions:

Safe default surface:

Generation path:

Stale-body check:

## Risk

N/A

## Output Contract

N/A

## Verification

## AI / Automation Disclosure

- [ ] some AI or automation was used
- [ ] AI-reviewed only: an AI agent reviewed the work and human reviewed it before submission

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ catalog/specs/*.graphql
 .claude/local/
 /refactor/artifacts/
 /tests/artifacts/perf/
+/cli-printing-press
 /printing-press
 /press-auth
 library/
--- a/internal/patch/ast_inject.go
+++ b/internal/patch/ast_inject.go
@@ -40,11 +40,27 @@ func injectRootAST(src []byte, opts injectOptions) ([]byte, bool, error) {
 	if !opts.skip("deliver") {
 		changed = addImports(file, "bytes", "io", "os") || changed
 	}
-	changed = addPersistentFlags(file, opts) || changed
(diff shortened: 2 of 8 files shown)


tmchow:
This issue description is slop. Sorry. Please ensure an issue is filed that it references and the PR description is actually coherent and follows the template.

PR #2456, opened 2026-05-28
fix(cli): pass context to async job no-cache polling

### LINKED ISSUE #2450: bug(cli): GetNoCache signature mismatch — jobs.go.tmpl calls 2-arg form, client.go.tmpl defines 3-arg form
## Repro

```
printing-press generate --spec https://api.polygon.io/openapi --name polygon-v2
```

Generation fails at the `govulncheck` quality gate:

```
C:\Users\pceci\printing-press\library\polygon-v2\internal\cli\jobs.go:150:38:
not enough arguments in call to c.GetNoCache
        have (string, nil)
        want (context.Context, string, map[string]string)
```

## Cause

Two templates are out of sync:

- `internal/generator/templates/client.go.tmpl:565` — defines `func (c *Client) GetNoCache(ctx context.Context, path string, params map[string]string)` (3 args)
- `internal/generator/templates/jobs.go.tmpl:149` — calls `c.GetNoCache(path, nil)` (2 args, old signature)
- `internal/generator/templates/live_ws.go.tmpl:175` — calls `apiClient.GetNoCache(ctx, "...", params)` (correct 3-arg form)

`jobs.go.tmpl` appears to have been missed when the `GetNoCache` signature was widened to accept `ctx` and `params`. This is the same line region that #1136 added to fix the async-poll cache-locking bug — `GetNoCache` is precisely what that fix introduced, so the regression sits on top of the original fix.

## Likely fix

In `jobs.go.tmpl:149`, change:

```go
resp, err := c.GetNoCache(path, nil)
```

to:

```go
resp, err := c.GetNoCache(ctx, path, nil)
```

`ctx` is already in scope at this point in the template (the surrounding loop already calls `ctx.Err()` on the previous line).

## Impact

Any spec whose endpoints are classified as long-running jobs (Polygon has several) currently can't be generated at v4.19.0 — govulncheck gate fails before downstream gates run. Other gates pass cleanly on this spec, so this is a one-line template fix away from passing.

## Environment

- printing-press v4.19.0 (`186e459e`)
- Windows 10 / Go 1.26.3
- Spec: https://api.polygon.io/openapi

### LINKED ISSUE #2458: v4.19.0: generated jobs.go calls c.GetNoCache(path, nil) — missing ctx arg, fails to compile
## Summary

v4.19.0 generates `internal/cli/jobs.go` whose job-polling helper calls
`c.GetNoCache(path, nil)`, but the generated client's signature is
`GetNoCache(ctx context.Context, path string, params map[string]string)`. The
generated project therefore **does not compile**, and the post-generation build
gate fails.

## Error

```
internal/cli/jobs.go:150:38: not enough arguments in call to c.GetNoCache
	have (string, nil)
	want (context.Context, string, map[string]string)
```

(Also surfaced via the generation gate: `validating generated project: gate "govulncheck ./..." failed`.)

## Offending generated code

```go
// inside the job-wait loop — note ctx is already in scope (used by ctx.Err() just above)
if err := ctx.Err(); err != nil {
    return nil, err
}
// GetNoCache: cache is keyed by (path, params) ...
resp, err := c.GetNoCache(path, nil)        // <-- missing the leading ctx argument
```

The fix is to pass the in-scope `ctx` as the first argument, matching the client
the generator also emits:

```go
resp, err := c.GetNoCache(ctx, path, nil)
```

## Repro

```
cli-printing-press generate --spec <openapi.json> --lenient --output /tmp/fresh
cd /tmp/fresh && go build ./...   # fails at internal/cli/jobs.go:150
```

cli-printing-press **v4.19.0**, Go 1.26.3, `modernc.org/sqlite`. Observed against
a BlueCat Address Manager v2 spec, but the call site is template-driven and
spec-independent.

## Suggested fix

The job-polling template should thread the in-scope `ctx` into `GetNoCache`
(`c.GetNoCache(ctx, path, nil)`) to match the generated client signature.

## Notes

- `jobs.go` is `DO NOT EDIT` generated code, so it can't be patched downstream.
- Distinct from #2457 (store typed-table `parent_id`); this one blocks the build
  outright for any spec.

ishaans:
## Intent

Fixes #2450 
Fixes #2458

Generated async job polling currently calls `GetNoCache` with the old two-argument shape even though the generated client now requires a `context.Context`:

```go
c.GetNoCache(path, nil)
```

Without this fix, async-job APIs fail generated-project validation before downstream gates can run.

## Approach

Thread the existing `ctx` already in scope inside `WaitForJob` into the no-cache polling call. Update the generator assertion and async-job golden fixture so the emitted-code contract locks onto the context-aware call shape.

## Scope

Primary area: generator templates and generated-output fixtures.

Why this belongs in this repo:

This is a machine change. The mismatch is emitted by `internal/generator/templates/jobs.go.tmpl` and affects every future generated CLI whose spec triggers async job polling.

## Catalog Justification

Embedded catalog fit: N/A

Distinct blueprint pattern: N/A

Closest existing entries checked: N/A

Source provenance: N/A

Auth and tenant assumptions: N/A

Safe default surface: N/A

Generation path: N/A

Stale-body check: N/A

## Risk

Low. `WaitForJob` already accepts `ctx` and checks `ctx.Err()` immediately before this call. The generated client already exposes the context-aware `GetNoCache` method. This change only makes the emitted call match the emitted method signature.

## Output Contract

Generated `internal/cli/jobs.go` for async-job CLIs now emits:

```go
resp, err := c.GetNoCache(ctx, path, nil)
```

instead of:

```go
resp, err := c.GetNoCache(path, nil)
```

Generated-output evidence:

- `TestGenerateWaitForJobBypassesResponseCache` now asserts the context-aware call.
- `generate-async-job-api` golden output is updated.
- `scripts/verify-generator-output.sh generate-async-job-api` compiles the affected generated module.

## Verification

Commands run:

```bash
go test ./...
scripts/golden.sh verify
scripts/verify-generator-output.sh generate-async-job-api
CLI_PRINTING_PRESS_BIN=/private/tmp/cli-printing-press-main-check/cli-printing-press npm run cli:pp:check
```

The final command was an additional downstream OpenAPI-generation smoke test run against a patched local `cli-printing-press` binary. It passed generation, generated CLI tests, CLI/MCP builds, `--help`, `doctor --json`, and `agent-context --json`.

## AI / Automation Disclosure

### DIFF
--- a/internal/generator/generator_test.go
+++ b/internal/generator/generator_test.go
@@ -14391,7 +14391,7 @@ func TestGenerateWaitForJobBypassesResponseCache(t *testing.T) {
 	require.NoError(t, err)
 	jobsBody := string(jobsData)
 
-	assert.Contains(t, jobsBody, "c.GetNoCache(path, nil)",
+	assert.Contains(t, jobsBody, "c.GetNoCache(ctx, path, nil)",
 		"WaitForJob must poll with GetNoCache to avoid locking on the cached initial response")
 	assert.NotContains(t, jobsBody, "c.Get(ctx, path, nil)",
 		"WaitForJob must not call c.Get(ctx, path, nil); cached non-terminal status would lock the poll")
--- a/internal/generator/templates/jobs.go.tmpl
+++ b/internal/generator/templates/jobs.go.tmpl
@@ -146,7 +146,7 @@ func WaitForJob(ctx context.Context, c *client.Client, statusPath string, jobID 
 		// GetNoCache: the cache is keyed by (path, params), so a cached
 		// non-terminal status would lock the poll on the initial response
 		// for the cache TTL.
-		resp, err := c.GetNoCache(path, nil)
+		resp, err := c.GetNoCache(ctx, path, nil)
 		if err == nil {
 			var body map[string]any
 			if uerr := json.Unmarshal(resp, &body); uerr == nil {
--- a/testdata/golden/expected/generate-async-job-api/async-job-api/internal/cli/jobs.go
+++ b/testdata/golden/expected/generate-async-job-api/async-job-api/internal/cli/jobs.go
@@ -147,7 +147,7 @@ func WaitForJob(ctx context.Context, c *client.Client, statusPath string, jobID 
 		// GetNoCache: the cache is keyed by (path, params), so a cached
 		// non-terminal status would lock the poll on the initial response
 		// for the cache TTL.
-		resp, err := c.GetNoCache(path, nil)
+		resp, err := c.GetNoCache(ctx, path, nil)
 		if err == nil {
 			var body map[string]any
 			if uerr := json.Unmarshal(resp, &body); uerr == nil {


tmchow:
CI green, Greptile 5/5, no open threads. Marking ready-to-merge so Mergify can queue it.

PR #2492, opened 2026-05-30
refactor(printing-press): lazy-load phase bodies to halve invocation-time context

melanson633:
## What

Splits `skills/printing-press/SKILL.md` into a **lean dispatcher + per-phase reference files**, extending the skill's existing `Read and apply references/X.md` idiom. The full body currently loads on every `/printing-press` invocation; this defers ~2,280 lines of procedural phase prose into `references/` that load only when their phase actually runs.

## Measured impact

| | lines | bytes |
|---|---|---|
| **Before** (loads every invocation) | 4,484 | 300,699 |
| **After** (dispatcher) | 2,200 | 149,395 |
| **Delta** | **−51%** | **−50.3%** |

Cold-start cost is unchanged (frontmatter `name`+`description` only). The deferred phase bodies (~2,280 lines across 15 files) load lazily — a typical run touches only a handful.

## Design principle: gates stay inline, bodies defer

The skill's whole design is forcing-functions around expensive/consent steps, so **every enforcement fragment stays resident** in the dispatcher; only procedural prose moved. Verified still inline:

- **1.7 Browser-Sniff Gate** — marker WRITE contract (`browser-browser-sniff-gate.json`), the four decision values (`approved`/`declined`/`skip-silent`/`pre-approved`), every-path-writes-an-entry, per-source one-entry rule, before-leaving exit count-check, banned-skip-reasons
- **1.5 Ecosystem Absorb Gate** — marker READ hard-stop (both halt messages), always-spawn-novel-features-subagent mandate, WAIT-for-approval, priority-inversion HALT, Phase Gate 1.5
- **Phase 1** ordering mandate (evaluate 1.6/1.7/1.8 before 1.5) + marker pre-announcement
- **1.9** reachability decision matrix + `probe-reachability` "don't present transport tiers as a peer menu" rule
- **Phase 2** Lock-and-Generate invocation + load-bearing flags (`--client-pattern proxy-envelope`, `--lenient`, `--category`)
- **Phase 3** Completion Gate — per-row Cobra resolution check, `skipped:true`=HALT, HALT-on-any-miss
- **Phase 5.6** — acceptance gate, Path A/B routing, regen-merge review-required halt verdicts, PII-ledger ordering, proceed-to-Phase-6 tail
- **Phase 6** not-optional + verdict precedence + do-not-improvise-publish
- Full Preflight bash contract (`PRESS_SETUP_CONTRACT_*`) + absolute-path rule (`PRINTING_PRESS_BIN=`)

## Fidelity

All 15 files extracted **verbatim** (`sed` slices, not paraphrase). Each block's first and last content lines were verified byte-identical to the pre-split body. All stub→reference pointers resolve.

## New files

`phase-0-resolve-and-reuse`, `phase-1-research-and-brief`, `phase-1.5-search-and-manifest`, `phase-1.5-research-json`, `phase-1.6-auth-classification`, `phase-1.7-reachability-modes`, `phase-1.9-oauth-probe`, `phase-2-enrichments`, `phase-3-build-checklist`, `phase-3-build-guidance`, `shipcheck-reviews`, `phase-5.6-promote-detection`, `phase-5.6-archive`, `phase-6-menus`, `fast-guidance`.

## Note for the maintainer

Left untouched to keep this a pure split: the marker filename is the doubled-prefix `browser-browser-sniff-gate.json` (7 inline sites). Happy to fold a `browser-sniff-gate.json` rename into this PR or a follow-up — your call.


tmchow:
@melanson633 how much testing did you do on this vs. having your agent just do the work and trust the vibes?  Labeling this as a 'chore' also seems crazy given how much code churn there is.

melanson633:
Somewhere between the two but not enough testing tbh and should have put this as issue first. That said, the cold start massive context load on invocation I do feel is real issue on non-trivial CLI generations when no public API. I set my default auto compact to 165k tokens (don't trust judgement after 200k regardless of window)and that gets eaten up before phase 1 completes. A leaner Skill.md with routing to reference files just seems to make more sense to me. 
Will be using the proposed Skill structure this week locally so happy to circle back with more concrete evidence. 
Any feedback or thoughts appreciated, I am still very much learning as I go :)

tmchow:
> Somewhere between the two but not enough testing tbh and should have put this as issue first. That said, the cold start massive context load on invocation I do feel is real issue on non-trivial CLI generations when no public API. I set my default auto compact to 165k tokens (don't trust judgement after 200k regardless of window)and that gets eaten up before phase 1 completes. A leaner Skill.md with routing to reference files just seems to make more sense to me. Will be using the proposed Skill structure this week locally so happy to circle back with more concrete evidence. Any feedback or thoughts appreciated, I am still very much learning as I go :)

Can you rebase against main and give it more testing? i'll do the same after you rebase.

tmchow:
@melanson633 — I ran this branch end-to-end against a real target to see how the lazy-load split behaves in practice: loaded the skill via `--plugin-dir` and generated a from-website CLI (OfferUp).

The headline: the agent loaded **none** of the seven phase reference files. Not one. The dispatcher pointed at every one of them — grepping the run transcript, `references/phase-2-enrichments.md` appears 4× in context, and the phase-3 build refs, phase-5.6, phase-6-menus, and shipcheck-reviews all show up too.  ut there's not a single corresponding `Read`. Every phase ran off the lean SKILL.md summary plus the model's own judgment. The only reference that reliably loaded was `setup-checks.md`, which is the one you already gate as mandatory.

It didn't blow up, and that's the problem: it degraded quietly. The CLI still shipped, but at B-grade with the entire auth surface bolted on *after* generation, which is the exact thing `phase-2-enrichments.md` warns against. With the detail moved out of SKILL.md, the dispatcher kept just enough inline skeleton that skipping the reference felt fine, and nothing downstream forced the issue. The binary's quality gates pass either way, so there's no signal that the judgment layer got skipped.

Separately, the branch ships with tests red. Five `TestPrintingPressSkill*` contract tests in `internal/pipeline/contracts_test.go` still read `SKILL.md` for blocks the refactor relocated into the reference files (`### Pre-Generation Auth Enrichment`, the RunE template, scan-and-filter caps, the multi-spec guard, category enrichment), so they fail with "missing start marker". `go test ./...` is red on the branch as-is.

I pushed two commits:

- **`test(cli): repoint printing-press skill contract tests to phase references`** — points those five tests at the reference files where the content now lives. Assertions are unchanged and pass against the relocated content, which also confirms the move was verbatim.
- **`docs(skills): stub Pre-Generation Enrichments to force reference load`** — collapses that section to a hard STOP stub that names the four decisions (category / cache / auth / MCP) but carries no inline how-to, so there's no skeleton to coast on and the reference has to be read before Lock and Generate.

Please run `go test ./...` — it was red before these two commits. And one thing worth deciding before this lands: the stub is a nudge, not a guarantee. A self-reported marker can be faked; the only mechanism that actually forces a load is a pre-tool hook that checks the transcript for the `Read`. Worth thinking about how hard you want to gate the pre-generation references, since the silent-skip is the real risk this split introduces.

This has to be tested way more because the CLI printing in my run took nearly twice as long as it should have because it made mistakes that would have been avoided if the skill was correctly followed

tmchow:
I played with this more even after more of my changes and this refactor isn't working. it's making the quality worse. 

I do want to make the skill leaner in many ways to help but this approach isn't going to do it unfortunately.

melanson633:
> unfortunately.

Understood, appreciate you looking at it. Will come back with a better approach more rigorously tested at some point I'm sure.

PR #2520, opened 2026-05-31
fix(cli): always emit RequestBaseURL() so novel commands keep BasePath

### LINKED ISSUE #2519: fix(cli): always emit RequestBaseURL() so novel commands don't drop BasePath
## Problem

Printed CLIs for APIs with a `BasePath` (e.g. Immich, whose API lives under `/api`) trap hand-written novel commands into open-coding `c.BaseURL + c.BasePath + path` when building request URLs. The generated client *has* a correct helper, `RequestBaseURL()`, that encapsulates `BaseURL + BasePath` (and tier routing) — but it is emitted **only when `.HasHTMLExtraction` is true** (`internal/generator/templates/client.go.tmpl:184` and `:262`).

For a CLI that doesn't trigger HTML-extraction (like Immich), `RequestBaseURL()` is never emitted, so a novel-command author has no canonical safe accessor and must concatenate by hand. Forgetting `BasePath` is the easy mistake.

### Observed in the wild (Immich CLI)

`grab.go`'s `streamDownload` built the download URL as `c.BaseURL + path`, dropping `/api`. Immich's web server answered the missing route with its SvelteKit SPA `index.html` (HTTP 200), so every downloaded `.JPG` was a byte-identical 6903-byte HTML page instead of a photo. Every other (client-routed) request worked because the client adds `BasePath`.

This is a **latent trap in every printed CLI that has a `BasePath` and any hand-written URL building**, not an Immich-specific issue.

## Fix

Emit `RequestBaseURL()` **unconditionally**, decoupled from `.HasHTMLExtraction`:
- `HasTierRouting` CLIs delegate to the existing `baseURLForRequest()`.
- Non-tier CLIs return `c.BaseURL` (+ `c.BasePath` when present).

Keep `LastContentType()` / `lastContentType` gated on `.HasHTMLExtraction` (genuinely HTML-extraction concerns) by splitting them out of the bundled block.

## Scope / verification
- `internal/generator/templates/client.go.tmpl` only.
- `go test ./...`, `scripts/golden.sh verify`, `scripts/verify-generator-output.sh`.
- Add generated-output coverage asserting `RequestBaseURL()` is emitted for a non-HTML-extraction CLI and returns `BaseURL+BasePath`.

## Out of scope (separate machine bugs, observed in same session)
- `/assets/random` selected as a paginated sync resource → infinite `sync_progress` loop.
- `/view/folder/unique-paths` selected as a syncable list resource → all-items-failed-id-extraction anomaly.

The `text/html` content guard in `streamDownload` is genuine novel-command logic and stays a printed-CLI fix.

restagner:
## Intent

Printed CLIs for APIs with a `BasePath` trap hand-written novel commands into open-coding `c.BaseURL + c.BasePath` when building request URLs, because the safe accessor (`RequestBaseURL()`) was emitted only for HTML-extraction CLIs. Forgetting `BasePath` there is the easy mistake.

Observed in the Immich CLI: `grab`'s `streamDownload` built `c.BaseURL + path`, dropped `/api`, and got Immich's SvelteKit SPA `index.html` (HTTP 200) written into every downloaded image file.

Issue: Closes #2519

## Approach

Emit `RequestBaseURL()` **unconditionally**, decoupled from `.HasHTMLExtraction`:
- Tier-routing CLIs delegate to the existing `baseURLForRequest()`.
- Non-tier CLIs return `c.BaseURL` (+ `c.BasePath` when present).

`LastContentType()` / `lastContentType` stay gated on `.HasHTMLExtraction` (genuinely HTML-extraction concerns) by splitting them out of the previously-bundled block. The non-tier doc comment is conditional on `.BasePath` so the no-BasePath client stays free of the word "BasePath" (keeps `TestClientWithoutBasePathByteIdentical` honest).

Now every printed CLI exposes one canonical accessor for the request base, so novel commands cannot drop `BasePath` by hand.

## Scope

Primary area: generator template (`internal/generator/templates/client.go.tmpl`).

Why this belongs in this repo: this is a machine change — the symptom surfaced in a printed CLI, but the cause is that the generator only emitted the safe URL-building helper for one CLI variant. The fix makes `RequestBaseURL()` available in every printed client, so the whole class of "novel command forgot BasePath" bugs is prevented at generation time rather than patched per CLI.

## Catalog Justification

N/A

## Risk

Low. The only generated-output change is the addition of the `RequestBaseURL()` accessor (and a relocation of `LastContentType` within its conditional). Golden diffs show exactly the new helper and nothing else across all five client.go fixtures; generated-output verification confirms every auth fork still compiles. No MCP, auth, catalog, publish, scorer, or release surface is touched. A CLI that already had `RequestBaseURL()` (HTML-extraction) is unchanged.

## Output Contract

Generator/template change. Evidence:
- Golden fixtures updated for 5 `client.go` cases (each adds only the new accessor); `scripts/golden.sh verify` passes.
- `scripts/verify-generator-output.sh` passes — all 5 forks tidy + build.
- New emitted-code assertions in `internal/generator/client_basepath_test.go` for both the BasePath path (`return c.BaseURL + c.BasePath`) and the no-BasePath path (`return c.BaseURL`); the existing live-request test already proves runtime routing.

## Verification

- `go test ./...` — pass
- `scripts/golden.sh verify` — pass (25 cases)
- `scripts/verify-generator-output.sh` — pass (5 cases)
- Rebuilt the local Immich CLI against this helper (`c.RequestBaseURL() + path` in `streamDownload`); `grab` now writes real JPEGs instead of the SPA HTML.

### DIFF
--- a/internal/generator/client_basepath_test.go
+++ b/internal/generator/client_basepath_test.go
@@ -46,6 +46,13 @@ func TestClientHonorsSpecBasePath(t *testing.T) {
 	assert.Contains(t, client, "c.BaseURL + c.BasePath + path",
 		"do() should construct request URLs as BaseURL+BasePath+path")
 
+	// RequestBaseURL() must be emitted for every printed CLI (not just HTML-
+	// extraction ones) so novel commands have a safe accessor and cannot drop
+	// BasePath by open-coding c.BaseURL. This spec has no HTML extraction.
+	requestBaseURL := regexp.MustCompile(`func \(c \*Client\) RequestBaseURL\(\) string \{\s*return c\.BaseURL \+ c\.BasePath\s*\}`)
+	assert.Regexp(t, requestBaseURL, client,
+		"RequestBaseURL() should be emitted and return BaseURL+BasePath even without HTML extraction")
+
 	cacheKeyBody := clientCacheKeyBody(t, client)
 	assert.Contains(t, cacheKeyBody, `"|base_path=" + c.BasePath`,
 		"cache key should include BasePath so a config change invalidates correctly")
@@ -82,6 +89,11 @@ func TestClientWithoutBasePathByteIdentical(t *testing.T) {
 		"client.go must not emit BasePath when the spec doesn't declare base_path")
 	assert.NotContains(t, client, "normalizeBasePath",
 		"client.go must not emit the normalizeBasePath helper when unused")
+	// RequestBaseURL() is still emitted (returning just c.BaseURL) so novel
+	// commands have a consistent accessor regardless of BasePath.
+	requestBaseURL := regexp.MustCompile(`func \(c \*Client\) RequestBaseURL\(\) string \{\s*return c\.BaseURL\s*\}`)
+	assert.Regexp(t, requestBaseURL, client,
+		"RequestBaseURL() should be emitted and return c.BaseURL when there is no BasePath")
 
 	configSrc, err := os.ReadFile(filepath.Join(outputDir, "internal", "config", "config.go"))
 	require.NoError(t, err)
--- a/internal/generator/templates/client.go.tmpl
+++ b/internal/generator/templates/client.go.tmpl
@@ -181,9 +181,10 @@ func (c *Client) baseURLForRequest() string {
 	}
 }
 
-// RequestBaseURL returns the base URL used for the current request, including
-// any BasePath. Novel commands that build request URLs by hand should use this
-// instead of concatenating c.BaseURL directly, so they cannot drop BasePath.
+// RequestBaseURL returns the base URL used for the current request (including
+// BasePath for tiers that use the default base URL). Novel commands that build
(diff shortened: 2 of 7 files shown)


restagner:
@tmchow  All checks are green and the review thread is resolved — ready for approval/merge.

- Addressed the Greptile P2 nit on the tier-routing `RequestBaseURL()` doc comment (it overstated that BasePath is always folded in). @trevinchow applied the suggested wording via a suggestion-commit; that web edit updated `client.go.tmpl` but couldn't regenerate the golden, so I pushed a follow-up regenerating the `generate-tier-routing-api` fixture to match. `scripts/golden.sh verify` now passes.
- The remaining `Mergify Merge Protections` red is just the pending maintainer-approval gate, not a CI/build failure — `full-test`, `full-golden`, `generated-output-compile`, and lint are all green.

Ready for your approval and the `ready-to-merge` label whenever you'd like to queue it.

Issue #910, opened 2026-05-10
GraphQL-aware client emission is half-implemented for /graphql-only specs


kylelol:
> Filed by `/printing-press-retro` from a run of **Monarch Money**
> (scorecard 80/100 → 83/100 after polish, 100% verify pass).

## Summary

The generator detects GraphQL-shaped specs (every endpoint POSTs to a single shared path) and emits a sophisticated `internal/client/graphql.go` helper with `Query()`, `Mutate()`, `PaginatedQuery()`, and typed `GraphQLAccessDeniedError`. But the wiring stops short: `graphqlEndpointPath` is left empty, `queries.go` emits unparseable placeholder GraphQL, and per-endpoint command templates continue to call `c.Post(path, body)` with an empty body instead of `c.Query(QueryConstant, vars)`. The result is half-working scaffolding that passes `verify` (because dry-run skips the network) but breaks instantly under live invocation.

## Where to look

- **Component:** generator
- **Likely area:** Three template locations:
  - `internal/generator/templates/client/graphql.go.tmpl` — `graphqlEndpointPath` const wiring
  - `internal/generator/templates/client/queries.go.tmpl` — placeholder query body
  - Per-endpoint command template family — `c.Post()` vs `c.Query()` selection
  - Plus the spec parser (`internal/spec/` and `internal/openapi/`) needs a GraphQL-shape detector that sets `APISpec.GraphQLEndpointPath`.
- **Triggered when:** ≥80% of spec endpoints share a single `path` with `method: POST` (GraphQL-only spec)

## What we observed

In the Monarch run, all three locations needed manual fixes:

**1. `graphqlEndpointPath` left empty** — every `c.Query()` call would have POSTed to BaseURL + "" instead of BaseURL + "/graphql":

```go
// internal/client/graphql.go:60
const graphqlEndpointPath = ""

// internal/client/graphql.go:64-67
func (c *Client) Query(query string, variables map[string]any) (json.RawMessage, error) {
    req := GraphQLRequest{Query: query, Variables: variables}
    raw, _, err := c.Post(graphqlEndpointPath, req)  // <-- POSTs to ""
    // ...
}
```

**2. `queries.go` emits unparseable GraphQL.** The placeholder body has a missing field name after `(`:

```go
// internal/client/queries.go:8-15 (verbatim from generator output)
const AccountsListQuery = `query($first: Int!, $after: String) {
  (first: $first, after: $after) {            // <-- missing field name
    nodes {
      id
[...]

Issue #669, opened 2026-05-07
generator emits REST sync template for GraphQL specs (incremental: emit TODO scaffold instead)


tmchow:
> Filed by `/printing-press-retro` from a run of **Linear** (scorecard 86/100, 100% verify pass).

## Summary

The v3.10.0 generator runs the same `sync.go.tmpl` and `client.go.tmpl` for GraphQL specs as for REST OpenAPI specs. The emitted `internal/cli/sync.go` for a GraphQL CLI is REST-shaped — it tries to GET `/graphql` instead of POSTing GraphQL queries — and silently passes the 7 quality gates (compiles, runs, returns help) while being functionally broken. Today's Linear reprint had to discard the v3 generated sync entirely and carry forward v1's hand-written `internal/client/graphql.go` + `internal/cli/sync.go`. The full fix (real GraphQL client + sync templates from parsed schema) is the same WU-3 from the v1 Linear retro (#182, closed unimplemented). This re-raise proposes the **incremental step**: emit a TODO scaffold instead of broken REST when the parser detects a GraphQL spec.

## Where to look

- **Component:** `comp:generator`
- **Likely area:**
  - `internal/generator/generator.go` — derive `IsGraphQL bool` from spec metadata.
  - `internal/generator/templates/sync.go.tmpl` — gate REST emit on `{{- if not .IsGraphQL }}`.
  - `internal/generator/templates/client.go.tmpl` — same gating shape.
- **Triggered when:** spec source is `graphql-sdl` OR `spec.BaseURL` ends in `/graphql` OR equivalent.

## What we observed

Today's Linear reprint:

```bash
$ printing-press generate --spec linear-schema.graphql --name linear --output ... --force --validate
PASS go mod tidy
PASS go vet ./...
PASS go build ./...
PASS build runnable binary
PASS linear-pp-cli --help
PASS linear-pp-cli version
PASS linear-pp-cli doctor
Generated linear at ...
```

All 7 gates green. Then:

```bash
$ ./linear-pp-cli sync --resources teams
unknown flag: --resources
```

The emitted `sync.go` has no `--resources` flag, no GraphQL POST, no Connection-pattern pagination — it's REST scaffolding. The generated `internal/client/client.go` has `Get`, `Post`, `Put`, `Patch`, `Delete` methods — no `Query`, `Mutate`, or `PaginatedQuery`.

[...]

PR #2556, opened 2026-05-31
fix(cli): resolve GraphQL custom root operation types

tmchow:
## Problem

The SDL parser hard-coded the query/mutation root type names (`parseSDLContent`: `types["Query"]` / `types["Mutation"]`) and never parsed the optional top-level `schema { ... }` definition. GraphQL lets a schema alias its roots:

```graphql
schema { query: RootQuery mutation: RootMutation }
type RootQuery { ... }
```

A valid schema using aliased root names discovered **zero operations** and parsed as an empty API (then failed `APISpec.Validate` with an opaque message).

## Fix

- Add `rootOperationTypes(cleaned)` to parse the `schema { ... }` block and resolve the query/mutation root type names, defaulting to the conventional `Query`/`Mutation` when the block (or an operation) is absent.
- Use the resolved names for the resource walk instead of the hard-coded literals.
- When neither resolved root type is defined, return a clear error naming what it looked for, instead of emitting an empty spec.
- The schema-block regex forbids a colon between the keyword and the brace, so a field literally named `schema` (e.g. `schema: String`) can't be mistaken for the definition.

## Verification

- `go test ./...` — green (incl. new `TestParseSDLCustomRootOperations` + `TestParseSDLMissingRootOperations`)
- Existing `TestParseSDLContent` (conventional `type Query`/`type Mutation`) still passes — defaults unchanged
- `scripts/golden.sh verify` — 25 cases, no diff
- `scripts/verify-generator-output.sh` — 5 cases pass, incl. `generate-graphql-shared-endpoint`
- `golangci-lint run ./internal/graphql/` — 0 issues
- Found via automated review; clawpatch revalidate → `fixed`

Related (closed, distinct): #910, #669 cover other GraphQL generation gaps.

### DIFF
--- a/internal/graphql/parser.go
+++ b/internal/graphql/parser.go
@@ -64,8 +64,18 @@ func ParseSDLBytes(source string, data []byte) (*spec.APISpec, error) {
 // IsGraphQLSDL checks if the data looks like a GraphQL schema.
 func IsGraphQLSDL(data []byte) bool {
 	s := string(data)
-	return strings.Contains(s, "type Query") || strings.Contains(s, "type Mutation") ||
-		(strings.Contains(s, "type ") && strings.Contains(s, "scalar "))
+	if strings.Contains(s, "type Query") || strings.Contains(s, "type Mutation") {
+		return true
+	}
+	// A `schema { query: ... }` block that maps a root operation is an
+	// unambiguous GraphQL schema definition even when the schema aliases its
+	// roots and defines no scalars. Requiring the operation mapping (not just
+	// the keyword) keeps a literal "schema {" inside an OpenAPI description
+	// from being misclassified.
+	if m := schemaBlockRE.FindStringSubmatch(s); m != nil && schemaOpRE.MatchString(m[1]) {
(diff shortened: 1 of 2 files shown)


tmchow:
Addressed both Greptile comments in 43c469a6:

- **P1 (custom root types leaked into `apiSpec.Types`)**: the type-export loop now skips the resolved `queryRootName`/`mutationRootName` alongside the conventional names, so `RootQuery`/`RootMutation` no longer appear as entity types. `TestParseSDLCustomRootOperations` now asserts their absence from `parsed.Types`.
- **P2 (`IsGraphQLSDL` missed scalar-less custom-root schemas)**: detection now also matches a `schema { ... }` block that maps a root operation. Requiring the operation mapping (not just the keyword) avoids false-positiving a literal `schema {` inside an OpenAPI description. Covered by `TestIsGraphQLSDLDetectsCustomRootSchema`.

Verified: `go test ./...`, golden (25, no diff), verify-generator-output, golangci-lint (0).

Issue #1736, opened 2026-05-21
generator: workflow archive default sync set should exclude resources whose list endpoint requires non-paginator params


NOVAITEC:
## Summary

`workflow archive` includes every spec resource in its default sync set, regardless of whether that resource's primary list endpoint can be called without arguments. Resources whose list endpoint requires path params (`/users/{user_id}/items`) or non-paginator query params (`/items?ids=`) get called bare and return 404, producing visible `sync_error` events on every archive run.

## Evidence

From polymarket Phase 5 dogfood (`workflow archive` happy_path output):

```
{"event":"sync_start","resource":"book"}
{"event":"sync_error","resource":"book","status":404,"method":"GET","path":"/books","body":"404 page not found\n","error":"GET /books returned HTTP 404: 404 page not found\n"}
{"event":"sync_start","resource":"holders"}
{"event":"sync_error","resource":"holders","status":404,"method":"GET","path":"/holders", ...}
{"event":"sync_start","resource":"prices"}
{"event":"sync_error","resource":"prices","status":404,"method":"GET","path":"/prices", ...}
{"event":"sync_start","resource":"prices-midpoints"}
{"event":"sync_error","resource":"prices-midpoints","status":404,"method":"GET","path":"/midpoints", ...}
{"event":"sync_start","resource":"trades"}
{"event":"sync_error","resource":"trades","status":404,"method":"GET","path":"/trades", ...}
```

5 of 12 resources in polymarket's default sync set are non-syncable: `book` and `prices`/`prices-midpoints` are batch endpoints (`/books?params=...` and `/prices?params=...`) that 404 without `params=`; `holders` and `trades` need filter params (`market=`, `user=`).

## Cross-API recurrence

- **Polymarket** (this retro) — batch endpoints `/books`, `/prices`, `/midpoints` and per-user `/holders`, `/trades`
- **GitHub-shaped APIs** — `/users/{user}/repos`, `/issues/{number}/comments` shape
- **Stripe-shaped APIs** — `/customers/{id}/invoices`, `/charges/{id}/refunds` shape

## Proposed fix

In the generator's sync emission logic (the template that produces `internal/cli/sync.go` and the archive resource list in `internal/cli/channel_workflow.go`):

**Heuristic for default sync inclusion:**
- INCLUDE if the resource's primary list endpoint has no required path params AND no required non-paginator query params.
- EXCLUDE if any `required: true` path param exists.
[...]

PR #2588, opened 2026-06-02
fix(cli): generate working dependent-resource sync for nested sub-collections

sidorovanthon:
## Intent

Issue: Refs #1736 (the canonical-list-selection half; this does not fully resolve it)

A spec with project-scoped nested resources (parent → child fan-out) generated a CLI whose `sync` silently wrote **zero** dependent rows. Two independent root causes, both in the generator/profiler:

1. **Non-deterministic list-endpoint selection.** When several parameterized list paths share one resource key — e.g. `/projects/{project_id}/cycles/` and the deeper `/projects/{project_id}/cycles/{cycle_id}/cycle-issues/` — Go map iteration order could let the deeper sub-collection win the slot. That sub-collection needs an id that isn't available during per-parent fan-out, so the resource synced nothing.
2. **Missing typed parent-scope column on fan-out.** `buildSubResourceTable` declares the scope column `<parent_table>_id` as `NOT NULL` and the typed upsert reads by that name, but the fan-out only injected the generic `parent_id`. The API item never carries `<parent_table>_id` (e.g. Plane returns `"project"`, not `"projects_id"`), so every dependent upsert hit `NOT NULL constraint failed` and stored nothing — silently, in non-strict mode.

## Approach

- **profiler** (`profiler.go`): add `moreCanonicalListPath` and apply it at the parameterized-list collision point. The most canonical collection is the shallowest (fewest path params); ties break by shorter then lexical path, so selection is deterministic regardless of map order.
- **sync template** (`templates/sync.go.tmpl`): during fan-out, inject `obj[dep.ParentTable+"_id"]` alongside the existing `obj["parent_id"]`, so the typed `NOT NULL` scope column is populated.

Both are general generator behavior, not specific to any one API.

## Scope

Primary area: `cli` (generator + profiler)

Why this belongs in this repo: these are bugs in the Printing Press generator/profiler that affect every spec with project-scoped nested (dependent) resources — the symptom surfaced on a printed CLI, but the cause and fix are general machine behavior, not printed-CLI-local.

## Risk

The template change adds one injection line to generated `sync.go` for dependent resources; every fan-out recipe's generated output gains that line on next regen. No documented command, flag, or manifest surface is removed or renamed. The profiler change only makes an already-existing selection deterministic (shallowest path), which is the behavior fan-out already assumed.

## Output Contract

Template change: generated `sync.go` now writes `obj[dep.ParentTable+"_id"] = parentIDJSON` in addition to `obj["parent_id"]` for dependent resources, and carries an expanded explanatory comment. The `generate-golden-api` fixture's generated `sync.go` is updated to match (the only golden whose case exercises a dependent/fan-out resource). No other golden behavior changed.

## Verification

- `go build ./...` — passes.
- `go test ./internal/profiler/ ./internal/generator/` — the two new tests (`TestProfileNestedSubCollectionDoesNotHijackParentList`, `TestGenerateDependentResourceInjectsParentScopeColumn`) pass RED→GREEN. Pre-existing failures on the author's machine are Windows-only environment issues (generated-binary exec / `%PATH%`) and are byte-identical on the unmodified tree (confirmed by stashing the change); zero new failures.
- `scripts/golden.sh update` then inspected the diff: the only behavioral change is `generate-golden-api/.../internal/cli/sync.go` (the injection line). Other regenerated diffs were Windows-only normalization artifacts (backslash paths bypassing `normalize_text`, `.exe` binary suffix, CRLF-based `spec_checksum`) and were deliberately **not** included. A clean full `scripts/golden.sh verify` cannot run on Windows for those normalization reasons; Linux CI will verify the suite.
- End-to-end sanity (downstream, self-hosted Plane): fresh generate → sync populated dependent tables with real UUID scope columns where it previously stored zero.

## AI / Automation Disclosure

### DIFF
--- a/internal/generator/generator_test.go
+++ b/internal/generator/generator_test.go
@@ -350,6 +350,62 @@ func TestGenerateFreshnessHelperEmitted(t *testing.T) {
 	runGoCommand(t, outputDir, "test", "./internal/cli/...", "./internal/cliutil/...")
 }
 
+// TestGenerateDependentResourceInjectsParentScopeColumn reproduces the Plane
+// "fan-out stores zero rows" bug. A project-scoped child (cycles under projects)
+// yields a sub-resource table whose parent scope column (<parent>_id, here
+// projects_id) is NOT NULL, and the typed upsert reads that column from a
+// same-named item field. Plane API items carry "project", never "projects_id",
+// so unless the dependent-sync fan-out injects the scope column, every upsert
+// fails the NOT NULL constraint and the resource silently syncs zero rows.
+func TestGenerateDependentResourceInjectsParentScopeColumn(t *testing.T) {
+	t.Parallel()
+
+	apiSpec := &spec.APISpec{
+		Name: "plane-fanout",
+		Resources: map[string]spec.Resource{
+			"projects": {
+				Endpoints: map[string]spec.Endpoint{
+					"list": {
+						Method:     "GET",
+						Path:       "/projects/",
+						Response:   spec.ResponseDef{Type: "array"},
+						Pagination: &spec.Pagination{CursorParam: "cursor", LimitParam: "per_page"},
+					},
+				},
+				SubResources: map[string]spec.Resource{
+					"cycles": {
+						Endpoints: map[string]spec.Endpoint{
+							"list": {
+								Method:     "GET",
+								Path:       "/projects/{project_id}/cycles/",
+								Response:   spec.ResponseDef{Type: "array"},
+								Pagination: &spec.Pagination{CursorParam: "cursor", LimitParam: "per_page"},
+							},
+						},
+					},
+				},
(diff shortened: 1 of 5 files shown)


sidorovanthon:
Closing as redundant. After opening this I synced with current `main` and both fixes already exist upstream in more evolved forms:

- Deterministic canonical list-endpoint selection: `applySyncCandidates` stable-sorts candidates by path length (lexical tiebreak) and takes the shallowest, which is what this PR's `moreCanonicalListPath` did.
- Typed parent-scope column on fan-out: `sync.go.tmpl` already computes `parentFKKey := dep.ParentTable + "_id"` and injects it conditionally (plus path-param fields), superseding this change.

My fork was branched off a stale main; apologies for the noise. Closing — no action needed.

PR #2633, opened 2026-06-04
feat(cli): four-kind path resolution with credential relocation for generated CLIs

### LINKED ISSUE #2632: feat(cli): four-kind path resolution (config/data/state/cache) with credential relocation for generated CLIs
## Problem

Generated CLIs compute filesystem paths at 19+ hardcoded sites across two packages (`cli` + MCP `main`), with no XDG support and no env overrides for data/cache. Auth secrets (API keys, OAuth/bearer tokens, cookie/composed header values) persist inside `~/.config/<name>-pp-cli/config.toml` — a directory commonly synced as dotfiles. There is no single-root relocation story for containers or agent fleets, and the CLI and its MCP sibling hand-roll divergent paths (the duplication is even acknowledged in a code comment in `mcp_tools.go.tmpl`).

## Proposal

Port gog's published "Paths and State" scheme (https://gogcli.sh/paths.html) into the generator templates so every printed CLI gets:

- A typed four-kind resolver emitted into `internal/cliutil/`: per-kind ladder `<PREFIX>_<KIND>_DIR` → `--home` → `<PREFIX>_HOME` (flat root) → XDG → platform default (= today's paths, so existing installs resolve identically with zero action)
- Credentials relocated to a data-kind `credentials.toml` (0600/0700, atomic writes) with read-first legacy compat, lazy migration on the next auth write, and an agentcookie exemption (daemon-managed CLIs keep config-pinned storage)
- MCP binary resolving identical paths through the same resolver (kills the institutionalized duplication)
- `doctor` reporting all four resolved paths + winning rung + split-credential WARN with a new `--fail-on warn` tier; agent-context schema v4 with a runtime paths block
- Verify/dogfood subprocess scoping extended so operator relocation env vars cannot leak into harness runs

## Scope notes

- Non-breaking per AGENTS.md's own definition: read-first legacy compat keeps the config-file shape compatible (`feat(cli)`, not `feat(cli)!`)
- Cross-repo: the new README/SKILL `## ` sections trigger the sweep-canonical lockstep rule — tracking issue will be filed at printing-press-library before merge, plus the `COMMON_FLAGS` allowlist addition for `--home`
- Implementation is complete on a branch (full test suite, 25 golden cases, verify-generator-output, and an end-to-end paths smoke all green); PR follows shortly.

bobeglz:
## Intent

Generated CLIs hardcode filesystem paths at 19+ sites across two packages, keep auth secrets inside `~/.config/<name>-pp-cli/config.toml` (a directory commonly synced as dotfiles), honor no XDG variables, and offer no single-root relocation for containers or agent fleets. The CLI and its MCP sibling binary each hand-roll divergent path logic.

Issue: Closes #2632

## Approach

Ports gog's published "Paths and State" scheme (https://gogcli.sh/paths.html) into the generator as a machine change:

- **One emitted resolver** (`internal/cliutil/paths.go` in every printed CLI): typed four-kind constants (config/data/state/cache) with a per-kind ladder — `<PREFIX>_<KIND>_DIR` → `--home` → `<PREFIX>_HOME` (flat root) → XDG variable → platform default. The platform-default rung returns exactly today's paths, so override-free installs resolve bit-for-bit identically.
- **Credentials leave config.toml** for a data-kind `credentials.toml` (0600 file / 0700 dirs, atomic CreateTemp+rename writes). Read precedence: env var → credentials file → legacy config fields. Writes commit the credentials file first, then atomically rewrite config.toml from a secrets-free `persistedConfig` view (structural exclusion — re-marshaling secrets is impossible by type). Migration is lazy (next auth write); reads never require user action. agentcookie-managed CLIs are exempted — the daemon owns config-pinned storage.
- **Resilient degradation:** corrupt/value-less credentials files and malformed *legacy* configs warn and degrade instead of bricking every command; parse errors in the *active* config source stay fatal.
- **MCP parity:** the `<slug>-mcp` binary resolves through the same emitted resolver (the duplicated `dbPath()` and its apologetic comment are gone); an emitted test asserts CLI↔MCP path agreement; the MCP `context` tool exposes runtime-resolved paths so hosts can detect env mismatch.
- **Surfacing:** `doctor` reports each kind's resolved dir + winning rung, warns on split credential locations and silently-skipped relative overrides, with a new `--fail-on warn` tier (orthogonality with `stale` documented in help). agent-context bumps schema v3→v4 with a runtime paths block (single golden consumer updated in-change).
- **Harness hardening:** verify/dogfood subprocess scoping now scrubs relocation env vars by suffix grammar (any `*_{HOME,CONFIG_DIR,DATA_DIR,STATE_DIR,CACHE_DIR}`), so a poisoned operator environment cannot leak into mock-mode runs.

## Scope

Primary area: `internal/generator/templates/**` (resolver, credentials store, config, auth flavors, doctor, agent-context, MCP, docs templates), `internal/generator` render wiring + tests, `internal/naming` (kind-suffix source), `internal/pipeline/subprocess_env.go`, `scripts/e2e-paths-smoke.sh`.

Why this belongs in this repo: pure machine change — every printed CLI inherits the scheme on its next regen; nothing here is API-specific. Templates use `{{.Name}}`/`{{envName .Name}}` placeholders throughout.

## Catalog Justification

N/A

## Risk

- **Legacy installs** are the load-bearing risk: mitigated by read-first fallback at every relocated file, platform defaults equal to the old hardcoded paths, lazy migration, and an e2e legacy-seed simulation (seed old layout → reads work untouched → first auth write migrates B→N → doctor clean).
- **Config-file shape**: non-breaking per this repo's own definition — read-compat preserves the documented surface, so this is `feat(cli)`, not `feat(cli)!`.
- **Generated-output churn**: all 25 golden cases updated intentionally; diff classes are the resolver/credentials/doctor/MCP shapes described above.
- **MCP surface**: additive only (context tool gains a `paths` block; agent-context v4 adds `paths`).
- **Verifier**: subprocess env scoping extended (XDG_STATE_HOME + per-CLI relocation vars); pipeline tests cover the poisoned-env path.
- **Go floor**: generated `go.mod` floor bumped 1.26.3 → 1.26.4 (govulncheck gate required it during e2e).

## Output Contract

- Emitted-code assertions + `requireGeneratedCompiles` across new generator tests (paths emission sweep, config paths, doctor paths, auth browser paths, credential gating)
- Emitted runtime test templates: `cliutil_paths_test.go.tmpl` (ladder, `t.Setenv`-controlled), `cliutil_credentials_test.go.tmpl` (state machine: both-present, B→N, logout-from-B, concurrent two-writer race, corrupt/value-less degradation, AuthHeader-level B reads), `mcp_tools_test.go.tmpl` (CLI↔MCP agreement)
- Golden fixtures: 25 cases updated, including auth-flavor variants (rich-auth, oauth2-cc, oauth2-device-code, cookie-auth) and `generate-mcp-api`
- `scripts/e2e-paths-smoke.sh`: fresh print → 8 publish gates → legacy-seed migration → relocation matrix (defaults / `<PREFIX>_HOME` / per-kind / `--home`) → MCP parity → poisoned-env scrub → second claimed output dir
- Cookie + composed flavors now execute their emitted credential round-trip tests via generator-side `runGoCommand` tests

## Verification

- [x] Generator/template change: verified generated output, including emitted-code assertions or compiled generated CLI output (`requireGeneratedCompiles`, emitted `go test` runs, 25 goldens)
- [x] Generator/template change: covered the affected fallback or variant shape, not only happy-path fixtures (auth-none gating, cookie/composed/device-code flavors, corrupt/value-less/malformed degradation paths, legacy-seed migration)
- [x] Generator/template change: checked emitted definitions and call sites for matching gates (single `shouldEmitAuth` predicate threaded through credentials emission; sweep test forbids path roots outside the resolver)

Commands run (all green): `go build ./...` · `go test ./...` · `bash scripts/golden.sh verify` (25 cases, after intentional updates) · `bash scripts/verify-generator-output.sh` (5 compile cases) · `bash scripts/e2e-paths-smoke.sh` · `go vet ./...`. Not run: `golangci-lint` locally (not installed on this machine — relying on CI/lefthook).

## AI / Automation Disclosure

Implementation was delegated to OpenAI Codex (GPT-5, `xhigh`/`high` reasoning) in five verified batches, orchestrated and code-reviewed by Claude Opus 4.8 (10-persona review + 15 independent finding validators + a four-angle simplification pass). A human (@bobeglz) directed scope, approved each gate decision, and authorized submission, but did not line-review the code.

## Known Residuals

- **Cross-repo (pre-merge blocker):** the new `## Paths & environment variables` (README) and `## Paths and state` (SKILL) template sections trigger the sweep-canonical lockstep rule, and generated SKILLs now mention `--home` (needs the library `verify_skill.py` `COMMON_FLAGS` allowlist). Tracking issue filed at printing-press-library (linked in comments).
- Advisory (non-blocking): live-check test timeouts raised 5s→15s under full-suite load (root cause: subprocess startup variance under parallel load — worth a baseline marker); `--fail-on warn` exits non-zero for legacy installs until their first auth write consolidates credentials (release-notes material); verify `xnet_guard`'s pinned version against the Go 1.26.4 module graph on the next sweep.

## Post-Deploy Monitoring & Validation

- **Validation window / owner:** next 2-3 fresh prints + first reprint of an existing published CLI, owned by whoever runs them.
- **Healthy signals:** `doctor --json` shows four `paths` kinds with `rung: "platform default"` on override-free installs; `credentials_location: "credentials file"` after first auth write; publish gate #8 (doctor) green on fresh prints.
- **Failure signals:** doctor `paths_warning` on machines with no relocation vars set (would indicate a resolver regression); legacy CLI reporting unauthenticated after upgrade (read-compat regression — check `credentials_location` and the legacy probe); `--fail-on warn` tripping in library CI unexpectedly.
- **Log/query anchors:** stderr `WARN`-prefixed once-per-path warnings from generated CLIs (`relative override skipped`, credentials degradation warnings); `doctor --json | jq .paths,.credentials_location`.
- **Rollback/mitigation:** revert is clean (template-only change; no persisted-state migration to unwind — legacy files are never deleted, only shadowed; a reverted binary reads the legacy locations it always read).

---

🤖 Compound Engineered: planned, implemented, and reviewed with [compound-engineering](https://github.com/EveryInc/compound-engineering-plugin) — Claude Opus 4.8 (orchestration, review) + OpenAI Codex GPT-5 (implementation batches).


bobeglz:
Cross-repo tracking issue filed per the sweep-canonical lockstep contract: mvanhorn/printing-press-library#1048 (covers the two new template sections + the `COMMON_FLAGS` `home` allowlist that should land library-side before/with the first regen that picks this up).

tmchow:
@bobeglz Fix linting?

bobeglz:
Lint findings fixed in `fae6b047` — `stringscutprefix` (modernize) in the pipeline test helper, `QF1008` promoted-field selector in the doctor test, and the `credentialTemplateData` type orphaned by an earlier dead-case cleanup. @tmchow

internal/generator/templates/config.go.tmpl
```diff
@@ -710,54 +892,101 @@ func (c *Config) ClearTokens() error {
...
+	c.{{.GoField}} = ""
 {{- end}}
-{{- range .Auth.AdditionalHeaders}}
-{{- if not (envVarIsBuiltinField .EnvVar.Name)}}
-	c.{{envVarField .EnvVar.Name}} = ""
-{{- end}}
+{{- if .HasRequiredRoles}}
+	c.AuthRole = ""
 {{- end}}
+	if c.AgentcookieManagedByExternalStore() {
```
tmchow, line 1003: This leaves secrets behind on `auth logout` when `.agentcookie-managed` exists. `ClearTokens` zeroes the credential fields in memory, but then returns before persisting the cleared config. Since `save()` intentionally persists the full config for agentcookie-managed stores, logout needs to save the zeroed fields in this branch as well; otherwise config-stored credentials remain on disk after logout. Please add a generated-output regression covering `.agentcookie-managed` + existing config credentials + `ClearTokens()` so the secret is actually removed from `config.toml`.

internal/generator/templates/root.go.tmpl
```diff
@@ -221,6 +223,7 @@ Run '{{.Name}}-pp-cli doctor' to verify auth and connectivity.{{backtick}},
 	rootCmd.PersistentFlags().BoolVar(&flags.plain, "plain", false, "Output as plain tab-separated text")
 	rootCmd.PersistentFlags().BoolVar(&flags.quiet, "quiet", false, "Bare output, one value per line")
 	rootCmd.PersistentFlags().StringVar(&flags.configPath, "config", "", "Config file path")
+	rootCmd.PersistentFlags().StringVar(&flags.homePath, "home", "", "Root directory for config, data, state, and cache files")
```
tmchow, line 224: Adding `--home` here also needs MCP command-mirror handling. `internal/generator/templates/cobratree/shellout.go.tmpl` blocks root flags like `--config`, `--profile`, and `--token`, but it does not block `home`, so an MCP caller can pass structured `{"home": ...}` and relocate filesystem state for shell-out tools while typed MCP tools continue using the server env paths. Please add `home` to the blocked root flags and cover it in the shell-out argument tests/goldens.

internal/generator/templates/config.go.tmpl
```diff
@@ -167,31 +196,20 @@ func Load(configPath string) (*Config, error) {
...
-	}
-{{- end}}
-{{- end}}
-{{- else if .Auth.EnvVars}}
-{{- range .Auth.EnvVars}}
-{{- if not (envVarIsBuiltinField .)}}
-	if cfg.AuthSource == "" && cfg.{{envVarField .}} != "" {
-		cfg.AuthSource = "config"
+{{- if .HasAuthCommand}}
+	if cfg.CredentialSource == "" && cfg.AuthSource == "config" {
```
tmchow, line 230: This fallback mislabels active relocated or explicit config-file credentials as `legacy config path`. In the non-legacy branch above, credentials loaded from the current config-kind path still reach this fallback with `AuthSource == "config"`, so doctor/auth-status report the wrong credential location. Please carry the parsed config owner through load, e.g. `config-kind path` vs `legacy config path`, and assert the doctor/JSON label for relocated and explicit config credentials.

internal/pipeline/subprocess_env.go
```diff
@@ -51,15 +54,18 @@ func newScopedConfigHome() (string, func(), error) {
...
-func applyScopedConfigHome(env []string, homeDir string) []string {
+func applyScopedConfigHome(env []string, homeDir string, cliRelocationEnvVars ...string) []string {
 	if homeDir == "" {
 		return env
 	}
 	overrides := scopedConfigHomeOverrides(homeDir)
 	out := make([]string, 0, len(env)+len(overrides))
 	for _, kv := range env {
-		if isScopedConfigHomeEntry(kv) {
+		if isScopedConfigHomeEntry(kv) || isRelocationOverrideEnvEntry(kv) || isScopedCLIEnvEntry(kv, cliRelocationEnvVars) {
```
tmchow: This scrub is broader than the stated CLI relocation scope. `isRelocationOverrideEnvEntry` drops any uppercase env var ending in `_HOME`, `_CONFIG_DIR`, `_DATA_DIR`, `_STATE_DIR`, or `_CACHE_DIR`, so verify/dogfood subprocesses can lose unrelated setup such as `JAVA_HOME`, `CARGO_HOME`, `ANDROID_HOME`, `CHROME_USER_DATA_DIR`, or API-specific variables ending in `_DATA_DIR`. The scoped helper already accepts explicit CLI relocation names; please restrict stripping to the active CLI prefixes plus the standard HOME/XDG vars, and add tests proving unrelated vars survive while the target CLI's relocation vars are removed.

internal/generator/templates/profile.go.tmpl
```diff
@@ -100,7 +113,7 @@ func ApplyProfileToFlags(cmd *cobra.Command, profile *Profile) error {
 	// Reserved flags that never come from a profile - they control profile
 	// resolution itself or are dangerous to overlay.
 	reserved := map[string]bool{
-		"profile": true, "config": true, "help": true,
+		"profile": true, "config": true, "home": true, "help": true,
```
tmchow, line 118: This makes `home` a reserved flag when applying a profile, but `profile save` still captures `--home` because its `skip` map below omits it. That produces profiles containing a value that will never apply, which is confusing and makes saved profiles lie about their effect. Please skip `home` during save as well and update the help text that currently says only `--profile` and `--config` are excluded.

internal/generator/templates/readme.md.tmpl
```diff
@@ -397,6 +397,55 @@ These capabilities aren't available in any other tool for this API.
...
 Run `{{.Name}}-pp-cli --help` for the full command reference and flag list.
 
+## Paths & environment variables
+
+This CLI separates local files into four path kinds:
+
+| Kind | Contents |
+|------|----------|
+| `config` | User-editable settings such as `config.toml` and saved profiles |
+| `data` | Durable local data: `credentials.toml`, `data.db`, cookies, browser-session proof files, and other auth sidecars |
```
tmchow: This docs block is emitted for no-auth CLIs too, but it describes `credentials.toml`, cookies, browser-session proof files, auth sidecars, and first-auth-write migration. Those files/commands are not present for no-auth outputs, so the generated README is misleading. Please conditionalize the auth-specific contents and migration paragraph on the auth surface, and update the no-auth golden/generated-output coverage.

tmchow (changes requested):
Thanks for pushing this forward. I think the path model is directionally right, but this PR is not ready yet. The blockers are around logout persistence, MCP parity for the new --home flag, credential-source reporting, env scrubbing, profile capture, and no-auth generated docs. Please address the inline items before merge.

internal/generator/templates/config.go.tmpl
bobeglz: Fixed — the agentcookie-managed branch of `ClearTokens` now persists the zeroed fields via `c.save()` instead of returning early. Since `save()` intentionally writes the full config for agentcookie-managed stores, the cleared credential fields now reach disk and the secret leaves `config.toml`.

Regression: `TestGeneratedAgentcookieLogoutRemovesConfigStoredSecrets` (internal/generator/config_paths_test.go) generates a CLI, seeds `.agentcookie-managed` + config-stored credentials, calls `ClearTokens()`, and asserts the secret is gone from `config.toml` on disk and that a reload carries no credentials. Verified the test fails against the previous early-return. Fixed in bad755ba — all six threads addressed there.

internal/generator/templates/root.go.tmpl
bobeglz: Done — `home` is now in `blockedRootFlags` in the shell-out mirror, so a structured `{"home": ...}` argument is dropped like `--config`/`--profile`/`--token`. Covered in `TestCliArgsFromMCP_BlocksRootFlags` (shellout_test.go.tmpl) and pinned in the golden-api goldens for both shellout.go and shellout_test.go. Fixed in bad755ba.

internal/generator/templates/config.go.tmpl
bobeglz: Fixed — `parseConfigData` now stamps the parsed config owner on the Config (unexported `configOwner`: "config-kind path" vs "legacy config path"), and the credential-source fallback labels from it instead of hardcoding the legacy label. Relocated (per-kind env / `--home`) and explicit `--config` credentials now report `config-kind path`; only a read that actually fell back to the pre-paths layout reports `legacy config path`.

Asserted in `TestGeneratedDoctorCredentialLocationLabels` (internal/generator/doctor_paths_test.go) against doctor's JSON `credentials_location` for all three cases (relocated, explicit `--config`, legacy fallback); also verified the test fails pre-fix. Fixed in bad755ba.

internal/pipeline/subprocess_env.go
bobeglz: Agreed — dropped `isRelocationOverrideEnvEntry` (and its `isUpperEnvPrefix` helper) entirely. The scrub is now exactly the standard HOME/XDG set plus the explicit per-CLI relocation names; all production entry points already pass `findCLINames(...)`, so the target CLI's relocation vars are still removed.

Tests: `TestSubprocessEnvKeepsUnrelatedRelocationShapedVars` proves `JAVA_HOME`, `CARGO_HOME`, `ANDROID_HOME`, `CHROME_USER_DATA_DIR`, `OTHERAPI_DATA_DIR`, and `FOO_CONFIG_DIR` survive while `PRINTING_PRESS_RICH_DATA_DIR` is removed; `TestSubprocessEnvKeepsRelocationVarsWhenCLINameUnknown` documents the no-names trade-off (nothing CLI-specific is stripped when no CLI names are installed). Fixed in bad755ba.

internal/generator/templates/profile.go.tmpl
bobeglz: Done — `home` added to profile save's skip map, with cross-reference comments pinning skip as a superset of apply's `reserved` set, and the help text now reads "other than --profile, --config, and --home, which are never captured". Pinned by the golden-api profile.go golden. Fixed in bad755ba.

internal/generator/templates/readme.md.tmpl
bobeglz: Done — the data-row contents, the "doctor cannot find credentials" phrase, and the first-auth-write migration paragraph are now gated on the emitted auth surface (`HasAuthCommand`, the same `shouldEmitAuth()` predicate that gates auth-command emission — deliberately not `HasAuth`, which diverges on the empty-auth-type edge). No-auth READMEs now read "Durable local data such as `data.db`" with a path-warnings-only doctor hint.

Coverage: the no-auth golden `generate-public-param-names` README updated accordingly, plus `TestGeneratedREADMEPathsSectionFollowsAuthSurface` (internal/generator/readme_test.go) asserts both shapes. Fixed in bad755ba.

bobeglz:
@tmchow thanks for the thorough review — all six blockers are addressed in bad755ba, and each thread has an inline reply with the fix + its regression. I've resolved the threads per the conversation-check flow; please re-review when you get a chance.

Highlights:

- **Logout persistence**: agentcookie-managed `ClearTokens` now saves the zeroed fields; generated regression verifies the secret leaves `config.toml` on disk (and fails against the old early-return).
- **MCP `--home` parity**: `home` joined `blockedRootFlags` in the shell-out mirror, with test-template + golden coverage.
- **Credential-source labels**: the parsed config owner now flows through `Load`, so doctor reports `config-kind path` for relocated/explicit `--config` credentials and `legacy config path` only for actual legacy-fallback reads (asserted against doctor's JSON for all three cases).
- **Env scrub**: the suffix-based sweep is gone — stripping is the standard HOME/XDG set plus the active CLI prefixes; tests prove `JAVA_HOME`/`CARGO_HOME`/`ANDROID_HOME`/`CHROME_USER_DATA_DIR`/`<OTHER>_DATA_DIR` survive.
- **Profile save**: `--home` skipped on save (superset comments keep save/apply in sync) and the help text lists the exclusions.
- **No-auth docs**: the paths section's credential prose and migration paragraph are gated on the emitted auth surface; the no-auth golden README updated accordingly.

Verification: full `go test ./...`, `golden.sh` (diff scoped to exactly these fixes), `verify-generator-output.sh` across auth/no-auth/cookie variants, and `golangci-lint` clean.

Cross-repo lockstep: the merge-ordering-sensitive piece of printing-press-library#1048 (allowing `home` in `COMMON_FLAGS`) is up as mvanhorn/printing-press-library#1054 so the fleet is ready before the first regen that picks this up.

bobeglz:
Hi @tmchow — all six review blockers have been addressed (in `bad755ba`) and the branch is now current with `main` (merge conflicts resolved). Ready for re-review. 🙏

tmchow:
Thanks! Will take some time to review it.

internal/generator/templates/config.go.tmpl
```diff
@@ -103,30 +100,60 @@ func Load(configPath string) (*Config, error) {
...
+			} else {
+				*cfg = parsed
+			}
 		}
 	}
-{{- else}}
-	if data, err := os.ReadFile(path); err == nil {
-		if err := json.Unmarshal(data, cfg); err != nil {
-			return nil, fmt.Errorf("parsing config %s: %w", path, err)
+	cfg.Path = path
```
tmchow, line 151: This is the blocker I hit in a relocated legacy install. When the primary config path is under XDG/`<PREFIX>_CONFIG_DIR` and `ReadFileWithLegacyFallback` reads the old `~/.config/<app>/config.toml`, this line restores `cfg.Path` to the new resolved path. Later `SaveCredential`/`SaveTokens` write the new `credentials.toml` and clean only the new config file, leaving the old legacy config file with the previous token still on disk. I reproduced it with a legacy `press_golden_api_key = "legacy-secret"`, `XDG_CONFIG_HOME` and `XDG_DATA_HOME` set, then `auth set-token new-secret`: the new files were clean, but old `~/.config/printing-press-golden-pp-cli/config.toml` still contained `legacy-secret`. Please track the fallback source and scrub/rewrite legacy credential fields on auth write/logout when fallback was used, with a generated-output regression for the relocated case.

internal/generator/templates/doctor.go.tmpl
```diff
@@ -885,6 +910,146 @@ func extractBearerToken(body []byte, pattern string) (string, error) {
...
+		ok, err := config.FileHasCredentialFields(path)
+		if err == nil && ok {
+			locations = append(locations, path)
+		}
+	}
+	if len(locations) > 0 {
+		report["credentials_locations"] = locations
+	}
+	if credsPresent && len(locations) > 1 {
+		report["credentials_location_warning"] = "WARN credentials stored in more than one location; current reads use credentials file; run auth set-token or auth logout to consolidate"
```
tmchow: This remediation is currently inaccurate for the relocated fallback case above. Doctor tells users to run `auth set-token` or `auth logout` to consolidate, but `auth set-token` leaves the legacy config secret behind when the resolved config path differs from the old path. Once the legacy scrub is fixed this text may be fine; otherwise this needs to name the exact legacy config file and say manual cleanup/move is required.

scripts/e2e-paths-smoke.sh
```diff
@@ -0,0 +1,298 @@
...
+locations = data.get("credentials_locations") or []
+assert locations == ["credentials file"], locations
+PY
+}
+
+assert_no_config_secrets() {
+  local config_path="$1"
+  python3 - "$config_path" <<'PY'
+import sys
+import tomllib
```
tmchow: This makes the new smoke script require Python 3.11+. On this Mac, `/usr/bin/python3` is 3.9.6, so `bash scripts/e2e-paths-smoke.sh` failed here with `ModuleNotFoundError: No module named 'tomllib'` after generation/build/verify and the legacy sync section. Please either avoid `tomllib` for this assertion, vendor a fallback such as `tomli` only when available, or explicitly preflight and fail fast with a Python 3.11+ requirement.

tmchow (changes requested):
Thanks for the fixes on the first round. I re-reviewed the latest head and I think the path model is still directionally right, but there is one compatibility/security blocker before merge: relocated legacy installs can keep the old config secret on disk after the advertised first auth write. Please address that, and also tighten the local smoke-script portability noted below.

Requested changes:

1. When config is loaded from the legacy fallback and the resolved config path is different, auth writes/logout need to scrub or rewrite the legacy config file too, or doctor/docs need to stop claiming auth writes consolidate the old secret. My preference is to scrub the legacy credential fields while preserving non-secret settings, and add generated-output/e2e coverage for: legacy ~/.config config with token, XDG_CONFIG_HOME/XDG_DATA_HOME set, run auth set-token, assert new credentials.toml exists and the old ~/.config/.../config.toml no longer contains credential fields.

2. Update the doctor warning so its advice matches the actual remediation. Right now it says to run auth set-token/logout, but in the relocated fallback case auth set-token is the action that still leaves the old config secret behind.

3. Make scripts/e2e-paths-smoke.sh runnable with the repo's local tooling expectations, or explicitly gate it on Python 3.11+. On my macOS Python 3.9.6, it fails at import tomllib after getting through the legacy migration section.

Non-blocking rollout note: existing data.db/local store state does not appear to have a legacy fallback when XDG/per-kind data relocation is active; please call out that users may need to move the DB or resync when they deliberately relocate data roots.

tmchow:

internal/generator/templates/config.go.tmpl
bobeglz: Fixed in 93b40d0 (on the branch since the second-round review on 2026-06-07; it sits below today's `main` merge so it's easy to miss in the diff).

`Load` now records `legacySourcePath` whenever the read falls back to the legacy config, and `save()` runs `scrubLegacyCredentials()`, which rewrites the legacy file secret-free (the same secrets-free `persisted()` view used for the active config) on every auth write and on logout, preserving non-secret settings. It degrades gracefully (warns, never aborts the primary save) on a missing/unreadable/unparseable legacy file, and is skipped for agentcookie-managed installs.

Regression test added: `TestAuthWriteScrubsLegacyConfigWhenRelocated` in `cliutil_credentials_test.go.tmpl` reproduces your exact repro — legacy `press_golden_api_key`, `XDG_CONFIG_HOME` + `XDG_DATA_HOME` set, then a credential write — and asserts the legacy `config.toml` is scrubbed of both the old and the new secret (with `base_url` preserved), the active relocated config is secret-free, and the credentials file lands in the new data dir.

internal/generator/templates/doctor.go.tmpl
bobeglz: Also addressed in 93b40d0. `collectCredentialsLocationReport` now detects the relocated-legacy case and names the exact legacy path in the warning:

> WARN credentials stored in more than one location; legacy secrets remain at <path>; run auth set-token or auth logout to consolidate and remove legacy secrets

With the legacy scrub above, that consolidation advice is now accurate — the next `auth set-token`/`auth logout` removes the legacy secret rather than leaving it behind.

scripts/e2e-paths-smoke.sh
bobeglz: Fixed in 93b40d0. Replaced the `tomllib` load in `assert_no_config_secrets` with a regex secret-key scanner that imports only `re` and `sys`, so the smoke script runs on macOS system Python 3.9. The scanner skips commented lines and anchors on `^<key>\s*=`, so prefixed keys don't false-match. The script also gained a legacy-scrub assertion under relocation env vars.

bobeglz:
@tmchow — all three second-round items are addressed in 93b40d0. It's been on the branch since 2026-06-07, so it sits underneath today's `main` merge and is easy to miss in the diff:

- **Legacy credential leak on relocation** — `save()` now scrubs/rewrites the legacy `config.toml` secret-free on every auth write and on logout when the active path was relocated, with the regression test `TestAuthWriteScrubsLegacyConfigWhenRelocated` reproducing your exact repro.
- **Doctor warning** — now names the exact legacy path and the consolidation step.
- **Smoke script** — `tomllib` replaced with a `re`/`sys`-only scanner, so it runs on macOS system Python 3.9.

Detail in the inline replies on each thread. I don't have permission to re-request review on this repo, so flagging here instead — could you take another look when you get a chance? And thanks for pulling `main` in.

internal/generator/templates/config.go.tmpl
bobeglz: Re-verified on the latest head before resolving:

- `scrubLegacyCredentials()` runs in `save()` on every auth write and on logout (gated on `HasAuthCommand`), before the active-config credential fields are cleared.
- It no-ops unless a *relocated* legacy fallback was actually used (`legacySourcePath != "" && legacySourcePath != c.Path`), exempts agentcookie-managed installs, and degrades gracefully (warns, never aborts the primary save) on a missing / unreadable / unparseable legacy file.
- It rewrites the legacy file from the secrets-free `persisted()` view via `cliutil.AtomicWritePrivateFile(..., 0o600, 0o700)`, so non-secret settings (e.g. `base_url`) survive and credential fields are structurally excluded.
- `TestAuthWriteScrubsLegacyConfigWhenRelocated` reproduces the exact repro (legacy `..._api_key` + `XDG_CONFIG_HOME`/`XDG_DATA_HOME` set, then a credential write) and asserts the legacy `config.toml` is scrubbed of both the old and the new secret while `base_url` is preserved, the active relocated config is secret-free, and `credentials.toml` lands in the new data dir. It runs against generated output via `TestCredentialsRoundTripTestsRunForBrowserAndDeviceAuthFlavors` (cookie / composed / device-code flavors) — green locally just now.

Marking this resolved. @tmchow — re-requesting your review; all three round-2 items are in and verified.

bobeglz:
@greptileai review — this PR never got an automated Greptile pass (likely the large generated-output diff). Requesting a review on the current head; all review feedback is addressed and threads are resolved.

internal/generator/templates/cliutil_paths.go.tmpl
```diff
@@ -0,0 +1,307 @@
...
+	PathKindState
+	PathKindCache
+)
+
+var (
+	pathHomeOverride string
+	pathWarnedMu     sync.Mutex
+	pathWarned       = map[string]struct{}{}
+)
+
```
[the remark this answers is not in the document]
bobeglz: Fixed in 49388ee: `pathHomeOverride` is now guarded by `pathHomeOverrideMu sync.RWMutex`, matching the `scopedHomeDirMu` pattern in `subprocess_env.go`. `SetHomeOverride` performs the swap under the lock and returns a restore func that also locks; `HomeOverrideActive` and `ResolveKindDir` read through a locked `homeOverride()` getter. Full `go test ./...` and `verify-generator-output.sh` (cookie-auth + rich-auth) are green.

internal/generator/templates/auth_browser.go.tmpl
```diff
@@ -1027,19 +1030,44 @@ func newAuthRefreshQueriesCmd(flags *rootFlags) *cobra.Command {
...
+		home, err := os.UserHomeDir()
+		if err != nil {
+			return "", false
+		}
+		path = filepath.Join(home, strings.TrimPrefix(path, "~/"))
+	}
+	if !filepath.IsAbs(path) {
+		return "", false
+	}
+	return filepath.Clean(path), true
```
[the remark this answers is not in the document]
bobeglz: Fixed in 49388ee: extracted the shared logic to an exported `cliutil.CleanPathOverride` (trim + tilde-expand + absolute-path validation). `cleanOutputPath` is removed and the `--output` path now calls `cliutil.CleanPathOverride`, so the tilde/abs handling lives in one place with a single error contract. The behavioral difference you flagged is gone — both sites now go through `expandTilde` + `filepath.IsAbs`.

bobeglz:
@greptileai review — addressed all three P2 findings in 49388ee (pathHomeOverride mutex, shared cliutil.CleanPathOverride, dropped the spurious doctor probe + narrowed its import). Re-review of the current head, please.

bobeglz:
@tmchow — ready for re-review. Everything from your second-round review and the Greptile pass is addressed and verified; the PR is green except the items only you can action (re-approval + `ready-to-merge`).

**Your round-2 blockers** (verified, with regressions):
1. **Legacy-config credential scrub on relocation** — `Config.save()` now scrubs the legacy file via `scrubLegacyCredentials()` (secrets-free `persisted()` view, atomic 0600 write); `TestAuthWriteScrubsLegacyConfigWhenRelocated` reproduces your exact repro (legacy secret + `XDG_CONFIG_HOME`/`XDG_DATA_HOME`, then `set-token`).
2. **Doctor warning accuracy** — the relocated-legacy case now names the actual legacy path.
3. **`e2e-paths-smoke.sh` portability** — `tomllib` replaced with a regex scanner (runs on macOS Python 3.9).

**Greptile P2s** (fixed in `49388ee`, threads resolved):
1. `pathHomeOverride` is now guarded by `pathHomeOverrideMu sync.RWMutex` (mirrors `scopedHomeDirMu`); `SetHomeOverride` swaps under the lock and reads go through a locked `homeOverride()`.
2. Tilde expansion deduped into exported `cliutil.CleanPathOverride`; the duplicate `cleanOutputPath` is removed.
3. Dropped the spurious same-dir config probe in `legacyCredentialProbePaths` (it was redundant with `cfg.Path` or could match an unrelated CLI's config, producing a false "credentials in more than one location" warning) and narrowed `doctor.go`'s `filepath` import to its remaining user.

**Verification:** `go test ./...`, `scripts/verify-generator-output.sh` (cookie-auth + rich-auth), and `scripts/golden.sh verify` (31 cases; 3 `doctor.go` fixtures updated) — all green. All 11 review threads resolved.

**One infra note on the `Greptile Review` check:** Greptile re-reviewed the current head (`49388ee`) at **5/5, "safe to merge"**, but its status *check* is stranded on the prior head — Greptile size-skips the auto-review on this 128-file diff, and the `@greptileai` re-review updated the comment without posting a new check-run. The dashboard "Re-trigger Greptile" link is login-gated, so it needs a Greptile-authenticated maintainer to re-trigger for `Greptile Review` to go green on the current head.

tmchow:
@greptileai review

tmchow (approved):

tmchow (approved):
Merge conflict resolved and Greptile feedback addressed; re-approving for queue.

PR #2650, opened 2026-06-05
fix(cli): surface row errors in generated MCP sql tool

tmchow:
## What

The generated `handleSQL` MCP tool (`internal/generator/templates/mcp_tools.go.tmpl`) discarded the errors from `rows.Columns()` and `rows.Scan()`, and never checked `rows.Err()` after the loop. Because `rows.Next()` returns `false` on a mid-iteration DB error *without* failing the loop, the tool returned a **silently truncated result set** to the agent as a successful tool call — no failure signal. This affects every printed CLI that ships the `sql` MCP tool (Store-backed CLIs).

## Fix

Capture and surface all three failure points via `mcplib.NewToolResultError`.

## Proof (generator-output)

- New `TestGenerateMCPSQLToolSurfacesRowErrors`: generates a Store+MCP CLI, pins the emitted error checks, and runs `requireGeneratedCompiles` so the `cols, err :=` redeclaration (reusing the `err` already bound by `db.Query`) can't regress the build.
- `go test ./...` — pass
- `scripts/golden.sh verify` — pass (regenerated the 4 affected `tools.go` fixtures; diff is exactly the three error checks)
- `scripts/verify-generator-output.sh` — pass (9 cases, compiles emitted modules)
- `golangci-lint run ./internal/generator/...` — 0 issues

Surfaced by Greptile review on library publish PR mvanhorn/printing-press-library#1051 (the per-CLI copy is fixed there in lockstep).

### DIFF
--- a/internal/generator/generator_test.go
+++ b/internal/generator/generator_test.go
@@ -3746,6 +3746,36 @@ func TestGenerateMCPSQLToolUsesReadOnlyStore(t *testing.T) {
 		"behavioral coverage of comment-prefix and statement-separator bypass vectors must ship into every printed CLI's mcp package")
 }
 
+// TestGenerateMCPSQLToolSurfacesRowErrors pins the emitted handleSQL error
+// handling so a regression to the silent-discard form (which returns a
+// truncated result set as a successful tool call) fails fast.
+func TestGenerateMCPSQLToolSurfacesRowErrors(t *testing.T) {
+	t.Parallel()
+
+	apiSpec := minimalSpec("rowerr-canary")
+	outputDir := filepath.Join(t.TempDir(), naming.CLI(apiSpec.Name))
+	gen := New(apiSpec, outputDir)
+	gen.VisionSet = VisionTemplateSet{Store: true, Search: true, MCP: true}
+	require.NoError(t, gen.Generate())
+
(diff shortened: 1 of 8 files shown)


tmchow:
Addressed the remaining \`json.MarshalIndent\` discard in adefd7a1.

The same \`data, _ := json.MarshalIndent(...)\` discard existed in three sibling handlers (\`handleSQL\`, \`handleSearch\`, \`handleContext\`), so rather than patch one I extracted a shared \`toolResultJSON(v any)\` helper that returns a tool error on a marshal failure, and routed all three through it. The endpoint-mirror envelope marshals at lines ~627/633 use compact \`json.Marshal\` of pre-validated \`json.RawMessage\`/base64 in a different handler layer (raw API passthrough, not pretty-printed local data) and can't fail in practice, so they're intentionally left as-is.

Extended \`TestGenerateMCPSQLToolSurfacesRowErrors\` to pin the helper + routed call sites; regenerated the affected golden \`tools.go\` fixtures (6 cases). Full proof cycle re-run green: \`go test ./...\`, \`golden.sh verify\` (31), \`verify-generator-output.sh\` (9), \`golangci-lint\` (0).

PR #2654, opened 2026-06-05
fix(generator): rely on installer default bin directory

tmchow:
## Summary
- remove explicit `--bin-dir ~/.local/bin` from generated skill and OpenClaw README install snippets
- document the installer's platform-aware default bin locations instead
- update generator tests to expect the default-bin behavior

## Verification
- `git diff --check`


tmchow:
Addressed Greptile feedback in 0f1daa9: regenerated the generator goldens from the updated templates so README/SKILL fixtures no longer carry the stale `--bin-dir ~/.local/bin` text.\n\nLocal verification:\n- `scripts/golden.sh verify` passed (25 cases)\n- `go test ./internal/generator` passed\n- `git diff --check` passed

PR #2662, opened 2026-06-05
docs(generator): install binaries for Hermes skills

tmchow:
## Summary
- add the binary install step to generated README Hermes instructions before installing the focused skill
- keep Hermes on the official `hermes skills install` flow for skill installation
- rely on the installer default per-user bin directory instead of hardcoding `--bin-dir`

## Verification
- `scripts/golden.sh verify`
- `git diff --check`

Follow-up to #2654: the OpenClaw section installed both binary and skill, but the Hermes section only installed the skill.


tmchow:
@mergifyio refresh

PR #2692, opened 2026-06-06
fix(cli): verify-skill detects alias-receiver flags and fixes positional/flag-value tokenization

### LINKED ISSUE #2412: Verify-skill walker: bind optional positionals before sibling resolution
## Summary

Verify-skill's command-resolution walker treats top-level resource commands as candidate sibling subcommands of a deeper command, even when that deeper command has declared optional positionals via Cobra's standard `Use: "leaf [name]"` syntax. This produces misleading `Use: "<sibling>" expects 0–0` errors that look like a CLI defect but are scorer misbehavior.

## Reproducer (from a real retro)

Generated `bir-pp-cli` with a spec declaring `rmc`, `rr`, `rmo` resources at the root. Three hand-coded stubs nested under `issuances` declare:

```go
Use: "cited-by [type] [number-year]"
Use: "supersedes [type] [number-year]"
Use: "superseded-by [type] [number-year]"
```

README/SKILL examples include `bir-pp-cli issuances cited-by rr 7-2003` (two positional args: type=rr, number-year=7-2003). Verify-skill emits:

```
[positional-args] bir-pp-cli issuances cited-by rr: got 1 positional args; Use: "rr" expects 0–0
  evidence: README.md: 7-2003
```

The walker has resolved `rr` against the **top-level** `rr` resource command (which expects 0 positionals) instead of binding it to cited-by's `[type]` positional. The reported `Use: "rr"` is for the wrong command — `cited-by`'s `Use:` is the one that should be consulted.

8 instances of this pattern fired in the session — 4 unique command paths × 2 (SKILL.md + README.md). All same root cause.

## Why this is scorer behavior, not a CLI defect

- Cobra's `Use: "leaf [name]"` declaration is canonical optional-positional syntax. It's what the printing-press skill itself documents (see `skills/printing-press/SKILL.md` "Verify-friendly RunE template" → square brackets for optional positionals).
- The polish skill (`printing-press-polish`) independently classified this as a verify-skill false-positive after running its own diagnostic loop.
- A counter-check confirms the CLI's runtime behavior is correct: `bir-pp-cli issuances cited-by rr 7-2003` returns the expected `not yet implemented` error (stub) — no Cobra parse error, no usage refusal. The arg parsing works at runtime; only verify-skill's static walker disagrees.

## Cross-CLI applicability

Triggers whenever a printed CLI has all three of:

1. A stub or pending command with optional positionals (`Use: "leaf [first] [second]"`),
2. A sibling resource at the root of the CLI's command tree, AND
3. The example phrasing uses a positional value that happens to match a sibling resource name (common when the positional is a "type" selector and sibling resources are per-type).

This is broadly applicable across the catalog. Examples where this pattern would trigger if the CLI had pending stubs:

- `linear-pp-cli` has top-level `teams`, `issues`, `cycles`. Any stub/recipe using `teams` as a positional ID resolves to the top-level command.
- `github-pp-cli` has top-level `repos`, `issues`, `users`. Same pattern.
- `stripe-pp-cli` has top-level `customers`, `charges`, `subscriptions`. Same pattern.

Reprints with pending features (the post-publish flow) routinely have stubs while the next-pass features are being hand-coded — this is the failure window.

## Proposed fix

In the verify-skill walker (likely `internal/verifyskill/...` or wherever the per-example resolver lives), when walking example tokens after the resolved command path:

1. Parse the current command's `Use:` string for bracketed `[name]` (optional) and `<name>` (required) positional tokens. Count them as `N_optional` and `N_required`.
2. Bind the next token to a positional first if positionals are available (`required + optional` budget remaining). Decrement budget.
3. Only fall through to sibling-command resolution when positional budget is exhausted.
4. When positional count is exceeded, emit a precise mismatch message referencing the **current command's** declared positionals, not a sibling's:

   ```
   got K positional args; cited-by declares [type] [number-year] (max 2)
   ```

   instead of:

   ```
   got 1 positional args; Use: "rr" expects 0–0
   ```

## Tests

**Positive:** Fixture CLI with `Use: "leaf [first] [second]"`, example `binary parent leaf foo bar` → verify-skill reports clean.

**Negative:** Same CLI, example `binary parent leaf foo bar baz extra` → verify-skill reports `got 4 positional args; leaf declares [first] [second] (max 2)`.

**Regression guard:** Fixture CLI with `Use: "leaf"` (no positionals) and example `binary parent leaf foo` still correctly resolves `foo` against any sibling subcommand if one exists, OR emits a clean "too many positional args" if not.

## Scope boundary

- Does NOT change Cobra positional declaration syntax (still bracketed in `Use:`).
- Does NOT change how examples are sourced (still README/SKILL recipes).
- Does NOT silence legitimate "wrong sibling" errors when the leaf truly has 0 positionals declared.

## Complexity

Small — localized walker change; the bind-positionals-first rule is a single ordering change with clear acceptance criteria.

## Context

Retro from generating `bir-pp-cli` (PH Bureau of Internal Revenue documentation index, single-site first print of a planned 13-site `ph-biz-cli`). Full retro doc preserved at `manuscripts/bir/20260527-180904/proofs/20260527-200053-retro-bir-pp-cli.md`. 1 finding survived strict Phase 3 Step G triage; 6 Skips and 1 Drop also documented in the retro for triage auditability.

---

*Filed via `/printing-press-retro` after Phase 5.5 polish completed with `further_polish_recommended: no`.*

### LINKED ISSUE #2428: [scorer] verify-skill tokenizer needs shell-quoting awareness for filter examples
**Component:** `scorer` (`verify-skill` tokenizer)
**Priority:** P3

## Summary

`verify-skill` parses SKILL.md and README.md examples to validate that every example command + flag actually exists in the shipped CLI. The current tokenizer doesn't respect double-quoted strings containing shell-special characters (`&`, `|`, `(`, `)`).

Example that triggered the bug:

```
todoist-pp-cli reschedule-cascade --filter "overdue & p1" --shift +2d --respect-deadlines --dry-run
```

verify-skill reported:

```
[positional-args] todoist-pp-cli reschedule-cascade: got 3 positional args; Use: "reschedule-cascade" expects 0–0
  evidence: SKILL.md: & p1 &
```

The parser saw `"overdue`, `&`, `p1"`, then `&` (from a different occurrence) as separate tokens, treating them as positional args.

## Why this is a P3

- Workaround: rewrite examples to avoid `&`/`|` inside quotes (loses pedagogical value of complex filter syntax demos).
- Affects any CLI with quoted filter/search examples: Todoist, GitHub search (`is:open AND label:bug`), Linear filter strings, Notion database query filters, Sentry/Datadog search syntax.

## Acceptance criteria

**Positive:** `cli foo --filter "a & b"` parses as `[cli, foo, --filter, "a & b"]` — zero positional args.

**Negative:** `cli foo a b c` still parses as 3 positional args. `cli foo --filter "a" b c` still detects `b c` as 2 positional args (i.e., quoting awareness doesn't accidentally collapse unquoted tokens).

## Likely fix direction

Adopt POSIX shell-quoting parsing. The simpler `github.com/google/shlex` is standard for this in Go; `mvdan.cc/sh/v3/syntax` is overkill. Either works.

## Test

Regen any CLI whose research.json includes a recipe with a quoted filter containing `&`. `verify-skill` should pass without rewriting the example to avoid shell-special chars.

## Artifacts
- Retro doc: https://files.catbox.moe/nzaruw.md
- Manuscripts (research, proofs): https://files.catbox.moe/38mw3c.zip
- CLI source (no binary, no vendor/): https://files.catbox.moe/ly0cxx.zip
- Source CLI: `todoist-pp-cli` (Todoist API v1 run `20260527-151452`)

### LINKED ISSUE #2452: scorer: verify-skill misses flags declared via 'f := cmd.Flags()' alias receiver
**Component:** scorer (verify-skill flag-declaration detection)
**Priority:** P2
**Related:** #1004 (adjacent verify-skill SKILL.md command-path matching)

Surfaced during a Printing Press retro on the **meta-marketing** CLI (Graph v25.0, hand-authored internal spec). Full triage retro saved locally; key fields inline below.

## What happened
All hand-written commands registered flags via the alias receiver: `f := cmd.Flags(); f.StringVar(&x, "level", ...)`. verify-skill emitted **62 errors** ("`--level` referenced in SKILL.md but not declared in any internal/cli/*.go") and failed the leg, blocking the ship verdict (6/6 → 5/6). The flags ARE declared — just via the alias, not the literal `cmd.Flags().StringVar(...)` chain the detector greps for. They appear correctly in `--help`.

## Root cause
Flag-declaration detection is a static text match on `cmd.Flags().<Type>Var("name"`. The `f := cmd.Flags()` then `f.<Type>Var(...)` form — standard Go when a command registers many flags — is invisible to it. Generated endpoint-mirror commands always use the direct chain (never trip it); hand-written novel commands (built in *every* run) frequently use the alias.

## Why it matters
This is an **API-independent Go idiom**, not an API-shape quirk. A scorer that false-fails valid idiomatic Go erodes trust and can drive an agent to "fix" SKILL.md by deleting valid flag references — making the docs *worse*.

## Durable fix
Prefer enumerating flags by running `<binary> <command> --help` and parsing (robust). Cheaper alternative: track `x := cmd.Flags()` / `cmd.PersistentFlags()` assignments and match `x.<Type>Var("name"`.

## Acceptance
- positive: an alias-registered `--foo` referenced in SKILL.md passes.
- negative: a SKILL reference to a flag no command declares still fails.

## Evidence
shipcheck-run1 verify-skill: 62 `[flag-names]`/`[flag-commands]` errors for flags that exist; resolved only by rewriting alias→direct form across 7 files.

vinnyp:
## Intent

Three independent verify-skill defects, each surfaced as a false-negative against a real printed-CLI cookbook recipe.

Issue: Fixes #2452. Fixes #2412. Fixes #2428.

## Approach

- **#2452** — verify-skill's flag-discovery walked `cmd.Flags().StringVar(...)` (inline-receiver) but missed the very common `f := cmd.Flags(); f.StringVar(...)` form (alias-receiver). The walker now resolves alias receivers back to the underlying `*pflag.FlagSet`, so alias-form flags are discovered alongside inline-form ones.
- **#2412** — the command-path walker previously attempted sibling-command descent before binding optional positionals on the current command, mis-resolving recipes that legitimately use a positional + a sibling-shadow name. It now binds optionals first, then descends.
- **#2428** — the tokenizer treated `--flag value` as `--flag` + a positional `value`, which broke filter-example parsing (e.g. `--filter 'venue=brooklyn'`). The tokenizer now consumes the next token as the flag's value when the flag is value-bearing.

Updates the canonical `scripts/verify-skill/verify_skill.py` and the bundled `internal/cli/verify_skill_bundled.py` in lockstep — `TestVerifySkillScriptInSync` covers drift between the two. The library repo's `.github/scripts` copy will need the same update; the existing `verify-skill-drift-check` workflow tracks that.

## Scope

Primary area: scorer / verify-skill

Why this belongs in this repo: verify-skill itself lives here (the generator's scorer). The library repo carries a synced copy that will follow via drift-check.

## Catalog Justification

N/A

## Risk

Moderate-low. All three fixes are inside the Python walker/tokenizer; existing behavior is preserved (no flag/positional shape that previously parsed now stops parsing — we only resolve cases that previously didn't). The canonical-vs-bundled drift test pins parity.

## Output Contract

N/A — verdicts shift only where previously false-negative (three specific tokenizer/walker gaps); no schema/output format change.

## Verification

- `go test ./internal/cli/...` — `TestVerifySkillScriptInSync` green (canonical + bundled stay in lockstep).
- `python scripts/verify-skill/verify_skill.py …` exercised against the cookbook examples that originally surfaced each bug; all three now resolve correctly.
- Walked the alias-receiver, positional-vs-sibling, and `--flag value` cases by hand against the new walker.

(No generator/template surface touched.)

## AI / Automation Disclosure

Implementation generated by an AI coding agent (codex), human-orchestrated and verified end-to-end before submission.

### DIFF
--- a/internal/cli/verify_skill_bundled.py
+++ b/internal/cli/verify_skill_bundled.py
@@ -462,6 +462,55 @@ def iter_flag_declarations(text: str) -> Iterable[tuple[bool, str]]:
         yield aliases[m.group(1)], m.group(3)
 
 
+def _iter_bool_flag_names(text: str) -> Iterable[str]:
+    """Long-names of boolean flags declared in text (BoolVar/BoolVarP and the
+    alias-receiver form). Mirrors iter_flag_declarations' direct + alias scan but
+    keeps only the boolean methods, so the recipe tokenizer can tell a value-less
+    boolean flag (`--json <positional>`) from a value-bearing one
+    (`--filter <value>`)."""
+    for m in FLAG_DECL_RE.finditer(text):
+        _persistent, method, name = m.groups()
+        if method in ("BoolVar", "BoolSliceVar"):
+            yield name
+
+    aliases = {
+        m.group(1): m.group(2) == "Persistent"
+        for m in FLAG_ALIAS_RE.finditer(text)
+    }
+    if not aliases:
+        return
+
+    alias_decl_re = re.compile(
+        r'\b(' + "|".join(re.escape(alias) for alias in aliases) + r')\.'
+        r'(' + FLAG_METHOD_PATTERN + r')P?\('
+        r'&[^,]+,\s*"([a-z][a-z0-9-]*)"'
+    )
+    for m in alias_decl_re.finditer(text):
+        if m.group(2) in ("BoolVar", "BoolSliceVar"):
+            yield m.group(3)
+
+
+@lru_cache(maxsize=None)
+def _boolean_flag_names(cli_dir: Path) -> frozenset[str]:
+    """Long-names of every boolean flag declared in the CLI's internal/cli/*.go
+    (cached per cli_dir). The recipe tokenizer consults this so it never consumes
+    the token after a value-less boolean flag as that flag's value — doing so
+    would silently drop a real positional. A CLI-wide scan (rather than
+    per-command) deliberately includes persistent/root booleans like --json or
+    --verbose, which are the common case a recipe writes before a positional."""
+    cli_pkg = cli_dir / "internal" / "cli"
+    if not cli_pkg.is_dir():
+        return frozenset()
+    names: set[str] = set()
+    for path in sorted(cli_pkg.glob("*.go")):
+        try:
+            text = read_utf8(path)
+        except Exception:
+            continue
+        names.update(_iter_bool_flag_names(text))
+    return frozenset(names)
+
+
 def flag_declared_in(files: Iterable[Path], flag_name: str) -> bool:
     for f in files:
         try:
@@ -680,11 +729,24 @@ def _cli_invocation_from_tokens(
             i += 1
             continue
         if t.startswith("--"):
-            flags.append(t.split("=", 1)[0])
-            # Skip a space-separated value (`--flag value`), but NOT when the
-            # value is inline (`--flag=value`) — there the next token is a
-            # positional, not this flag's value.
-            if "=" not in t and i + 1 < len(tokens) and not tokens[i + 1].startswith("-"):
+            flag_name = t.split("=", 1)[0]
+            flags.append(flag_name)
+            # Skip a space-separated value (`--flag value`), but NOT when:
+            #  - the value is inline (`--flag=value`) — the next token is a
+            #    positional, not this flag's value; or
+            #  - the flag is a known boolean flag, which takes no value, so the
+            #    next token is a positional (consuming it would under-count the
+            #    recipe's positional args).
+            is_bool = (
+                cli_dir is not None
+                and flag_name.lstrip("-") in _boolean_flag_names(cli_dir)
+            )
+            if (
+                "=" not in t
+                and not is_bool
+                and i + 1 < len(tokens)
+                and not tokens[i + 1].startswith("-")
+            ):
                 i += 2
                 continue
         elif t.startswith("-"):
--- a/internal/cli/verify_skill_bundled.py
+++ b/internal/cli/verify_skill_bundled.py
@@ -87,15 +87,19 @@ USE_RE = re.compile(r'Use:\s*"([^"]+)"')
 ARGS_RE = re.compile(
     r'Args:\s*cobra\.(ExactArgs|MinimumNArgs|MaximumNArgs|RangeArgs|NoArgs|OnlyValidArgs|ExactValidArgs)\s*\(([^)]*)\)'
 )
-FLAG_DECL_RE = re.compile(
-    r'(Persistent)?Flags\(\)\.'
-    r'(StringVar|BoolVar|IntVar|Int32Var|Int64Var|Float32Var|Float64Var|DurationVar|'
+FLAG_METHOD_PATTERN = (
+    r'StringVar|BoolVar|IntVar|Int32Var|Int64Var|Float32Var|Float64Var|DurationVar|'
     r'StringSliceVar|StringArrayVar|IntSliceVar|Int32SliceVar|Int64SliceVar|'
     r'Float64SliceVar|BoolSliceVar|DurationSliceVar|'
     r'UintVar|Uint32Var|Uint64Var|UintSliceVar|IPVar|IPSliceVar|Float32SliceVar|'
-    r'StringToStringVar|StringToIntVar|StringToInt64Var)P?\('
+    r'StringToStringVar|StringToIntVar|StringToInt64Var'
+)
+FLAG_DECL_RE = re.compile(
+    r'(Persistent)?Flags\(\)\.'
+    r'(' + FLAG_METHOD_PATTERN + r')P?\('
     r'&[^,]+,\s*"([a-z][a-z0-9-]*)"'
 )
+FLAG_ALIAS_RE = re.compile(r'\b([A-Za-z_]\w*)\s*(?::=|=)\s*cmd\.(Persistent)?Flags\(\)')
 @dataclass
 class Finding:
     check: str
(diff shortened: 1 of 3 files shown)


tmchow:
@vinnyp address greptile comment

"The alias-receiver and optional-positional fixes are logically correct and well-tested. The --flag value tokenizer fix works correctly for value-bearing flags. However, the same i+=2 skip applies unconditionally to boolean long flags, which can drop a real positional that follows a boolean flag.

The _cli_invocation_from_tokens function in scripts/verify-skill/verify_skill.py (and its bundled copy) around the new i+=2 long-flag skip warrants a second look for the boolean-flag case."

vinnyp:
@tmchow Addressed in `3668923c`.

The tokenizer (`_cli_invocation_from_tokens`) skipped the token after any value-less long flag as that flag's value — correct for `--filter <value>`, but for a boolean flag (`--json <positional>`) it silently dropped the positional, under-counting the recipe's args in the positional-args check.

Fix: added a CLI-wide boolean-flag name set (`BoolVar`/`BoolVarP` + the alias-receiver form this PR introduced — deliberately including persistent/root booleans like `--json`/`--verbose`, the common case), and the space-separated-value skip now fires only when the flag is **not** a known boolean. Value-bearing flags are unchanged.

Added a regression test (`test_boolean_long_flag_does_not_swallow_following_positional`) that asserts both branches against a synthetic CLI carrying one `BoolVar` and one `StringVar` flag. `verify_skill_bundled.py` re-synced (sha256 sync test green); branch rebased onto current `main`.

tmchow (approved):

Issue #2387, opened 2026-05-26
cli: printed CLI's MCPB manifest reports Press version (4.11.0) instead of release tag; MCP serverInfo hardcodes 1.0.0


TUCU-Adam:
## Summary

Two related version-leak bugs surface in every printed CLI's MCP surface:

### Bug 1: MCPB `manifest.json` reports the Press version, not the printed CLI's release version

`internal/pipeline/mcpb_manifest.go:232` — `bundleVersion()` returns `m.PrintingPressVersion` as the manifest's `version` field. After tagging the CIPP printed CLI at `v0.1.4`, the installed Claude Desktop connector shows version `4.11.0` (the Press release that generated it), not `0.1.4`.

The comment at lines 228-231 claims this is intentional ("tracks the printing-press release that produced it so Claude Desktop's update detection sees a fresh value on regeneration") — but that rationale is wrong. MCPB `version` semantics are "this bundle's version" (per [mcpb spec](https://github.com/modelcontextprotocol/mcpb/blob/main/MANIFEST.md)). The right update-detection signal is the printed CLI's release tag — a new tag is a meaningful update, a Press regen alone isn't.

**Verified:** `/Users/adamthorn/printing-press/library/cipp/manifest.json` line 5 → `"version": "4.11.0"`. Same in the deployed `TUCU-Managed-IT-Services/cipp-pp-cli/manifest.json` shipped in v0.1.4's `.mcpb`.

`CLIManifest` (in `internal/pipeline/climanifest.go:61-`) has no field for the printed CLI's own release version — only `PrintingPressVersion` (generator), `APIVersion` (spec.info.version, explicitly marked "not the CLI version"), and `GeneratedAt`. The release tag isn't known at `cli-printing-press generate` time, so the right shape is to inject it at MCPB-bundle time.

### Bug 2: MCP server reports hardcoded `1.0.0` via protocol-level `serverInfo`

`internal/generator/templates/main_mcp.go.tmpl:31,84` — both `server.NewMCPServer(...)` calls pass literal `"1.0.0"`. The CLI binary's version IS ldflag-injected from the git tag (see `goreleaser.yaml.tmpl:12` — `-X .../internal/cli.version={{ .Version }}`), but the MCP binary's goreleaser entry at lines 21-27 has no `-X` ldflag, so every installed MCP server reports `1.0.0` to MCP clients forever.

## Proposed fix (5 parts)

[...]

PR #2699, opened 2026-06-06
fix(cli): inject MCP server version via ldflag instead of hardcoded 1.0.0

### LINKED ISSUE #2554: generator: MCP server reports a hardcoded "1.0.0" and is silent on stdio startup — no way to identify the running build [retro]
**Source:** retro — dogfooding the published `splitwise` MCP in Claude Cowork / Desktop (2026-05-31).

## Symptom

When iterating on multiple builds of a printed CLI's MCP (e.g. the published `4.19.0` vs a local integration build), there is **no way to tell which build a host actually loaded**:

- The host's MCP log has **no startup line** for the server (on stdio).
- The MCP `initialize` handshake reports a **hardcoded `"1.0.0"`** regardless of the real release.

## Root cause

The generated `cmd/<slug>-pp-mcp/main.go`:

1. Passes a **literal `"1.0.0"`** to `server.NewMCPServer(name, "1.0.0", …)`, ignoring the ldflags-injected `internal/cli.version` that the CLI already reports via `--version`.
2. On the **stdio** transport, writes **nothing** to stderr at startup (only the `http` branch logs a line). Hosts capture MCP-server stderr into their logs, so stdio servers are silent there.

## Affected

Generator-shaped, cross-CLI: every printed CLI's MCP binary hardcodes the version and is silent on stdio startup.

## Proposed fix

The generated MCP `main.go` should:

1. **Report the real build version** — wire `NewMCPServer` to the same ldflags var the CLI uses (`internal/cli.version`, via a small exported accessor). Goreleaser already injects `-X …/internal/cli.version={{ .Version }}` into **all** binaries, so this makes the published MCP handshake report the real release version with **no goreleaser change**.
2. **Write a startup banner to stderr** on load (both transports): `"<binary> <version> starting (transport=<t>)"`, so the host's MCP log identifies exactly which build is running.

Fixed locally in the published `splitwise` CLI (patch id `mcp-version-startup-banner`): added `cli.Version()`, wired it into `NewMCPServer`, and added a `startupBanner` printed to stderr.

## Related

- Same Cowork/Desktop dogfood session: #2515 (typed-tool firehose), #2516 (sql description), #2517 (cobratree positionals/flags), #2553 (app-name tool discovery).

**Severity:** P2 (observability — makes multi-build dogfooding and field debugging tractable).

vinnyp:
fix(cli): inject MCP server version via ldflag instead of hardcoded 1.0.0

The generated MCP server passed a literal "1.0.0" to server.NewMCPServer in both
transport branches, so the MCP binary always reported 1.0.0 regardless of the
release tag. Declare a package-level `var version = "1.0.0"` (overridable at build
time) and inject it on the MCP goreleaser build entry via `-X main.version`,
mirroring how the CLI binary's version is injected. Covers Bug 2 of #2387 (MCP
server-info version); the MCPB-manifest version (#2387 Bug 1) needs
release-tag-at-bundle-time design and is left out.

Fixes #2554.
Refs #2387.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/internal/generator/device.go
+++ b/internal/generator/device.go
@@ -2532,10 +2532,13 @@ import (
 	mcptools "{{.ModulePath}}/internal/mcp"
 )
 
+// version is the printed MCP server's version, overridable at build time via ldflags.
+var version = "1.0.0"
+
 func main() {
 	s := server.NewMCPServer(
 		{{quote .DisplayName}},
-		"1.0.0",
+		version,
 		server.WithToolCapabilities(false),
 	)
 	mcptools.RegisterTools(s)
--- a/internal/generator/device_test.go
+++ b/internal/generator/device_test.go
@@ -113,6 +113,7 @@ func TestGeneratedBLEDeviceEmitsPublishArtifacts(t *testing.T) {
 	assert.Contains(t, goreleaser, "main: ./cmd/"+naming.CLI(ds.Name))
 	assert.Contains(t, goreleaser, "main: ./cmd/"+naming.MCP(ds.Name))
 	assert.Contains(t, goreleaser, naming.CLI(ds.Name)+"/internal/cli.version=")
+	assert.Contains(t, goreleaser, "-X main.version={{ .Version }}")
 	assert.Contains(t, goreleaser, `description: "`)
 
 	// AGENTS.md is the device-aware variant: it uses BLE/replay concepts and the
--- a/internal/generator/device_test.go
+++ b/internal/generator/device_test.go
@@ -163,6 +163,15 @@ func TestGeneratedBLEDeviceEmitsMCPSurface(t *testing.T) {
 	assert.FileExists(t, filepath.Join(outputDir, "cmd", naming.MCP(ds.Name), "main.go"))
 	assert.FileExists(t, filepath.Join(outputDir, "internal", "mcp", "cobratree", "walker.go"))
 
+	mcpMainSrc, err := os.ReadFile(filepath.Join(outputDir, "cmd", naming.MCP(ds.Name), "main.go"))
+	require.NoError(t, err)
+	mcpMain := string(mcpMainSrc)
+	// The MCP server version must be an ldflag-overridable var, not a hardcoded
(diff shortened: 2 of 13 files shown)


tmchow:
@vinnyp Address the Greptile comment:

> The template and goreleaser changes are correct and complete. However, testdata/golden/expected/generate-device-ble/ble-temperature-sensor/cmd/ble-temperature-sensor-pp-mcp/main.go was not updated — it still has the hardcoded "1.0.0" literal and no var version declaration. Any golden-file test that compares generated output against this file will fail on the current branch.
> 
> testdata/golden/expected/generate-device-ble/ble-temperature-sensor/cmd/ble-temperature-sensor-pp-mcp/main.go — the only MCP main.go golden not updated to match the new template.

vinnyp:
@tmchow Addressed — and it turned out deeper than a stale golden.

The device MCP `main.go` isn't rendered from `main_mcp.go.tmpl` (the template #2699 updated). Device CLIs use a separate inline template, `deviceMCPMainTemplate` in `internal/generator/device.go`, which still hardcoded `"1.0.0"`. Because #2699 had already added `-X main.version={{ .Version }}` to the device `.goreleaser.yaml`, the ldflag injection was a **silent no-op** — the device MCP binary would always report `1.0.0` regardless of release version. Updating only the committed golden would have made it diverge from real generator output.

Fix (`34aa7e61`):
- `device.go`: `deviceMCPMainTemplate` now declares `var version = "1.0.0"` and passes `version` to `NewMCPServer`, matching `main_mcp.go.tmpl`.
- Regenerated the `ble-temperature-sensor-pp-mcp/main.go` golden to match.
- Added an assertion in `device_test.go` that the device MCP main uses the version var (not a bare literal) — mirrors the `assertMCPMainUsesVersionVar` this PR added on the API side, so this path can't silently drift again.

Branch rebased onto current `main`; `golden.sh verify` (31 cases), generator tests, and build all green.

PR #2720, opened 2026-06-06
fix(generator): don't persist env-sourced credentials into config.toml (#2710)

### LINKED ISSUE #2710: generated OAuth clients: SaveTokens persists env-sourced credential overrides into config.toml, and a static extra-credential shadows the refresh-managed token (permanent 401 after expiry)
Found while live-validating the google-ads sync fix (printing-press-library#1065). Two composable hazards in the generated OAuth + extra-credentials auth path left the CLI permanently 401 even though its refresh grant was alive and refreshing correctly.

## Incident (google-ads-pp-cli, real config)

1. A dogfood/verify-style run exported dummy credentials (`GOOGLE_ADS_ACCESS_TOKEN=dummy`, `GOOGLE_ADS_DEVELOPER_TOKEN=dummy`) in a process pointed at the user's real `~/.config/google-ads-pp-cli/config.toml`.
2. `config.Load` copies env overrides into the Config struct (by design). But a later token refresh called `Config.SaveTokens(...)`, which **rewrites the whole config file from the struct — env-sourced values included**. Result: the real `ads_developer_token` (and `ads_access_token`) were replaced by the literal string `"dummy"` on disk. The real developer token had to be recovered from another machine's config copy.
3. Independently: `Config.AuthHeader()` gives a non-empty `GoogleAdsAccessToken` (the *extra credential* slot, `ads_access_token` in TOML) **priority over the refresh-managed `AccessToken`** — even when it came from the config file rather than env (the code comment says "Env-var token wins" but the check is on the field, not the source). A static OAuth access token expires in ~1h, so any config that carries `ads_access_token` works during setup and then 401s forever after, while the auto-refresh machinery keeps minting perfectly good tokens into `access_token` that are never used.

Net effect: every live call returned HTTP 401 while `auth status` said authenticated and the refresh loop was demonstrably healthy (tokeninfo on the refreshed `access_token`: valid, `adwords` scope). Diagnosis was further slowed because the client truncates Google's error body before the `details` array, hiding `DEVELOPER_TOKEN_INVALID` behind Google's misleading top-level "Request is missing required authentication credential" message.

## Expected generator behavior

- **SaveTokens must not persist env-sourced values.** Either track provenance per field (only write back file-sourced/refresh-produced fields: access_token, refresh_token, token_expiry, client_id, client_secret), or re-read the file and patch only the token fields. An env override should never outlive the process.
- **Refresh-managed OAuth should outrank a file-stored static extra-credential** (or at minimum, an expired static token should fall through to the refreshable pair instead of being sent forever). Env-sourced override keeping priority is fine — that's the explicit escape hatch.
- Nice-to-have: don't truncate the error body before structured `details` (the `errorCode.authenticationError` is the actionable part of Google Ads 401s).

## Pointers

- `library/marketing/google-ads/internal/config/config.go` — `Load` env overrides, `AuthHeader()` priority, `SaveTokens`
- `library/marketing/google-ads/internal/client/client.go` — `refreshAccessToken()` → `SaveTokens` (the persistence trigger)
- Incident write-up lives in the operator's local runbook; happy to add detail here if useful.

Affects any printed CLI that combines OAuth refresh + env credential overrides + extra credential slots — the google-ads shape is just where it fired first.

vinnyp:
`config.Load` wrote env-var values (e.g. a CI `<PREFIX>_ACCESS_TOKEN=dummy`) directly onto config fields with no provenance, and `save()` (called by `SaveTokens`/`SaveCredential`/`SaveBearerToken`/`ClearTokens`) serialized the whole struct — so a transient env credential **permanently overwrote the real on-disk credential** in `config.toml`.

The fix tracks env-sourced fields (unexported `envOverrides`) against a snapshot of the on-disk config (`fileConfig`). `configForSave` restores env-overridden fields to their on-disk value before marshaling, so env values never reach disk, while env precedence still wins for the running process. Each `Save*`/`ClearTokens` mutator deletes the env-override for every field it legitimately writes (before updating the snapshot), so refreshed tokens persist and an explicit logout clears the on-disk credentials — including builtin-field collisions like `<PREFIX>_ACCESS_TOKEN` → `AccessToken`.

Tests (`auth_env_precedence_test.go`) run real generated CLIs and assert: a plain save never leaks an env value; a refreshed token persists over a colliding env override; `ClearTokens` empties the on-disk credentials; and a no-env load+save round-trips unchanged. Verified across oauth auth-code, bearer+refresh, client_credentials, api_key, and composed additional-header shapes.

Note: correctness relies on the `Load` markers, `configForSave`, `ClearTokens`, and `updateFileConfigField` keeping the same env-field ranges in sync; a future field added to one but not the others should be caught by extending the collision tests. (A single-rule "authoritative write-set" refactor is a reasonable follow-up.)

Closes #2710

### DIFF
--- a/internal/generator/auth_env_precedence_test.go
+++ b/internal/generator/auth_env_precedence_test.go
@@ -322,6 +322,109 @@ func TestSaveBearerTokenPersistsOverBuiltinEnvOverride(t *testing.T) {
 	runGoCommand(t, outputDir, "test", "./internal/config", "-run", "TestSaveBearerTokenPersistsOverBuiltinEnvOverride")
 }
 
+// TestConfigSaveBearerTokenClearsNonBuiltinEnvCollision covers the #2720 review
+// gap: SaveBearerToken zeroed non-builtin custom env-var fields in memory but
+// did not delete their envOverrides entries, so configForSave restored the stale
+// on-disk value and the credential field was never cleared from config.toml on
+// refresh. Also exercises the fileConfig map-isolation (deep-copy) fix.
+func TestConfigSaveBearerTokenClearsNonBuiltinEnvCollision(t *testing.T) {
+	t.Parallel()
+
+	apiSpec := minimalSpec("bearer-refresh-custom-env")
+	apiSpec.Auth = spec.AuthConfig{
+		Type:   "bearer_token",
+		Header: "Authorization",
+		Format: "Bearer {access_token}",
+		EnvVarSpecs: []spec.AuthEnvVar{
+			{Name: "API_TENANT", Kind: spec.AuthEnvVarKindPerCall, Required: false, Sensitive: true},
+		},
+	}
+	apiSpec.BearerRefresh = spec.BearerRefreshConfig{
+		BundleURL: "https://cdn.example.com/main.js",
+		Pattern:   `"(AAAAAAAA[^"]+)"`,
+	}
+
+	outputDir := filepath.Join(t.TempDir(), "bearer-refresh-custom-env-pp-cli")
+	require.NoError(t, New(apiSpec, outputDir).Generate())
+
+	const runtimeTest = `package config
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+)
+
+func TestSaveBearerTokenClearsNonBuiltinEnvOverride(t *testing.T) {
+	configPath := filepath.Join(t.TempDir(), "config.toml")
+
+	// Seed a stale on-disk value for the non-builtin custom credential field,
+	// with no env override in effect.
+	seed, err := Load(configPath)
+	if err != nil {
+		t.Fatalf("Load() error = %v", err)
+	}
+	seed.ApiTenant = "disk-tenant"
+	if err := seed.save(); err != nil {
+		t.Fatalf("seed save() error = %v", err)
+	}
+
+	// Now the env var overrides the disk value.
+	t.Setenv("API_TENANT", "env-tenant")
+	cfg, err := Load(configPath)
+	if err != nil {
+		t.Fatalf("Load() error = %v", err)
+	}
+	if cfg.ApiTenant != "env-tenant" {
+		t.Fatalf("ApiTenant after Load() = %q, want env-tenant", cfg.ApiTenant)
+	}
+
+	refreshedAt := time.Date(2032, 3, 4, 5, 6, 7, 0, time.UTC)
+	if err := cfg.SaveBearerToken("refreshed-access-token", refreshedAt); err != nil {
+		t.Fatalf("SaveBearerToken() error = %v", err)
+	}
+
+	after, err := os.ReadFile(configPath)
+	if err != nil {
+		t.Fatalf("ReadFile() error = %v", err)
+	}
+	text := string(after)
+	// On refresh the custom credential field must be cleared: neither the env
+	// value nor the stale on-disk value may survive. (Before the fix the
+	// lingering envOverride entry made configForSave write the stale disk value.)
+	for _, stale := range []string{"disk-tenant", "env-tenant"} {
+		if strings.Contains(text, stale) {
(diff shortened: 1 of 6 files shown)


tmchow:
@vinnyp have your agent look at the Greptile comment:

```
The credential-isolation logic is correct for SaveTokens, SaveCredential, and ClearTokens, but SaveBearerToken has an asymmetry: it zeros non-builtin custom env-var fields in memory without removing their envOverrides entries, so configForSave() restores stale on-disk values instead of writing empty strings for those fields.

The core mechanism is well-designed and Save*/ClearTokens mutators correctly handle the delete-then-update pattern for all builtin fields. The gap is in SaveBearerToken: non-builtin custom fields are zeroed in c but their envOverrides keys are not deleted, so configForSave() sees the override still active and writes the old disk value back. A bearer-token refresh on a spec combining BearerRefresh with at least one non-builtin custom env var spec will silently fail to clear that field on disk. The existing test suite covers the builtin-collision variant for SaveBearerToken but not this non-builtin path, so the gap is currently untested.

internal/generator/templates/config.go.tmpl — specifically the SaveBearerToken block where non-builtin EnvVarSpecs and EnvVars fields are zeroed without the corresponding delete(c.envOverrides, ...) calls that ClearTokens correctly includes for the same set of fields.
```

and the 2 isuses in "Comments outside the diff" please

vinnyp:
@tmchow Both addressed in `09b29bd5`.

**P1 (SaveBearerToken):** added the missing `delete(c.envOverrides, …)` to the non-builtin custom env-var clearing in both the `EnvVarSpecs` and legacy `EnvVars` branches, mirroring `ClearTokens`. Without it, `configForSave` saw the override still active and wrote the **stale on-disk** credential back instead of clearing the field on refresh.

**P2 (snapshotFileConfig):** added a `cloneStringMap` helper and deep-copy `Headers`/`TemplateVars`. I applied the same isolation in `save()`'s `fileConfig` assignment too — it has the identical shallow-share (`configForSave` does `out := *c`, then `save` stores it as `fileConfig`), so fixing only `snapshotFileConfig` would have left the same trap one call away.

One thing I deliberately did **not** change: `SaveBearerToken` still doesn't touch `AdditionalHeaders` custom fields (unlike `ClearTokens`, which wipes them). That asymmetry is correct — a bearer *refresh* replaces only the primary bearer; an additional-header sibling credential (e.g. an apiKey sent alongside) is unrelated and still valid, so it must survive the refresh. Leaving it in `envOverrides` keeps both the no-env-leak invariant (#2710) **and** sibling-preservation true; zeroing it the way `ClearTokens` does would be a data-loss bug on every refresh.

Regression test added (generate+compile+run): seeds a stale on-disk value for a non-builtin custom env credential, sets the env override, calls `SaveBearerToken`, and asserts neither the stale disk value nor the env value survives — plus a `snapshotFileConfig` map-isolation check. Verified it fails without the P1 delete. Goldens regenerated; branch rebased onto current `main`.

tmchow:
@vinnyp failing CI

vinnyp:
@tmchow Addressed the follow-up Greptile P1 (`SaveCredential`) in `c7c9044d`.

Same class as the SaveBearerToken fix: `SaveCredential` zeroed `AuthHeaderVal`/`AccessToken` but only deleted the *canonical* env-var's override. A non-canonical env var whose placeholder collides with the `auth_header`/`access_token` builtin tag (resolving to those fields) left its override active, so `configForSave` restored the stale on-disk value instead of the cleared `""`. Paired each builtin-field zeroing with `delete(c.envOverrides, …)` + `updateFileConfigField`, matching `ClearTokens`.

Also added the missing regression test — `SaveCredential` was the one credential-write method in the family without a builtin-collision test: an `api_key` CLI with a canonical `SVC_API_KEY` plus a non-canonical `SVC_ACCESS_TOKEN` that collides with `AccessToken`, asserting `SaveCredential` clears the alternate `AccessToken` slot from disk (neither the stale disk value nor the env value survives).

Peer-reviewed; full `go test ./internal/generator/` + golden verify green; rebased onto current `main`.

PR #2757, opened 2026-06-07
fix(generator): make store list zero limit unlimited

### LINKED ISSUE #2468: store.List(resource, 0) caps at 200 but every caller means 'all'
**Component:** generator (store template) · **Priority:** P1

**What happened**
Every BI command (`funnel-count`, `funnel-convert`, `query`, `analytics`, the resolver) calls `db.List(resource, 0)` expecting all rows, but `List` defaults `limit <= 0` to **200**. Aggregates silently undercounted over 200 records (`funnel-count` showed 200 deals instead of 870). Generated `data_source.go` even comments `// 0 = no limit, return all synced data`.

**Cross-API impact**
Universal — `List` ships in every printed CLI's store; any resource with >200 synced rows feeding a novel/analytics command undercounts silently.

**Proposed fix**
Make `limit <= 0` mean unlimited (omit the `LIMIT` clause). Keep positive limits as a cap.

**Test**
- positive: `List(x, 0)` returns all rows.  - negative: `List(x, 50)` returns 50.

_Found while printing a Pipedrive V1+V2 combo CLI (pipedrive-pp-cli, 269 endpoints, two OpenAPI specs sharing host api.pipedrive.com with /v1 and /api/v2 prefixes, V2 cursor pagination). Confirmed bug, worked around in the printed CLI; root cause is in the machine._

tmchow:
## Summary
Generated stores now treat `List(resource, 0)` and other non-positive limits as an unlimited read, matching the callers that use `0` to aggregate all synced rows. Positive limits still apply a parameterized SQL cap, so commands that intentionally request a bounded sample keep their existing behavior.

This prevents generated analytics, BI, and data-source commands from silently undercounting resources after 200 cached rows.

## Tests
- `go test ./internal/generator -run '^TestGeneratedStoreListLimitSemantics$' -count=1`
- `scripts/golden.sh verify`
- `go fmt ./...`
- `golangci-lint run ./...`
- `go test -p 1 ./...`

Closes #2468

### DIFF
--- a/internal/generator/store_list_limit_test.go
+++ b/internal/generator/store_list_limit_test.go
@@ -1,0 +1,68 @@
+package generator
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestGeneratedStoreListLimitSemantics(t *testing.T) {
+	t.Parallel()
+
+	apiSpec := minimalSpec("store-list-limit")
+	outputDir := filepath.Join(t.TempDir(), "store-list-limit-pp-cli")
+	require.NoError(t, New(apiSpec, outputDir).Generate())
+
+	testPath := filepath.Join(outputDir, "internal", "store", "store_list_limit_test.go")
+	require.NoError(t, os.WriteFile(testPath, []byte(`package store
+
+import (
+	"fmt"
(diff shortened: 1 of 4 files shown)


tmchow:
@mergifyio queue

PR #2764, opened 2026-06-07
fix(cli): apply api key auth prefixes

### LINKED ISSUE #2494: generator: api_key AuthHeader() drops auth.prefix; non-Bearer prefix APIs ship 401-broken
> Filed by `/printing-press-retro` from a run of **Make.com** (`make-pp-cli`, scorecard 94/100, verify 100%, 12/12 live tests against the real Make.com API).

## Summary

When a spec declares `auth: {type: api_key, in: header, header: "Authorization", prefix: "<word>"}`, the generated `Config.AuthHeader()` returns the raw token without the prefix word. Every API call goes out with a malformed header and gets 401. doctor reports "Auth: configured / Credentials: valid" because its check is too superficial to catch the prefix omission.

## Evidence

The spec for Make.com declared:

```yaml
auth:
  type: api_key
  in: header
  header: "Authorization"
  prefix: "Token"
  env_vars:
    - MAKE_API_TOKEN
```

The spec parser correctly populated `AuthConfig.Prefix = "Token"` (`internal/spec/spec.go:447`). But the generated `internal/config/config.go` AuthHeader() emitted:

```go
func (c *Config) AuthHeader() string {
	if c.AuthHeaderVal != "" {
		return c.AuthHeaderVal
	}
	token := c.MakeApiToken
	if token == "" {
		return ""
	}
	if c.MakeApiToken == "" {
		return ""
	}
	return token   // ← missing "Token " prefix
}
```

Live API result:

```
$ make-pp-cli users me
Error: GET /users/me returned HTTP 401: {"detail":"Access denied.","message":"User is not logged in.","code":"IM015"}
$ curl -H "Authorization: Token <token>" https://us1.make.com/api/v2/users/me
{"authUser":{"id":...,"name":"Wade Carpenter",...}}   # works
```

After patching AuthHeader() to prepend `"Token " + token`, all 12 live dogfood tests passed.

## Root cause

The generator's auth-header template has a special case for `auth.type: bearer_token` that prepends `"Bearer "`. The `api_key` branch returns the raw token regardless of `c.Prefix`. The `Prefix` field is parsed and stored on `AuthConfig` but never consumed in the api_key code path.

## Cross-API examples

| API | Auth scheme | Evidence |
|-----|-------------|----------|
| **Make.com** | `Authorization: Token <token>` | This run, live-verified |
| **Gitea** | `Authorization: token <token>` (lowercase) | [docs.gitea.com/development/api-usage](https://docs.gitea.com/development/api-usage) — same pattern |
| **Forgejo** | Same as Gitea (fork) | Inherited from Gitea |

Plus various forge/ticketing platforms whose Authorization scheme word is something other than "Bearer".

## Proposed fix

In the auth-header template's `api_key` branch (the template that emits `Config.AuthHeader()`'s body for `api_key` type):

```go
// Today (api_key branch):
return token

// Proposed:
if "<spec.Auth.Prefix>" != "" {
    return "<spec.Auth.Prefix> " + token
}
return token
```

This mirrors the existing `bearer_token` branch's prefix handling.

## Acceptance criteria

- **Positive:** A spec with `auth: {type: api_key, prefix: "Token"}` emits a `Config.AuthHeader()` that returns `"Token " + token`. A live call to a `Token`-scheme API succeeds.
- **Negative:** A spec with `auth: {type: api_key}` (no prefix, e.g., `X-API-Key` header style) regenerates byte-identical to today. The `bearer_token` path is unchanged.

## Guard

Only fires when `auth.prefix != ""`. Default behavior unchanged.

## Anti-pattern observed

doctor's "Credentials: valid" check passed even with a malformed Authorization header. The "valid" line came from a check that confirmed the env var is set, not that the assembled header is accepted by the upstream API. Renaming to "Credentials: present" — or actually round-tripping a `verify_path` probe and reporting based on that — would have surfaced the problem before any user-facing API call.

## Related issues

- #1158 — `Generated config.go AuthHeader() unconditionally overwrites AuthSource`. Sibling bug in the same function: #1158 fixes the bearer_token branch's AuthSource label overwrite; this issue fixes the api_key branch's missing prefix. Both are AuthHeader-template issues but in different branches and with different symptoms.

## Run context

- API: Make.com
- Generated: 2026-05-30
- Scorecard: 94/100 Grade A
- Library entry: `library/productivity/make/` (published as printing-press-library PR #956)

tmchow:
## Summary

Generated CLIs now honor spec-declared API-key auth prefixes such as `Authorization: Token <token>`. APIs that require a non-Bearer scheme word no longer ship with raw-token Authorization headers that 401 even though credentials are configured.

The fix stays in the Printing Press generator template, so future printed CLIs get the corrected `Config.AuthHeader()` behavior. Prefixless API-key headers continue returning the raw token, and formatted auth templates still take precedence over `auth.prefix`.

Closes #2494

## Verification

- `go test ./internal/generator -run 'TestGeneratedAPIKeyAuthHeader(AppliesPrefix|WithoutPrefixKeepsRawToken)'`
- `scripts/golden.sh verify`
- `go fmt ./...`
- `go test ./...`

### DIFF
--- a/internal/generator/auth_header_prefix_test.go
+++ b/internal/generator/auth_header_prefix_test.go
@@ -41,6 +41,46 @@ func TestAPIKeyAuthHeaderPrefix(t *testing.T) {
 	runGoCommandRequired(t, outputDir, "test", "./internal/config")
 }
 
+func TestGeneratedAPIKeyAuthHeaderORCaseAppliesPrefix(t *testing.T) {
+	t.Parallel()
+
+	apiSpec := minimalSpec("api-key-prefix-or")
+	apiSpec.Auth = spec.AuthConfig{
+		Type:   "api_key",
+		In:     "header",
+		Header: "Authorization",
+		Prefix: "Token",
+		EnvVarSpecs: []spec.AuthEnvVar{
+			{Name: "OR_PREFIX_PRIMARY_TOKEN", Kind: spec.AuthEnvVarKindPerCall, Required: false, Sensitive: true, Description: "Set this OR OR_PREFIX_FALLBACK_TOKEN."},
+			{Name: "OR_PREFIX_FALLBACK_TOKEN", Kind: spec.AuthEnvVarKindPerCall, Required: false, Sensitive: true, Description: "Set this OR OR_PREFIX_PRIMARY_TOKEN."},
+		},
+	}
+
+	outputDir := filepath.Join(t.TempDir(), "api-key-prefix-or-pp-cli")
+	require.NoError(t, New(apiSpec, outputDir).Generate())
+
+	const inlineTest = `package config
+
+import "testing"
+
+func TestAPIKeyAuthHeaderORCasePrefix(t *testing.T) {
+	cfg := &Config{OrPrefixPrimaryToken: "primary"}
+	if got := cfg.AuthHeader(); got != "Token primary" {
+		t.Fatalf("AuthHeader() with primary token = %q", got)
+	}
+
+	cfg = &Config{OrPrefixFallbackToken: "fallback"}
+	if got := cfg.AuthHeader(); got != "Token fallback" {
+		t.Fatalf("AuthHeader() with fallback token = %q", got)
+	}
+}
+`
+	testPath := filepath.Join(outputDir, "internal", "config", "auth_header_prefix_or_test.go")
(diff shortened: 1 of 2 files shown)


tmchow:
@mergifyio queue

PR #2812, opened 2026-06-07
fix(cli): harden force regen merge

### LINKED ISSUE #2674: generator: generate --force merge is two-way — version drift masquerades as hand-edits, emitted-code fixes silently don't land, and a version-var collision breaks the build
## Context

Regenerating a 4.19.0-printed CLI under 4.22.0 with `generate --force` (verified on two CLIs from the 2026-05-28/30 factory batch).

## Bug 1: two-way merge can't tell hand-edits from version drift

The `--force` AST-merge compares the existing file against the **fresh emission only**. Any file that drifted between 4.19.0 and 4.22.0 emission looks 'hand-edited' and is preserved wholesale — so the very fixes a regen is supposed to deliver (e.g. the `rows.Err()` check in the MCP sql tool from 4.22.0, the `syncWarningJSON` JSON-safety fix from 4.21.0) **silently don't land** in those files. `Force regen merged 30 preserved files` reads as success; the user has no signal that the fix-target files kept the old buggy code.

Three-way comparison fixes this precisely: regenerate a reference emission with the *original* printing_press_version (recorded in `.printing-press.json`), then classify each differing file as
- `lib == ref-original` → drift-only, take fresh;
- `lib != ref-original` → real hand-edit, preserve (and report it).

I used exactly this method manually (built the old binary from the version tag, generated a reference, classified) and it cleanly separated real hand-edits (custom auth wiring, novel MCP capability registries, deliberately un-hidden commands) from pure drift on both CLIs.

## Bug 2: version-var collision breaks the merged build

4.22.0 moved `var version` from `root.go` into a new `version.go` (shared with the device generator). The merge preserves old `root.go` (it carries the novel-command `AddCommand` wiring) **and** adds fresh `version.go`:

```
internal/cli/version.go:13:5: version redeclared in this block
	internal/cli/root.go:20:5: other declaration of version
```

Reproduced on **both** CLIs regenerated this way (the post-merge gate output even reports the MCP binary build failure as a warning while exiting 0). The merge needs declaration-level de-duplication: when a fresh file introduces a declaration that already exists in a preserved file, drop one side deterministically.

## Bug 3 (related, cross-file): drift-only replacement can break against preserved files

Fresh `channel_workflow.go` (4.22.0) calls `syncResource(..., maxPages, bool)` — the preserved hand-edited `sync.go` still has the 7-arg signature, so taking the fresh caller breaks the build. A merge that refreshes some files while preserving others needs a compile gate **after** the merge, not before it (the current PASS gates run pre-merge, so the final tree can be broken while the command exits 0).

## Ask

1. Three-way merge using the recorded `printing_press_version` to synthesize the base emission
2. Declaration-level collision handling (the `version` var case)
3. Post-merge compile gate that fails loudly instead of warning + exit 0

tmchow:
## Summary

Force regen can now tell pure template version drift apart from real hand edits when the prior `.printing-press.json` records an older Printing Press version. It synthesizes the original emission for same-spec regens, lets fresh template fixes land when the snapshot matches that base, and still falls back to the existing two-way merge when the base cannot be synthesized.

The merge also prunes fresh declarations that collide with preserved same-package files, covering the `version` var split case without deleting unrelated declarations. When `--validate` is enabled, `generate --force` now runs a final `go build ./...` after the merge so broken mixed trees fail loudly instead of shipping with a warning-only pre-merge gate.

Closes #2674

## Tests

- `go test ./internal/pipeline/regenmerge`
- `go test ./internal/cli`
- `go test ./internal/pipeline -run '^Test(RunLiveDogfoodProcessPreservesLargeJSONUnderCap|HelpScanIndicatesSideEffect)$' -count=1`
- `go test ./...` reached completion but failed once in `internal/pipeline` on `TestRunLiveDogfoodProcessPreservesLargeJSONUnderCap` timeout and `TestHelpScanIndicatesSideEffect`; both passed when rerun directly, and `go test ./internal/pipeline` passed afterward.

---

Codex


tmchow:
@mergifyio queue

PR #2823, opened 2026-06-07
fix(cli): make graphql latest-only choose newest page

### LINKED ISSUE #2354: generator: sync --latest-only assumes newest-first list ordering; fetches the OLDEST page for oldest-first APIs
**Component:** generator (sync templates) · **Priority:** P2 · Surfaced + fixed in dice-fm during live verification.

## Summary
`sync --latest-only` is implemented in both sync templates as "clear the resume cursor and **cap at page 1**" (`internal/generator/templates/sync.go.tmpl:252`, `internal/generator/templates/graphql_sync.go.tmpl:107-110,246`). The help text promises "Refresh head of each resource only." This silently assumes the list endpoint returns **newest-first**. For an endpoint that returns **oldest-first**, page 1 is the OLDEST records — so `--latest-only` fetches the exact opposite of "latest."

## Evidence (dice-fm, live)
DICE's GraphQL connections are oldest-first. Confirmed against the live API:
- `sync --latest-only` (page 1) populated the store with **2021** orders (the oldest), not recent ones.
- The first records the connection returns are from `2021-08-02`.
- A multi-year account therefore "refreshes" to ancient history. `revenue`/`fans`/`returns` analytics computed off that store are biased to the oldest data.

## Root cause
The page-1 cap conflates "first page of the connection" with "newest records." That only holds for newest-first endpoints.

## Fix
- **GraphQL (Relay connections):** page **backward** for `--latest-only` — query `last: N` (optionally with `before:`) instead of `first: N`. Relay connections that support forward pagination support backward pagination; the newest N edges come back. (dice-fm's reference fix does exactly this: a `latest` flag flips the query builder to `last`/`before` and fetches one newest page.)
- **REST:** if the profiler detects a sort/order param, request descending for `--latest-only`; otherwise the flag cannot deliver "latest" by paging from the start — either honor a date-descending param or drop/redefine the flag rather than shipping a misnamed one.

Either way the default should not promise "latest" while returning oldest.

## Related (also seen in dice-fm, worth considering together)
The default `sync` (oldest-first + a `--max-pages` cap) front-loads ancient data and never reaches recent activity within the cap, so store-backed analytics are systematically incomplete on long-lived accounts. dice-fm's reference fix defaults sync to uncapped (`--max-pages 0`) with `--since <window>` as the bounded escape hatch. The generator default + the oldest-first interaction is the same root concern as this flag.

## Reference fix
dice-fm printed-CLI fix on the library PR (`feat(dice-fm): complete-by-default sync & reads, newest-first --latest-only, show-date filters`) — backward pagination for `--latest-only`, uncapped sync default, loud truncation warnings. The generator templates should adopt the backward-pagination pattern for the GraphQL path at minimum.

## Acceptance
- positive: an oldest-first list endpoint synced with `--latest-only` returns the NEWEST page (assert by timestamp), not page 1.
- negative: a newest-first endpoint still returns its newest records (unchanged behavior).

tmchow:
## Summary
Generated GraphQL CLIs no longer assume the first Relay page is the latest page for `sync --latest-only`. When a list field exposes `last`, sync now probes the backward page and uses timestamp evidence to choose the genuinely newest page, while timestamp-less or already-newest-first connections keep the existing forward-page behavior.

Closes #2354

## Tests
- `go test ./internal/generator -run '^TestGeneratedGraphQLLatestOnlyChoosesNewestPage$' -count=1`
- `go test ./internal/generator -count=1`
- `scripts/golden.sh verify`
- `go test ./...`
- `golangci-lint run ./...`

### DIFF
--- a/internal/generator/generator.go
+++ b/internal/generator/generator.go
@@ -394,7 +394,9 @@ func New(s *spec.APISpec, outputDir string) *Generator {
 		"publicFlagAliases":            publicFlagAliases,
 		"flagChangedExpr":              flagChangedExpr,
 		"graphqlListParams":            graphqlListParams,
+		"graphqlLatestParams":          graphqlLatestParams,
 		"graphqlVariableType":          graphqlVariableType,
+		"hasGraphQLParam":              hasGraphQLParam,
 		"mcpInputName":                 mcpInputName,
 		"mcpToolInputParams":           mcpToolInputParams,
 		"mcpParamBindings":             mcpParamBindings,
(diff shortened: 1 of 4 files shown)


tmchow:
@mergifyio queue

PR #2826, opened 2026-06-07
fix(cli): derive live dogfood fixtures from store

### LINKED ISSUE #2431: [scorer] Dogfood matrix should derive parent-ID fixtures from the synced local store
**Component:** `scorer` (`dogfood --live` matrix builder)
**Priority:** P2

## Summary

Dogfood `--live` mechanically tests each endpoint command's `happy_path` and `json_fidelity` paths. For endpoints with required parent IDs (`workspace_id`, `project_id`, etc.), the matrix passes synthetic fixtures from the OpenAPI `examples` block — often placeholder strings like `"example-id"` or `"0"`.

Todoist returns HTTP 400 `Invalid argument value` for these placeholders. 9 of the 14 remaining dogfood failures in the Todoist run were this class of failure. None are real CLI bugs.

## Why this is a P2

- Affects every CLI with parent-keyed endpoints: Linear (team_id), Notion (database_id), Stripe (customer), Cal.com (event_type_id), GitHub (owner/repo), etc. Universal pattern.
- The matrix already runs its own sync step for some commands — extending that to use synced data for fixtures is well-bounded.

## Acceptance criteria

**Positive:** Matrix against a CLI where sync populated tasks resolves the `tasks get-task` happy_path with a real task ID (extracted from the local store) and the test passes.

**Negative:** For resources where the store is empty after sync (or has no qualifying row), the matrix falls back to synthetic placeholders and the failure detail includes `"fixture_source": "synthetic"` so retros can correctly classify it.

## Likely fix direction

After the matrix's pre-sync step:
1. Open the local store.
2. For each endpoint with a path-or-query placeholder of shape `{resource}_id` or `{resource_id}`, query `SELECT id FROM <resource> LIMIT 1` (or `SELECT id FROM resources WHERE resource_type='<resource>' LIMIT 1` — depending on the typed-table fix in #related-area).
3. Substitute the resolved ID into the example invocation.
4. If no row exists, mark `fixture_source: synthetic` and use the OpenAPI example value as today.

## Test

Run `dogfood --live` against `todoist-pp-cli` with a populated store. `tasks get-task` happy_path passes (real task ID resolved). For workspaces (empty in this user's case), the failure detail explicitly shows `fixture_source: synthetic` so the retro can distinguish real bugs from fixture issues.

## Artifacts
- Retro doc: https://files.catbox.moe/nzaruw.md
- Manuscripts (research, proofs): https://files.catbox.moe/38mw3c.zip
- CLI source (no binary, no vendor/): https://files.catbox.moe/ly0cxx.zip
- Source CLI: `todoist-pp-cli` (Todoist API v1 run `20260527-151452`)

tmchow:
## Summary

Live dogfood now primes the printed CLI store with a quiet `sync` pass when the CLI exposes one, then uses the scoped local SQLite cache to replace ID-shaped example placeholders when companion list calls cannot provide a real ID.

If the store exists but has no matching row, the matrix keeps the synthetic example and marks the row with `fixture_source: synthetic`, giving retros a reliable way to classify fixture-driven failures.

Closes #2431

## Tests

- `go test ./internal/pipeline -run 'TestRunLiveDogfood(StoreFixture|ResolveSuccess|ResolveCommandPositionalsSkipPaths)'`
- `go test ./internal/pipeline -run 'TestRunLiveDogfoodStoreFixture'`
- `go test ./...`

---

Codex

### DIFF
--- a/internal/pipeline/live_dogfood.go
+++ b/internal/pipeline/live_dogfood.go
@@ -580,10 +580,18 @@ func runLiveDogfoodPreSync(commands []liveDogfoodCommand, ctx resolveCtx) {
 	}
 	for _, command := range commands {
 		if len(command.Path) == 1 && command.Path[0] == "sync" {
-			_ = runLiveDogfoodProcess(ctx.binaryPath, ctx.cliDir, []string{"sync"}, ctx.timeout)
+			_ = runLiveDogfoodProcess(ctx.binaryPath, ctx.cliDir, []string{"sync"}, liveDogfoodPreSyncTimeout(ctx.timeout))
 			return
 		}
 	}
+}
+
+func liveDogfoodPreSyncTimeout(timeout time.Duration) time.Duration {
+	const maxPreSyncTimeout = 5 * time.Second
+	if timeout <= 0 || timeout > maxPreSyncTimeout {
+		return maxPreSyncTimeout
+	}
+	return timeout
 }
 
 // buildSiblingMap groups commands by their joined parent path so the chain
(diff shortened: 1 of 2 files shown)


tmchow:
@mergifyio queue

PR #2828, opened 2026-06-07
fix(cli): dedupe multi-spec duplicate endpoint commands

### LINKED ISSUE #2455: generator: de-duplicate command registration when two specs register the same pp:endpoint
**From retro:** [`ga4-goat` retro 2026-05-28](https://github.com/mvanhorn/cli-printing-press/blob/main) (local: `~/printing-press/manuscripts/ga4-goat/20260528-172450/proofs/20260528-180038-retro-ga4-goat-pp-cli.md`)

## Goal
During multi-spec merge, when both `<resource>` and `<spec-title-slug>-<resource>` would be registered for the same `pp:endpoint`, keep only the un-prefixed name and discard the title-slug alias. Provide `--name-prefix` opt-in for users who want the namespaced form.

## Problem
Generation with `--spec a --spec b --spec c` produces duplicate command paths at the top level. Observed during a 3-spec merge for GA4:

- `ga4-goat-pp-cli accounts list` AND `ga4-goat-pp-cli google-analytics-admin-accounts list` both exist and both resolve to `GET https://analyticsadmin.googleapis.com/v1beta/accounts`.
- Same pattern for `accounts search-change-history-events`, `account-summaries`, several others.

The duplicate registration burns MCP surface and `--help` real estate for no semantic gain. The user can't tell which form is the "real" command.

## Cross-API evidence
Recurs whenever a user passes 2+ specs that share resource names (common shape: vendor v1beta + v1alpha of the same API). Only one concrete catalog example with direct evidence (this `ga4-goat` run, three GA4 specs); per the Step B rule, this is P3.

## Durable fix
During multi-spec resource merge, when both `<resource>` and `<spec-title-slug>-<resource>` would be registered for the **same** `pp:endpoint` annotation (same method + same path), keep only the un-prefixed form and discard the title-slug alias. Both forms remain when `pp:endpoint` differs (genuinely distinct resources that happened to share a name).

## Acceptance criteria
- **positive:** generating from two specs containing identical `accounts` resources (same method + same path) emits one top-level `accounts list` command, not two.
- **negative:** generating from two specs containing distinct `accounts` resources with different `pp:endpoint` still emits both, under their namespaced prefixes (current behavior preserved).

## Scope boundary
Does NOT change single-spec generation. Does NOT auto-resolve cross-version method-shape changes (different request bodies for same path) — those still need two distinct registrations.

## Complexity
Small (generator emitter walks resource trees, compares `pp:endpoint`, dedups).

## Adversarial check (Phase 3 Step G)
Case-against: "Multi-spec is rare; this is a P3 corner case." Yes — and the doc is honest about that (subclass:multi-spec-runs). Case-for: the fix is small (no risk of hurting single-spec runs) and the duplicate cost compounds quickly when each spec adds N resources.

## Related issues
- #1208 — `mcp-sync --force` creates spurious `cmd/<spec-title-slug>-pp-{cli,mcp}/` when `info.title` slug differs from CLI slug. Adjacent area (multi-spec debris), different specific bug.

## Evidence
`ga4-goat-pp-cli --help` from this session shows `account-summaries` AND `google-analytics-admin-account-summaries` as siblings; `accounts list` AND `google-analytics-admin-accounts list` also both registered.

tmchow:
## Summary

Multi-spec generation now drops redundant spec-prefixed resource aliases when another resource already registers the same endpoint signature. A duplicate `accounts list` endpoint now emits one command by default, while genuinely distinct same-name resources still keep the namespaced fallback.

Users who want fully namespaced multi-spec resource commands can pass `--name-prefix` to prefix resource names with the source spec name.

Closes #2455

## Approach

The merge step now compares endpoint signatures during resource-name collisions before creating a `<spec-name>-<resource>` fallback. If the incoming resource is already covered by the existing resource's method, effective base URL, and path signatures, it is skipped. If any endpoint differs, the previous prefixed fallback remains.

## Scope

Primary area: multi-spec generation in the cli-printing-press binary.

Why this belongs in this repo: this is Printing Press generator behavior that affects future printed CLIs, not a one-off printed CLI fix.

Catalog Justification: N/A

## Risk

The main risk is changing command names for multi-spec duplicate endpoint collisions. The tests pin the default dedupe behavior, the distinct-endpoint fallback, the `--name-prefix` opt-in, and generated Cobra output for the duplicate case.

## Output Contract

Generated-output proof is covered by `TestGenerateMultiSpecDeduplicatesSameEndpointResourceCollision`, which runs generation and asserts the emitted root command registers the unprefixed command without the duplicate spec-prefixed command.

## Verification

- `go fmt ./...`
- `go test ./internal/cli -run 'Test(GenerateMultiSpecDeduplicatesSameEndpointResourceCollision|MergeSpecs(DeduplicatesSameEndpointResourceCollision|PrefixesDistinctEndpointResourceCollision|NamePrefixOptInKeepsNamespacedResourceForm|PreservesPerSpecBaseURLPrefixes|CarriesMCPConfigFromFirstDeclaringSpec|RewritesMCPIntentRefsForRenamedResources))$'`
- `go test ./...`
- `scripts/golden.sh verify`

### DIFF
--- a/internal/cli/generate_test.go
+++ b/internal/cli/generate_test.go
@@ -1370,6 +1370,53 @@ func TestMergeSpecsPrefixesDistinctEndpointResourceCollision(t *testing.T) {
 	assert.Equal(t, "/v1alpha/accounts", merged.Resources["analytics-admin-accounts"].Endpoints["list"].Path)
 }
 
+func TestMergeSpecsDeduplicatesTrailingSlashEndpointResourceCollision(t *testing.T) {
+	t.Parallel()
+
+	specA := &spec.APISpec{
+		Name:    "admin",
+		Version: "0.1.0",
+		BaseURL: "https://analytics.example.com",
+		Resources: map[string]spec.Resource{
+			"accounts": {
+				Endpoints: map[string]spec.Endpoint{
+					"list": {Method: "GET", Path: "/v1beta/accounts"},
+				},
+			},
+			"reports": {
+				Endpoints: map[string]spec.Endpoint{
+					"list": {Method: "GET", Path: "/v1beta/reports"},
+				},
+			},
+		},
+		Types: map[string]spec.TypeDef{},
+	}
+	specB := &spec.APISpec{
+		Name:    "analytics-admin",
+		Version: "0.1.0",
+		BaseURL: "https://analytics.example.com",
+		Resources: map[string]spec.Resource{
+			"accounts": {
+				Endpoints: map[string]spec.Endpoint{
+					"list": {Method: "GET", Path: "/v1beta/accounts/"},
+				},
+			},
+			"metadata": {
+				Endpoints: map[string]spec.Endpoint{
+					"list": {Method: "GET", Path: "/v1beta/metadata"},
+				},
+			},
+		},
+		Types: map[string]spec.TypeDef{},
+	}
+
+	merged := mergeSpecs([]*spec.APISpec{specA, specB}, "ga4")
+
+	assert.Contains(t, merged.Resources, "accounts")
+	assert.NotContains(t, merged.Resources, "analytics-admin-accounts")
+	assert.Equal(t, "/v1beta/accounts", merged.Resources["accounts"].Endpoints["list"].Path)
+}
+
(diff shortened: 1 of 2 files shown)


tmchow:
@mergifyio queue

PR #2830, opened 2026-06-07
fix(cli): skip idless resources in default sync

### LINKED ISSUE #2461: generator: emit empty defaultSyncResources() when no spec endpoint returns an array
**Source:** Retro from `apple-docs` print on 2026-05-28 ([PR #920](https://github.com/mvanhorn/printing-press-library/pull/920)).

## What happened

The generator emits `defaultSyncResources() []string` returning all top-level resources. The sync command then iterates each, calling its `list` endpoint and expecting items with extractable IDs. For document-addressable APIs (every endpoint returns a single object, not a paginated list), this fails with:

```
{"event":"sync_start","resource":"technologies"}
{"event":"sync_error","resource":"technologies","error":"missing id for technologies"}
{"event":"sync_summary","resources":1,"success":0,"errored":1}
Error: 1 resource(s) failed to sync
```

This exits 1 and breaks the Phase 5 dogfood `sync [happy_path]` test. The workaround is to hand-patch `defaultSyncResources()` to return `nil` so sync becomes a no-op exit 0.

## Cross-API recurrence

Affects content-addressable APIs where every endpoint returns a single object instead of a list. **Currently named:** apple-docs (DocC JSON) is the only on-disk concrete example I can point to in this user's library. Filed at P3 because the named-3-APIs Step B evidence is currently thin; promote to P2 when 2+ more APIs hit this shape.

Likely candidates (not verified in catalog): MDN docs, Apple Sample Code archives, Stack Exchange API single-question fetches, any RSS-feed-based reader, GitHub raw content URLs.

## Durable fix

During spec profiling, mark a resource as "syncable" only if it has an endpoint with `response.type: array` and a clear pagination or list shape. `defaultSyncResources()` returns only resources flagged syncable. When the set is empty, the generated help block should also explain why:

```
this CLI's API doesn't have list-shaped endpoints; use per-command commands for fetching
```

## Test

- **Positive:** Generate from a spec where every endpoint has `response.type: object` (not `array`). Assert emitted `defaultSyncResources()` returns `nil` and `<cli> sync` (no args) exits 0 with the no-list-endpoints message.
- **Negative:** Generate from a normal list-shaped spec (one or more `response.type: array`). Assert `defaultSyncResources()` returns the list-shaped resource names — unchanged from current behavior.
- **Negative:** Don't suppress explicit `sync --resources <name>` — that should still run (and surface the current "missing id" error if invoked on a non-array resource). The change is only to the empty-args default.

## Scope boundary

Default sync set + sync help message only. Don't alter the per-resource syncer.

## Artifacts

Retro doc with the full triage rationale (including Skip + Drop tables): `/tmp/printing-press/retro/20260529-073000-apple-docs-retro.md`

tmchow:
## Summary

Empty-args `sync` now skips typed list resources whose response item shape cannot yield a runtime primary key. Those resources remain valid explicit sync targets through `--resources`, so maintainers can still force the current per-resource behavior while generated defaults avoid the known `missing id for <resource>` failure mode.

The empty-default warning now describes the broader condition: no default sync resources are available. This covers both specs with no bulk list endpoints and specs where every known sync target is intentionally excluded from `sync all`.

Closes #2461

## Verification

- `go test ./internal/profiler -run 'TestProfile(SkipsTypedIDlessListsFromDefaultSync|ExcludesScalarArrayAndSamplerEndpoints|WrapperObjectDetection)'`
- `go test ./internal/generator -run 'TestGenerateSync(DefaultsSkipTypedIDlessResources|EmitsEmptyHintWhenNoBulkList|SkipsEmptyHintWhenBulkListExists|DefaultsSkipAuthTaggedResources)'`
- `scripts/golden.sh verify`
- `go test ./internal/pressauth -run TestCaptureTimeoutCleansTempDir -count=1`
- `go test ./...`

Note: the first `go test ./...` run failed only in `internal/pressauth` on `TestCaptureTimeoutCleansTempDir` reporting a leaked tempdir. The targeted rerun passed, and the second full `go test ./...` run passed.

### DIFF
--- a/internal/generator/generator.go
+++ b/internal/generator/generator.go
@@ -297,6 +297,7 @@ func New(s *spec.APISpec, outputDir string) *Generator {
 		"composeMCPDesc":                     composeMCPDesc,
 		"composeMCPSubDesc":                  composeMCPSubDesc,
 		"mcpParamDesc":                       g.mcpParamDescription,
+		"hasDefaultSyncResources":            hasDefaultSyncResources,
 		"flagName":                           flagName,
 		"paramIdent":                         paramIdent,
 		"paramWireName":                      paramWireName,
@@ -3257,6 +3258,15 @@ func paginationSupportedResources(syncable []profiler.SyncableResource, dependen
 	}
 	sort.Strings(names)
 	return names
+}
+
+func hasDefaultSyncResources(syncable []profiler.SyncableResource) bool {
+	for _, resource := range syncable {
+		if !resource.SkipDefaultSync {
+			return true
+		}
+	}
+	return false
 }
 
 func specDateTimeFieldNames(api *spec.APISpec) []string {
--- a/internal/generator/sync_param_passthrough_test.go
+++ b/internal/generator/sync_param_passthrough_test.go
@@ -215,8 +215,8 @@ func TestGenerateSyncDefaultsSkipTypedIDlessResources(t *testing.T) {
 		"explicit --resources should still accept the idless list endpoint")
 	assert.Contains(t, syncSrc, "no default sync resources",
 		"sync should explain why empty-args sync is a no-op")
-	assert.Contains(t, syncSrc, `"reason":"no_default_sync_resources"`,
-		"JSON sync warnings should classify the empty default set")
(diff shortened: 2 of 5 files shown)


tmchow:
@mergifyio queue

PR #2835, opened 2026-06-07
fix(cli): synthesize html page sync ids

### LINKED ISSUE #2444: generator: upsertSingleObject should synthesize id fallback for HTML page-mode resources
> Filed from a Printing Press retro. Full retro doc: `/tmp/printing-press/retro/20260528-152300-revenuecat-retro.md` (preserved in `manuscripts/revenuecat/20260528-143019/proofs/` locally; not uploaded — no PII/secrets in artifacts).

## What

`sync` upserts of HTML page-mode resources crash with `missing id for <resource>` when the page-mode extractor returns a record with no `id` / `slug` / `name` field. The page-mode shape commonly emitted is `{title, description, canonical_url, image_url, links}` — none of those are detected by `extractID` in the generated `internal/cli/sync.go`'s `upsertSingleObject`.

This was caught by `cli-printing-press verify`'s data-pipeline gate (FAIL: `sync crashed`) and hand-fixed by the polish skill, which patched the printed CLI's `sync.go` to synthesize an id from `canonical_url` → `url` → `image_url` → resource name when `extractID` returns empty. Polish's fix is per-CLI; the next HTML page-mode CLI's `sync.go` ships with the same template-level bug and needs polish to catch it again.

## Reproducer

Generate a CLI for a spec with `response_format: html` + `mode: page` resources whose page-mode extractor doesn't surface a slug/id-shaped field. `revenuecat-pp-cli`'s `glossary` and `platform` resources both hit this. Sync against the live API and observe `missing id for glossary` / `missing id for platform` warnings; verify reports `Data Pipeline: FAIL: sync crashed`.

## Cross-API evidence

- **revenuecat-pp-cli** (this run): `glossary` and `platform` resources — polish-patched.
- **food52-pp-cli** (in library): same `response_format: html` + `mode: page` extraction shape on recipe pages.
- **Any future website-itself CLI** via the `/printing-press <URL>` "the website itself" flow.

## Proposed durable fix

In the generator template for `upsertSingleObject` (`internal/generator/`), when `extractID` returns empty AND the resource declares `response_format: html` + `mode: page`, synthesize id in this fallback order: `canonical_url` → `url` → `image_url` → resource name. Inject into the JSON payload before dispatching to the typed `UpsertXxx`.

The fallback only fires when `extractID` returns empty, so explicit slug fields keep winning. The condition is gated on `response_format: html`/`mode: page` so JSON-API resources are untouched.

## Acceptance criteria

- **Positive:** Generate a CLI for a spec with HTML page-mode resources whose extraction has no slug field. Run `sync`; assert no `missing id for X` warnings, all rows land with synthesized ids matching the fallback chain.
- **Negative:** Generate a CLI for a spec with HTML page-mode resources that DO have explicit slug fields. Assert the explicit slug wins (no regression on the well-formed case).
- **Negative:** Generate a JSON-API CLI. Assert `upsertSingleObject` shape is unchanged.

## Scope boundary

Do NOT modify the HTML extractor itself. Its page-mode output shape is fine; the upsert path absorbs the id-synthesis responsibility.

## Complexity

Small — generator template edit + targeted regression test.

## Related issues

None at retro time (no prior dedup matches in `gh issue list --label retro --state open`).

---
*Retro session: `revenuecat`, run `20260528-143019`. Final verdict: `hold` (Phase 5 live dogfood blocked by separate sitemap-sync gap, which dropped at retro Step B due to insufficient cross-API evidence).*

tmchow:
## Summary
- add HTML page-mode sync ID synthesis for single-object resources using canonical_url, url, image_url, then the resource name
- inject synthesized IDs into typed upsert payloads while preserving spec/profile ID overrides such as x-resource-id
- add generated-runtime regression coverage for HTML page metadata, URL fallbacks, resource fallback, JSON-only output shape, and spec-declared ID precedence

## Tests
- go test ./internal/generator -run '^(TestGeneratedHTMLPageModeSingleObjectIDSynthesis|TestUpsertDispatchPreservesMultiWordResourceCasing)$' -count=1
- go test ./internal/openapi -run '^TestGenerateDataEnvelopeAllOfBodyFlags$' -count=1
- scripts/golden.sh verify
- scripts/verify-generator-output.sh
- go test ./...
- go build -o ./cli-printing-press ./cmd/cli-printing-press
- golangci-lint run ./...
- git diff --check

Closes #2444

### DIFF
--- a/internal/generator/generator.go
+++ b/internal/generator/generator.go
@@ -3133,6 +3133,7 @@ type visionRenderData struct {
 	GraphQLFieldPaths            map[string]string
 	AgentMoneyWorkflow           AgentMoneyWorkflow
 	HTMLSyncStub                 bool
+	HTMLPageModeResources        []criticalResourceEntry
 }
 
 type resourceIDFieldOverrideEntry struct {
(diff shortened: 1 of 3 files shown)


tmchow:
@mergifyio queue

PR #2866, opened 2026-06-08
fix(cli): allow synthetic phase5 external credential skips

### LINKED ISSUE #2479: phase5 promote-gate has no skip path for synthetic/SDK CLIs needing out-of-band credentials
## Finding (P3 · comp:scorer · subclass:cloud-sdk-auth)

**Cloud/SDK credential-chain APIs have no auth mode → forced `auth: none` → the phase5 promote-gate has no valid skip path when credentials are unavailable in-session.**

### What happened
A CLI consumed through a vendor SDK (here: AWS Cost Explorer via AWS SDK for Go v2, SigV4 + the native credential chain — env / shared profile / SSO / assume-role / IMDS) has no auth mode that fits. The mode list (api_key, bearer_token, cookie, composed, oauth, session_handshake, none) has no entry for "credentials resolved out-of-band by an SDK," so the only workable choice is `auth: none`.

But `auth: none` signals to the rest of the machine "freely testable, no creds needed" — which is false. Concretely it breaks the **phase5 promote gate**:

```
lock promote ... → phase5 gate failed: phase5 skip marker auth type "oauth2" does not match manifest auth type "none"
```

`phase5-skip.json` is rejected for an `auth:none` manifest (no-auth APIs are expected to produce a passing acceptance, not skip). So a `kind: synthetic` / SDK CLI that genuinely **cannot be live-tested without external credentials the automated agent can't obtain** (interactive `aws sso login`, MFA, no creds on CI) has **no valid path to promote** — it's a dead end unless the agent happens to get live creds or hand-edits the gate.

### Scorer correct?
Partially. The gate is right that `auth: none` usually means testable; it's wrong that `auth: none` can never legitimately require out-of-band credentials.

### Cross-API check / frequency
First sighting in this ecosystem (1 catalog example today: AWS). The class is obvious and recurring — every cloud-provider CLI built on a vendor SDK lands here identically:
- **AWS** — all services via SDK + SigV4 (this run)
- **GCP** — Cloud Billing / google-cloud SDKs, ADC credential chain
- **Azure** — Cost Management SDK, DefaultAzureCredential

`subclass:cloud-sdk-auth` — rare now, inevitable as cloud CLIs get printed. Filed P3 on thin current evidence (1 API) + high future certainty, per the retro's "first sighting of a behavior you'd encounter again" clause.

### Fallback if unfixed
Agent must obtain live creds in-session to write acceptance (often impossible), OR hand-edit the gate, OR leave the CLI stranded unpromoted. All three are bad.

---

## WU-1: phase5 promote-gate skip path for synthetic / external-credential CLIs

- **Priority:** P3
- **Component:** scorer (phase5 acceptance/skip gate in promote + dogfood acceptance tooling)
- **Goal:** A `kind: synthetic` (or a new `external`/`sdk` auth mode) CLI that cannot be live-tested without out-of-band credentials can write a valid `phase5-skip.json` and promote, instead of being rejected on auth-type mismatch.
- **Acceptance criteria:**
  - **positive:** a `kind: synthetic` CLI with `phase5-skip.json {skip_reason: external_credentials_unavailable}` promotes.
  - **negative:** a standard `auth: none` HTTP CLI still requires a passing `phase5-acceptance.json` (skip rejected) — no regression.
- **Scope boundary:** The minimal fix is the gate's skip-acceptance for `kind: synthetic`; it does NOT require building the full `external`/`sdk` auth mode (that's a larger follow-on so doctor/README/auth surfaces describe credential-chain auth honestly).
- **Complexity:** small (gate) / medium (if the new auth mode is included)

### Durable fix
Smallest increment: allow `phase5-skip.json` with `skip_reason: external_credentials_unavailable` to satisfy the promote gate when the manifest is `kind: synthetic` (already a signal), instead of rejecting on auth-type mismatch. Larger: introduce an `external`/`sdk` auth mode end-to-end.

### Evidence
Generated `aws-billing-pp-cli` (synthetic, SDK transport). shipcheck 6/6 PASS, scorecard 80/A, verify 100%. The promote dead-end reproduced until aws-vault creds were located out-of-band; live dogfood then wrote a real acceptance and promote succeeded — but the general no-creds case has no skip path.

### Related issues
None (no matching open retro issue).

---
*Filed via /printing-press-retro. Artifacts withheld — manuscripts contain real AWS cost figures.*

tmchow:
## Summary
Synthetic SDK-backed CLIs that rely on out-of-band credential chains can now promote when live credentials are unavailable. A generated CLI manifest records `spec_kind: synthetic`, and the phase5 promote gate accepts `phase5-skip.json` with `skip_reason: external_credentials_unavailable` only for that synthetic no-auth case.

Standard `auth: none` HTTP CLIs still require a real phase5 pass marker, and the promote path now carries manifest spec provenance into the gate before validating skips.

Closes #2479

## Tests
- `go test ./internal/pipeline -run 'TestValidatePhase5Gate_(SyntheticNoAuthAllowsExternalCredentialSkip|StandardNoAuthRejectsExternalCredentialSkip|NoAuthRequiresPassMarker|SkipCannotOverrideManifestAuthType)$'`
- `go test ./internal/pipeline -run 'TestWriteManifestForGenerateStampsSyntheticSpecKind|TestPromoteWorkingCLI_AllowsSyntheticExternalCredentialPhase5Skip'`
- `go test ./internal/pipeline -run 'TestValidatePhase5Gate|TestWriteManifestForGenerateStampsSyntheticSpecKind|TestPromoteWorkingCLI_AllowsSyntheticExternalCredentialPhase5Skip' -count=1`
- `scripts/golden.sh verify`
- pre-push hook: `fmt` and `golangci-lint` passed

`go test ./...` currently fails in `internal/pipeline` at `TestResolveCommandPositionalsMixedStoreAndCompanionSourceIsUntagged`; rerunning that test alone reproduces the same positional live-dogfood failure, outside the phase5/manifest paths changed here.

### DIFF
--- a/internal/pipeline/climanifest.go
+++ b/internal/pipeline/climanifest.go
@@ -114,6 +114,7 @@ type CLIManifest struct {
 	SpecURL            string            `json:"spec_url,omitempty"`
 	SpecPath           string            `json:"spec_path,omitempty"`
 	SpecFormat         string            `json:"spec_format,omitempty"`
+	SpecKind           string            `json:"spec_kind,omitempty"`
 	SpecSource         string            `json:"spec_source,omitempty"`
 	SpecChecksum       string            `json:"spec_checksum,omitempty"`
 	RunID              string            `json:"run_id,omitempty"`
@@ -188,6 +189,12 @@ func (m CLIManifest) IsLocalDatastore() bool {
 		return true
 	}
 	return strings.Contains(source, "local") && strings.Contains(source, "sqlite")
+}
+
+// IsSyntheticSpec reports whether the manifest came from a spec marked
+// `kind: synthetic`, which relaxes gates that assume HTTP API reachability.
+func (m CLIManifest) IsSyntheticSpec() bool {
+	return strings.EqualFold(strings.TrimSpace(m.SpecKind), spec.KindSynthetic)
 }
 
 // NovelFeatureManifest is a compact representation of a transcendence feature
@@ -669,6 +676,7 @@ func populateMCPMetadata(m *CLIManifest, parsed *spec.APISpec) {
 	if parsed == nil {
 		return
 	}
+	m.SpecKind = parsed.Kind
 	total, public := parsed.CountMCPTools()
 	mcpName := m.APIName
 	if mcpName == "" {
--- a/internal/pipeline/climanifest_test.go
+++ b/internal/pipeline/climanifest_test.go
@@ -654,6 +654,7 @@ resources:
 	assert.Equal(t, 1, got.MCPPublicToolCount)
 	assert.Equal(t, "full", got.MCPReady)
 	assert.Equal(t, "bearer_token", got.AuthType)
+	assert.Empty(t, got.SpecKind)
 	assert.Equal(t, []string{"INTERNAL_SPEC_TOKEN"}, got.AuthEnvVars)
 }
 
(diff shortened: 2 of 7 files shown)


tmchow:
@mergifyio queue

PR #2876, opened 2026-06-08
fix(cli): preserve Windows companion CLI fallback

### LINKED ISSUE #2503: generator: Windows companion-CLI sibling lookup misses .exe -> sync/sql/search unavailable
**Component:** generator (cobratree `cli_path.go` template) · **Priority:** P2
**Source:** FreshBooks retro (live debug+fix session; session + CLI source evidence, no artifact upload)

## What happened
On Windows, MCP tools that shell out to the companion CLI (`sync`/`import`/`sql`/`search`) failed with "companion CLI binary not found" even though `<cli>-pp-cli.exe` sat directly beside `<cli>-pp-mcp.exe`. `SiblingCLIPath` builds the sibling candidate as `filepath.Join(dir, cliName)` with **no extension**; `os.Stat` of an extensionless name fails on Windows, so it falls through to `FRESHBOOKS_CLI_PATH` / `PATH` (both typically unset) and errors with a message that misleadingly points at PATH/env rather than the real `.exe` miss.

## Cross-API blast radius
Affects **100% of generated CLIs that ship shell-out local-store tools** (the sync/import/sql/search family the press adds for APIs with bulk-list endpoints) **when run on Windows**. Not API-shape-dependent — it's OS-gated and template-wide. The press declares win32 support (`manifest.json` `platforms` includes `win32`) and emits Windows binaries, so the platform is in scope. Fallback reliability is low: manifests only on Windows at runtime; a whole feature class is silently unavailable.

## Durable fix
In `SiblingCLIPath`, when `runtime.GOOS == "windows"`, try `cliName + ".exe"` before the extensionless name; keep the extensionless candidate as a fallback (the `.mcpb` bundle ships the sibling without an extension) and for unix. Loop candidates; first `os.Stat` hit wins.

```go
candidates := []string{filepath.Join(dir, cliName)}
if runtime.GOOS == "windows" {
    candidates = []string{filepath.Join(dir, cliName+".exe"), filepath.Join(dir, cliName)}
}
for _, c := range candidates { if _, err := os.Stat(c); err == nil { return c, nil } }
```

## Acceptance
- positive: on `GOOS=windows`, a sibling `<cli>-pp-cli.exe` resolves with no `*_CLI_PATH` env set.
- negative: on linux/darwin the extensionless sibling still resolves (no regression); on Windows the extensionless bundle name still resolves as a fallback.

## Evidence
Reproduced: deployed config points at the library dir where the sibling is `*.exe`; lookup missed it → "companion CLI binary not found". After adding the `.exe` candidate, the rebuilt MCP ran the sibling (returned a CLI flag-parse error, proving the binary was found and executed).

_Filed via /printing-press-retro. No duplicate found among open `retro` issues._

tmchow:
## Summary

Windows generated MCP servers now check both sibling companion CLI names when resolving shell-out tools: the target-platform `.exe` name first, then the extensionless fallback. That keeps Windows installs working when the companion binary is staged as `<cli>-pp-cli.exe` while preserving the existing extensionless bundle fallback and non-Windows behavior.

The generated-output test now asserts the Windows candidate order and the Linux single-candidate path, and the golden fixture reflects the emitted resolver contract.

Closes #2503

## Tests

- `go fmt ./...`
- `go test ./internal/generator -run '^TestMCPCobraTreeSiblingCLIPathUsesWindowsExecutableSuffix$' -count=1`
- `go test ./...`
- `scripts/golden.sh verify`
- `scripts/verify-generator-output.sh generate-golden-api`
- `golangci-lint run ./...`

### DIFF
--- a/internal/generator/mcp_novel_features_test.go
+++ b/internal/generator/mcp_novel_features_test.go
@@ -236,7 +236,10 @@ func TestMCPCobraTreeSiblingCLIPathUsesWindowsExecutableSuffix(t *testing.T) {
 	var testSrc strings.Builder
 	testSrc.WriteString(`package cobratree
 
-import "testing"
+import (
+	"path/filepath"
+	"testing"
+)
 
 func TestCLIExecutableNameUsesWindowsSuffix(t *testing.T) {
 	if got := cliExecutableName("windows"); got != "pathcheck-pp-cli.exe" {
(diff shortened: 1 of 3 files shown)


tmchow:
@mergifyio queue

PR #2969, opened 2026-06-14
fix(generator): mark tail command no-error-path-probe

### LINKED ISSUE #2968: generator: tail command guarantees a dogfood error_path failure (missing pp:no-error-path-probe)
## Summary
The generated `tail` command guarantees a dogfood `error_path` failure on every printed CLI that has a data layer. `tail` is a resilient poller: on a failed initial fetch it prints `warning: initial fetch failed: ...` and `return nil` (exit 0). The dogfood `error_path` probe runs `tail <invalid-resource>` and expects a non-zero exit, so it scores FAIL unless the command is hand-annotated with `pp:no-error-path-probe`.

## Component
`generator` — the tail command template that emits `internal/cli/tail.go`.

## Evidence
- **stripe** (run `20260614-143744-aa749241`): Phase 5 live dogfood `error_path` probe `tail __printing_press_invalid__` → exit 0 with `warning: initial fetch failed`; scored FAIL. Cleared only by hand-adding `"pp:no-error-path-probe": "true"` to the command Annotations.
- **twilio** (`library/twilio/internal/cli/tail.go`): byte-identical template — line 41 `warning: initial fetch failed`, line 44 `return nil`, and **zero** `pp:no-error-path-probe` annotations. Would fail the same probe.

This is deterministic template output (tail.go is emitted for every CLI with a syncable data layer), so it is structural, not sampled — the two local CLIs both confirm the identical pattern.

## Why it matters
Every tail-bearing CLI either eats a dogfood `error_path` failure that blocks the Phase 5 Full gate, or requires a manual annotation each run. A future agent will forget the annotation, so runs intermittently hold on a non-defect.

## Scorer assessment
Partially correct. The `error_path` probe is right to expect non-zero for invalid arguments on most commands, but `tail` is a poller for which that contract does not apply. Fix the generator, not the probe.

## Proposed fix
**Recommended (safe, no behavior change):** In the generator tail template, add `"pp:no-error-path-probe": "true"` to the command Annotations. Tells dogfood to skip the inapplicable `error_path` probe.

**Alternative (more correct, riskier):** Make `tail` return a non-zero error when the *initial* fetch fails (before entering the poll loop) — a tail given a bad resource should error rather than silently poll nothing. Changes runtime behavior; maintainer's call.

## Acceptance criteria
- Positive: generate any CLI with a data layer → `internal/cli/tail.go` Annotations include `pp:no-error-path-probe` (or tail exits non-zero on initial fetch failure); `dogfood --live` Full gate passes with no hand-edit.
- Negative: a normal generated command with an invalid argument still gets the `error_path` probe and must exit non-zero (the annotation is scoped to `tail` only).

## Complexity
Small (one-line template change for the recommended fix).

---
Filed by `/printing-press-retro`. Full retro doc (drops, skips, what-went-right) retained in run manuscripts.

cbolden15:
## What

The generated `tail` command is a resilient poller: on a failed initial fetch it prints `warning: initial fetch failed: ...` and `return nil` (exit 0). dogfood's `error_path` probe runs `tail <invalid-resource>` and expects a non-zero exit, so **every printed CLI with a data layer fails the `error_path` check on `tail`** unless it's hand-annotated with `pp:no-error-path-probe`.

This adds `"pp:no-error-path-probe": "true"` to the `tail` command template so dogfood skips the inapplicable `error_path` probe.

## Why it's a generator fix (not per-CLI)

`tail.go.tmpl` is emitted for every CLI with a syncable data layer, so the failure is deterministic and recurs on every such CLI:

- **stripe**: Phase 5 live dogfood `error_path` on `tail __printing_press_invalid__` → exit 0, scored FAIL; cleared only by a manual annotation.
- **twilio** (`library/twilio/internal/cli/tail.go`): byte-identical template (warning + `return nil`), zero `pp:no-error-path-probe` annotations — would fail the same probe.

## Changes

- `internal/generator/templates/tail.go.tmpl`: add `pp:no-error-path-probe` to the tail command Annotations.
- `internal/generator/generator_test.go`: broaden `TestTemplatesEmitReadOnlyAnnotation`'s regex to match the `mcp:read-only` annotation inside an `Annotations` map literal that carries additional annotations (the prior regex assumed read-only was the sole entry, so the new tail annotation broke the exact match). `[^}]*` stays within a single map literal, so the per-command match counts are unchanged.

## Verification

- `go test ./internal/generator/` → `ok` (full package).
- `go test ./internal/pipeline/` → `ok` (annotation consumer).
- End-to-end: rebuilt the binary, generated a CLI, confirmed `internal/cli/tail.go` now emits `Annotations: map[string]string{"mcp:read-only": "true", "pp:no-error-path-probe": "true"}`.

## Alternative considered

Make `tail` return a non-zero error when the *initial* fetch fails (before entering the poll loop) — arguably more correct, but changes runtime behavior. Went with the annotation as the low-risk fix; happy to switch if you'd prefer the behavioral change.

fix #2968

### DIFF
--- a/internal/generator/generator_test.go
+++ b/internal/generator/generator_test.go
@@ -742,7 +742,7 @@ func TestGenerateAgentContextCommand(t *testing.T) {
 	// agent-context only reads CLI tree state and emits JSON to stdout.
 	// The runtime walker uses this annotation to set readOnlyHint on
 	// the resulting MCP tool so hosts skip the per-call permission prompt.
-	assert.Regexp(t, `Annotations:\s+map\[string\]string\{"mcp:read-only":\s*"true"\}`, src,
+	assert.Regexp(t, `Annotations:\s+map\[string\]string\{[^}]*"mcp:read-only":\s*"true"[^}]*\}`, src,
 		"agent_context.go must carry mcp:read-only annotation")
 
 	// The subcommand must be registered in root.go so the CLI picks it up.
@@ -16847,7 +16847,10 @@ func TestToKebab_SnakeCaseInput(t *testing.T) {
 // export.go.tmpl (--output writes user files), import/sync/feedback/graphql_sync (writes).
 func TestTemplatesEmitReadOnlyAnnotation(t *testing.T) {
 	t.Parallel()
-	annotationRE := regexp.MustCompile(`Annotations:\s+map\[string\]string\{"mcp:read-only":\s*"true"\}`)
+	// Match the read-only annotation inside an Annotations map literal,
+	// tolerating additional annotations (e.g. pp:no-error-path-probe on tail).
+	// [^}]* stays within a single map literal so per-command counts hold.
+	annotationRE := regexp.MustCompile(`Annotations:\s+map\[string\]string\{[^}]*"mcp:read-only":\s*"true"[^}]*\}`)
 
 	cases := []struct {
 		template string
--- a/internal/generator/templates/tail.go.tmpl
+++ b/internal/generator/templates/tail.go.tmpl
@@ -24,7 +24,7 @@ func newTailCmd(flags *rootFlags) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:         "tail [resource]",
 		Short:       "Stream live changes by polling the API at regular intervals",
-		Annotations: map[string]string{"mcp:read-only": "true"},
+		Annotations: map[string]string{"mcp:read-only": "true", "pp:no-error-path-probe": "true"},
 		Long: `Tail streams live data changes by polling the API at configurable intervals.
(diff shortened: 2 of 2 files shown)


cbolden15:
Ready for merge whenever you have a moment, @mvanhorn / @tmchow. All 26 CI checks are green, conversations resolved, and the code-approval requirement is auto-waived (no `.github/`, `scripts/`, or CODEOWNERS changes). The only outstanding Mergify requirement is the `ready-to-merge` label, which needs a maintainer. Thanks!

tmchow:
@cbolden15 when you open PRs that resolve issues, do not use a `refs` you should use `fixes`.  I've updated the PR description but moving forward please ensure your agent follows the process correctly.

PR #3023, opened 2026-06-17
fix(generator): qualify FTS ORDER BY rank to the FTS table

### LINKED ISSUE #2973: Generator: qualify FTS `ORDER BY rank` to the FTS table (ambiguous when a resource has a `rank` column)
## What happens
Generated `Search<Resource>` store methods emit:

```sql
SELECT t.data FROM "<r>" t
 JOIN "<r>_fts" ON "<r>_fts".rowid = t.rowid
 WHERE "<r>_fts" MATCH ?
 ORDER BY rank LIMIT ?
```

FTS5 exposes an implicit `rank` relevance column on `<r>_fts`. When the resource also has a domain field named `rank` (promoted to a typed column on `<r>`), the unqualified `ORDER BY rank` is **ambiguous** and SQLite errors at query time:

```
SQL logic error: ambiguous column name: rank (1)
```

This breaks `Search<Resource>` for that resource at runtime (not just in tests), and fails the generated `TestSearch<Resource>QuotesFTSQuerySyntax`.

## Where
`internal/generator/` — the store search-method template (emits the SQL above into `internal/store/store.go`).

## Repro
Generate a CLI from any spec whose syncable resource has a field named `rank` (e.g. a bid-leveling / ranking / leaderboard / search-result entity). `go test ./internal/store/` fails with `ambiguous column name: rank`. Observed on a real run: OpenConstructionERP's `bid_management` resource (a `rank` bidder-ranking field) → `SearchBidManagement` broke.

## Fix
Qualify the order column to the FTS table:

```sql
 ORDER BY "<r>_fts".rank LIMIT ?
```

`<r>_fts.rank` always exists (FTS5 built-in) and is the intended relevance ordering, so this is **always-correct and zero-downside** — it never affects resources without a `rank` column. Canonical SQLite FTS5 pattern.

Optional follow-up: audit other generated FTS SQL for unqualified references to FTS5 auxiliary names.

## Test
- Positive: resource WITH a `rank` field — `Search<Resource>(q)` returns rows without error; the generated FTS-syntax test passes.
- Negative: resource WITHOUT a `rank` field — still orders by relevance; no regression (only the qualified column name changes in golden output).

## Note
Shipcheck's verify/dogfood legs did not catch this — they don't run the generated store unit tests; it only surfaced under a manual `go test ./...`.

_Filed via /printing-press-retro. Full retro doc + manuscripts available on request (kept local — the bug is generic, no API-specific artifacts needed)._

Sathvik-1007:
## Intent

Fixes #2973

When a synced resource has a domain field named `rank`, the unqualified `ORDER BY rank` in generated store search methods is ambiguous — SQLite errors with `ambiguous column name: rank (1)` at query time.

## Approach

Qualify the `rank` column to the FTS table alias in every emitted search query:

- Generic search (resources_fts aliased `f`) → `ORDER BY f.rank`
- Typed-table search (per-domain FTS table) → `ORDER BY "<table>_fts".rank`
- Insights similar query (resources_fts aliased `fts`) → `ORDER BY fts.rank` + `SELECT ... fts.rank`

Always-correct — FTS5 always exposes a built-in `rank` column, so qualifying it never breaks resources without the field.

## Scope

Primary area: `internal/generator/templates/`

Why this belongs in this repo: the bug is in the generator templates that emit store code for every printed CLI. Fixing it here fixes all future generations.

## Catalog Justification

N/A

## Risk

None. The qualified column reference is semantically identical when no ambiguity exists — it just becomes necessary when one does.

## Output Contract

Golden fixtures updated for the two affected cases (sync-walker, learn-loop). The typed-table template change affects per-domain search methods only when a domain table is generated.

## Verification

```
grep -rn "ORDER BY rank" --include="*.go" --include="*.tmpl" .
# (empty — no unqualified references remain)
```

## AI / Automation Disclosure

- [x] No AI or automation was used

### DIFF
--- a/internal/generator/templates/insights/similar.go.tmpl
+++ b/internal/generator/templates/insights/similar.go.tmpl
@@ -97,12 +97,12 @@ work items, tickets, or tasks.`,
 			}
 
 			// Use FTS5 to find similar items (exclude the source item)
-			query := `SELECT r.id, r.resource_type, r.data, rank
+			query := `SELECT r.id, r.resource_type, r.data, fts.rank
 				FROM resources_fts fts
 				JOIN resources r ON r.id = fts.id AND r.resource_type = fts.resource_type
 				WHERE resources_fts MATCH ?
 				  AND NOT (r.id = ? AND r.resource_type = ?)
-				ORDER BY rank
+				ORDER BY fts.rank
 				LIMIT ?`
 
 			rows, err := db.DB().Query(query, searchText, itemID, sourceType, limit)
--- a/internal/generator/templates/store.go.tmpl
+++ b/internal/generator/templates/store.go.tmpl
@@ -972,7 +972,7 @@ func (s *Store) Search(query string, limit int, resourceTypes ...string) ([]json
 			 JOIN resources_fts f ON r.id = f.id AND r.resource_type = f.resource_type
 			 WHERE resources_fts MATCH ?
 			 AND r.resource_type = ?
-			 ORDER BY rank
+			 ORDER BY f.rank
 			 LIMIT ?`,
 			matchQuery, resourceType, limit,
 		)
@@ -995,7 +995,7 @@ func (s *Store) Search(query string, limit int, resourceTypes ...string) ([]json
 		`SELECT r.data FROM resources r
 		 JOIN resources_fts f ON r.id = f.id AND r.resource_type = f.resource_type
 		 WHERE resources_fts MATCH ?
-		 ORDER BY rank
+		 ORDER BY f.rank
 		 LIMIT ?`,
 		matchQuery, limit,
 	)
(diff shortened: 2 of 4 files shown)


Sathvik-1007:
Good catch on the typed-table golden coverage gap. The existing golden cases (sync-walker, learn-loop) don't exercise the `{{if .FTS5}}` branch since neither spec has domain tables with FTS enabled.

The `generated-output-compile` CI check does compile specs that hit this path (it compiles every catalog entry), so a broken `safeNameSuffix` expansion would still fail CI — just not via golden diff.

Happy to add a typed-table golden fixture in a follow-up if the maintainers want it locked in, but keeping this PR scoped to the bugfix for now.

tmchow:
Maintainer triage: rebased onto current `main`. Greptile P2 on typed-table golden coverage is a fair follow-up, but `generated-output-compile` already compiles catalog entries that exercise per-domain FTS search — acceptable to land the bugfix without a new golden in this PR.

Issue #3063, opened 2026-06-18
fix(cli): refresh OAuth2 authorization-code bearer tokens for doctor probes


tmchow:
Generated CLIs that model an OAuth2 authorization-code flow as `bearer_token` can keep using stale per-call bearer fields after a refresh writes a new `access_token` / `refresh_token` tuple. Doctor auth probes also precompute and override the auth header/query parameter, so the client's refresh-on-401 retry path can refresh but still replay the stale probe credential.

This surfaced while investigating the x-twitter printed CLI user-context OAuth2 refresh path: a refresh token was present, but doctor reported the user-context token invalid after HTTP 401 instead of refreshing and probing with the rotated access token.

Intended fix: make authorization-code bearer auth use the minted/refreshed OAuth access token as the generated request credential, persist rotated refresh tokens as today, and let doctor probes use the refresh-capable client auth injection path rather than stale header/query overrides.

tmchow:
Picking this up now for the x-twitter OAuth2 refresh follow-up.

tmchow:
Triage note: this appears to have been fixed by merged PR #3064 (`fix(generator): refresh authorization-code bearer auth`), which implements the OAuth access-token/doctor probe refresh path described here. Closing as already fixed; please reopen if there is a remaining facet not covered by #3064.

PR #3064, opened 2026-06-18
fix(generator): refresh authorization-code bearer auth

tmchow:
## Summary

Fix generator behavior for OAuth2 authorization-code APIs that use bearer-token request auth.

The generated runtime already had refresh-token plumbing, but authorization-code bearer CLIs could still ignore the refreshed access token because generated `Config.AuthHeader()` did not treat those APIs as OAuth access-token auth. Generated doctor probes also captured auth into header/query overrides before probing, which could replay stale credentials instead of letting the client inject refresh-capable auth.

This PR:

- treats `bearer_token` + OAuth2 authorization-code + token URL as OAuth access-token auth
- keeps refreshed access tokens ahead of stale per-call/generated bearer fields
- updates generated doctor probes to let the client inject auth so refresh-on-401 can work
- adds generator/runtime regression coverage for refreshed access-token precedence and rotated refresh-token persistence

Related published-library symptom/fix: mvanhorn/printing-press-library#1273
Issue: #3063

## Verification

Passed:

```bash
go test ./internal/generator -run 'TestAuthorizationCodeBearerUsesRefreshedAccessTokenOverStalePerCallToken|TestGeneratedDoctor_AuthVerifyPathProbesEndpoint' -count=1
scripts/golden.sh verify
scripts/verify-generator-output.sh
git diff --check
```

Also ran:

```bash
go test ./...
```

It failed in `internal/generator` during the existing generated-CLI `govulncheck ./...` quality gate. The targeted generator tests, golden verification, and generated-output verification all passed; the visible failure is the same generated verifier/govulncheck path Woz hit before PR creation, not a compile or targeted regression failure in this change.


tmchow:
@Greptileai review

PR #3065, opened 2026-06-18
feat(generator): add xml response_format for XML-backed APIs

coopdogGGs:
## Intent

Generated CLIs/MCP servers could not cleanly consume APIs that return XML. The
HTTP layer already let XML bytes through as text, but with the default
`response_format: json` the body was opaque to `--json`, `--select`, table
output, and MCP tools. This adds first-class XML support so an XML API behaves
like any JSON API in a printed CLI.

Issue: N/A

## Approach

Add `xml` as a fifth `response_format`, mirroring how `csv` already turns a
non-JSON body into a generic JSON-compatible structure inside the generated
client. The decision was XML→JSON normalization (Design A) over an
`xml_extract` selector contract mirroring `html_extract` (Design B): A matches
the CSV precedent, needs no per-mode DOM/XPath machinery or profiler
propagation, and is enough for the motivating consumer (read-only XML APIs).

Three layers:
- **spec**: `ResponseFormatXML`, `Endpoint.UsesXMLResponse`,
  `APISpec.HasXMLResponse`, and validation accepts `xml`.
- **parser**: XML-only success responses are auto-detected and set
  `response_format: xml` plus an `Accept: application/xml` override so the
  server returns XML instead of 406-ing the default `application/json`. Mixed
  JSON+XML responses keep the JSON default.
- **generator/client**: when any endpoint uses `response_format: xml`, emit a
  `cliutil.XMLToJSON` helper (stdlib `encoding/xml`, BadgerFish-lite: `@attr`,
  `#text`, repeated siblings → array, graceful raw pass-through on decode
  failure). The client runs XML-only success bodies through it before
  returning, so every downstream consumer sees ordinary JSON.

The helper, the client branch, and the `isXMLResponseContentType` helper are
all gated on `HasXMLResponse`, so JSON-only CLIs are byte-identical to before.

## Scope

Primary area: `internal/spec`, `internal/openapi`, `internal/generator`
(templates + emission).

Why this belongs in this repo: this is core Printing Press generation behavior
— a new response-format class that any future XML-backed spec can use — not a
fix to one printed CLI.

## Catalog Justification

N/A — no `catalog/*.yaml` or `catalog/specs/**` changes. (Motivated by
BoardGameGeek's XML API, but no catalog entry is added here; that remains a
separate follow-up gated on BGG application approval.)

## Risk

Low and contained. All new generation is gated on `HasXMLResponse`, which is
false for every existing spec, so JSON/CSV/HTML/binary CLIs are unaffected. The
new parser branch only fires when a success response is XML-only (never when
JSON is also offered), so it cannot reroute existing JSON endpoints. Worst case
for a future XML API is a decode failure, which returns the raw XML bytes
unchanged rather than corrupting the body.

## Output Contract

Generated output changes only for specs that declare/auto-detect
`response_format: xml`:
- New emitted file `internal/cliutil/xml_parse.go` (gated on `HasXMLResponse`).
- New `isXMLResponseContentType` helper and a normalization branch in the
  client's success path (both gated).

Evidence:
- Emitted-code + compiled-CLI assertion: `TestGenerateXMLResponseParseHelper`
  generates an XML-API CLI, then compiles and runs the generated
  `cliutil.XMLToJSON` against table cases (attributes, nesting, repeated
  siblings → array, mixed attr+text, malformed pass-through).
- Gating proof: `TestGenerateJSONOnlyOmitsXMLResponseParseHelper` asserts
  JSON-only CLIs emit no `xml_parse.go`.
- No-drift proof: regenerated the `generate-golden-api` fixture; all 39
  generated `.go` files are byte-identical to the committed golden expected
  output (only the copyright-owner line differs, an author-handle artifact).

## Verification

- `go build ./...` — pass
- `go vet ./internal/spec/... ./internal/openapi/... ./internal/generator/...` — pass
- `gofmt -l` on touched files — clean
- `go test ./internal/spec/...` — pass
- `go test ./internal/openapi/...` — pass except the pre-existing
  Windows-only failure `TestGenerateDataEnvelopeAllOfBodyFlags` (builds the
  binary without the `.exe` suffix at `parser_test.go:1411`); passes when
  skipped. Confirmed it fails identically on `main`.
- `go test ./internal/generator/... -run XMLResponse` — pass (compiles + runs
  generated helper)
- `scripts/golden.sh verify` — Not conclusive locally: on the author's Windows
  box it reports 31 changed cases, but `main` reports the **same 31** (CRLF
  line-ending normalization + local Go import ordering), so the baseline itself
  does not match here. None of the diffs are XML-related. The per-file
  byte-identity check above isolates this change's zero drift; CI (Linux) runs
  the authoritative golden verify.

## AI / Automation Disclosure


tmchow:
@coopdogGGs Your agents needs to follow directions in AGENTS.md if it's going to open PRs. Specifically, it needs to babysit the PRs to address code review feedback.  

In addition to addressing code review feedback, please fix the Greptile comment:

> The global Accept fallback is changed from application/json to application/xml for any spec that contains at least one XML endpoint. Since the spec-level HasXMLResponse flag drives both the Accept override and the response-normalization branch, a spec that legitimately mixes XML and JSON endpoints would have all its JSON endpoints default to Accept: application/xml — and have any application/xml-typed success response silently normalized, regardless of which endpoint produced it. The per-endpoint headerOverride already correctly pins Accept: application/xml for XML endpoints, so the global change is redundant for pure XML specs and broken for mixed ones.

When you've done your fixes, feel free to mark as ready for review

coopdogGGs:
## Review follow-up (commit b87db4ba)

Thanks for the thorough pass. Addressed the two **P1** findings; the two **P2** findings are deliberate scope decisions explained below.

### ✅ P1 — Global Accept override breaks JSON endpoints in mixed specs (`client.go.tmpl`)
### ✅ P1 — XML normalization is spec-scoped, not endpoint-scoped (`client.go.tmpl`)

Both stem from the same root cause: the spec-wide XML behavior was gated on `HasXMLResponse()` (spec has **at least one** XML endpoint). I added `APISpec.AllResponsesXML()` — true only when **every** endpoint is `response_format: xml` — and moved all four spec-wide gates onto it:

- the `application/xml` Accept default,
- the `XMLToJSON` normalization call site,
- the `isXMLResponseContentType` helper,
- `xml_parse.go` emission.

So a mixed spec now keeps `application/json` and never XML-normalizes its JSON endpoints. XML endpoints in a mixed spec still get their per-endpoint `Accept: application/xml` override from the parser (the layer you correctly identified) — only the spec-wide fallback is withheld. Pure-XML APIs (BoardGameGeek) are unchanged. Added a spec unit test for `AllResponsesXML` and a generator test asserting a mixed XML/JSON spec emits no spec-wide XML code. All golden cases are non-XML, so output is byte-identical.

> Full per-endpoint XML in a *mixed* spec (so a mixed spec's XML endpoints still normalize) is intentionally out of scope here — it needs a per-request marker like `binaryResponse`. Filed as a follow-up; this PR keeps the conservative, safe behavior.

### ℹ️ P2 — XML namespace collision (`cliutil_xml_parse.go.tmpl`)
Accepted scope. The mapper is intentionally BadgerFish-*lite* and keys by local name; as the review notes, this is correct for flat XML (BGG and the vast majority of XML REST APIs). Namespace-qualified keys would complicate the JSON shape for every consumer to handle a case no current spec hits. Documenting as a known limitation rather than expanding the output contract.

### ℹ️ P2 — XML decode failure surfaces as opaque JSON error (`client.go.tmpl`)
Accepted scope. The raw-bytes passthrough on decode failure is a deliberate "never corrupt the body" choice — the caller still receives the original XML to inspect. A structured `{"__xml_raw": …}` envelope is a reasonable future enhancement, but changing the failure contract is beyond this PR; noting as a follow-up.

tmchow:
@coopdogGGs still ned this fixed:

> The global Accept fallback is changed from application/json to application/xml for any spec that contains at least one XML endpoint. Since the spec-level HasXMLResponse flag drives both the Accept override and the response-normalization branch, a spec that legitimately mixes XML and JSON endpoints would have all its JSON endpoints default to Accept: application/xml — and have any application/xml-typed success response silently normalized, regardless of which endpoint produced it. The per-endpoint headerOverride already correctly pins Accept: application/xml for XML endpoints, so the global change is redundant for pure XML specs and broken for mixed ones.
> 
> internal/generator/templates/client.go.tmpl — both the Accept fallback and the XML normalization branch are gated at the spec level rather than the endpoint level.

tmchow:
@greptileai

PR #3073, opened 2026-06-19
fix(dogfood): skip top-level login alias in live dogfood matrix

coopdogGGs:
## What

Add `login` to `liveDogfoodFrameworkSkip` so the live dogfood matrix does not probe the top-level `login` alias.

## Why

A generated CLI may expose `login` at the top level as a convenience alias for `auth login`. That command launches the OAuth authorization-code browser flow and never issues a probeable API call — it is framework scaffolding, exactly like `auth`, `completion`, `help`, and `version`, all of which are already skipped.

Because `login` was missing from the skip set, `collectLiveDogfoodCommands` treated it as a leaf API command and live dogfood tried to probe it, producing a guaranteed failure for every CLI that exposes the alias. This surfaced while generating an Oura CLI (`oura-pp-cli`), whose OAuth template registers a top-level `login` alias: live dogfood reported a spurious failure on `login` even though the OAuth flow and every real API call succeeded.

## Change

- Add `"login": true` to `liveDogfoodFrameworkSkip`.
- Document the map's intent (framework scaffolding vs. API surface), calling out why `login` belongs there.
- Add `TestCollectLiveDogfoodCommandsSkipsLoginAlias`, a Windows-friendly regression test asserting that a top-level `login` alias and the `auth` subtree are both excluded while real API subtrees survive.

## Verification

- `go build ./...` — pass
- `go vet ./internal/pipeline/` — pass
- `go test ./internal/pipeline/ -run TestCollectLiveDogfoodCommandsSkipsLoginAlias` — pass

No behavior change for CLIs that do not expose a top-level `login` alias.

### DIFF
--- a/internal/pipeline/live_dogfood.go
+++ b/internal/pipeline/live_dogfood.go
@@ -458,11 +458,17 @@ func discoverLiveDogfoodCommands(binaryPath string) ([]liveDogfoodCommand, error
 	return commands, nil
 }
 
+// liveDogfoodFrameworkSkip names top-level commands that are framework
+// scaffolding rather than API surface, so live dogfood does not probe them.
+// "login" is the top-level alias for `auth login`: it launches the OAuth
+// browser flow and never makes a probeable API call, so it belongs here
+// alongside the other framework commands.
 var liveDogfoodFrameworkSkip = map[string]bool{
 	"agent-context": true,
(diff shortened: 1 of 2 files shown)


tmchow:
@mergifyio queue

PR #3075, opened 2026-06-19
fix(cli): bind MCP array/object body params natively (WithArray/WithObject)

### LINKED ISSUE #3074: MCP array/object body params bound as JSON-string (WithString) instead of native type
## Summary

The generator binds **array** and **object** MCP tool params as JSON-Schema `type:"string"` instead of their native type. For **body** params this is a wire-correctness bug, not just a schema cosmetic.

## Root cause

`internal/generator/generator.go` — `mcpBindingFunc()` routes array/object through its `default` branch:

```go
func mcpBindingFunc(t string) string {
	switch primitiveKind(t) {
	case "int", "float": return "WithNumber"
	case "bool":         return "WithBoolean"
	default:             return "WithString"   // <- array & object land here
	}
}
```

So every generated MCP tool declares array/object params as `WithString` (`{"type":"string"}`).

## Impact (body params)

In the generic handler (`internal/generator/templates/mcp_tools.go.tmpl`, `makeAPIHandler`), a body param is stored verbatim and the body map is JSON-marshaled:

```go
case "body":
    bodyArgs[binding.WireName] = v   // v is a *string* when bound via WithString
```

With `WithString`, an agent passes a JSON-encoded string and the request body becomes `{"tags":"[...]"}` — a JSON **string** — when the API expects `{"tags":[...]}` — a JSON **array**. The CLI path is unaffected (each `internal/cli/*.go` command `json.Unmarshal`s the `--flag` string into a slice before building the body); only the MCP path ships the malformed value.

Concretely, AQL's MCP-contract check flags 4 such params in a Readwise CLI generated by the current machine (`documents_save.tags`, `documents_update.tags`, `documents_bulk-update.updates`, `highlights_create.highlights`) as `param_type_mismatch` (subject `string` vs API `array`). Any printed CLI with array/object body params has the same latent bug.

## Proposed fix

Map `array → WithArray` and `object → WithObject` in `mcpBindingFunc` so these params bind to their native JSON type; the handler then serializes them correctly with no body-path change. Harden `formatMCPParamValue` to JSON-encode a native composite in its default branch (covers the rare query/path-located array/object param). Keep the CLI surface as JSON-string flags (cobra cannot accept a native array) — an intentional CLI=JSON-string / MCP=native asymmetry. Object params with declared fields already flatten to dotted scalar flags, so `WithObject` only targets opaque objects; the `body_json` oneOf/anyOf fallback stays `WithString`.

I have a fix ready and will open a PR.

vinnyp:
## Problem

`mcpBindingFunc` mapped **array**/**object** params to `mcplib.WithString`, so generated MCP tools declared them as JSON-Schema `type:"string"`. For **body** params this is a wire-correctness bug: the generic handler stores the value verbatim (`bodyArgs[wire] = v`) and JSON-marshals the body map, so an array param shipped as a *stringified* array — `{"tags":"[...]"}` — to APIs that expect `{"tags":[...]}`. The CLI path was unaffected (it `json.Unmarshal`s the JSON-string flag before building the body); only the MCP path sent the malformed value. See #3074.

## Fix

- `mcpBindingFunc`: `array → WithArray`, `object → WithObject`. The handler then serializes the native `[]any` / `map[string]any` correctly with **no body-path change**.
- `formatMCPParamValue`: JSON-encode a native composite in its `default` branch (covers the rare query/path-located array/object param so it emits valid JSON instead of Go's `%v`).
- Object params with declared fields still flatten to dotted scalar flags, so `WithObject` only targets opaque objects; the `body_json` oneOf/anyOf fallback stays `WithString`.

## Consumer impact / migration

This changes the MCP tool input-schema for affected params from `string` to `array`/`object`. **Agents now pass a native array/object, not a JSON-encoded string.** Because the old stringified form was already non-functional on the wire (the API received a quoted string where it required an array), no *working* caller is broken — but a host that validates args against the input schema will now correctly reject the old string form. Already-published CLIs with array/object body params are reprint/amend candidates.

## CLI surface unchanged

Array/object **flags** remain JSON-string (`--tags '[...]'`) — cobra cannot accept a native array. Intentional asymmetry: **CLI = JSON-string flag, MCP = native typed**.

## Tests

- `TestMCPBindingFunc` — unit mapping incl. array→`WithArray`, object→`WithObject` (mutation guard: revert the cases → red).
- `TestGenerateInternalYAMLBodyObjectSchema` — generated MCP output object→`WithObject`, array→`WithArray`; CLI-side `StringVar`/`json.Unmarshal` assertions kept (locks the asymmetry); asserts `formatMCPParamValue`'s `json.Marshal` default branch.
- `TestGenerateBodyNameAcrossCLISurfaces` — array cursor (`startAfter`) MCP binding native; CLI flag stays string.
- `TestGenerateMultipartRequestBodyUsesMultipartClient` / `TestGenerateFormRequestBodyUsesFormClient` — object/array body param on `multipart/form-data` and `x-www-form-urlencoded` endpoints binds native and flows through the `json.Marshal` field helper.
- Golden fixtures regenerated (`scripts/golden.sh update`).

Gates green locally: `go build ./...`, `go vet ./...`, `go test ./...`, `scripts/golden.sh verify`, `scripts/verify-generator-output.sh`.

> **Note for maintainers:** the `internal/generator` package test run is ~9–10 min on my machine and brushes the default 10-minute `go test` timeout (pre-existing; this PR only adds schema-shape assertions + golden regen). CI may want a longer `-timeout` for that package to avoid a spurious timeout-red.

Fixes #3074

### DIFF
--- a/internal/generator/body_nested_object_test.go
+++ b/internal/generator/body_nested_object_test.go
@@ -204,13 +204,18 @@ resources:
 	mcpGot := string(mcpSrc)
 	for _, want := range []string{
 		`mcplib.WithString("query", mcplib.Required(), mcplib.Description("GraphQL document"))`,
-		`mcplib.WithString("variables", mcplib.Description("GraphQL variables"))`,
+		`mcplib.WithObject("variables", mcplib.Description("GraphQL variables"))`,
 		`mcplib.WithBoolean("serializer-settings-include-nulls"`,
-		`mcplib.WithString("queries"`,
+		`mcplib.WithArray("queries"`,
 		`PublicName: "serializer-settings-include-nulls", WireName: "includeNulls", Location: "body", BodyPath: []string{"serializerSettings", "includeNulls"}`,
 		`setNestedBodyArg(bodyArgs, binding.BodyPath, v)`,
 	} {
 		require.Containsf(t, mcpGot, want, "expected generated MCP fragment %q", want)
 	}
 	require.NotContains(t, mcpGot, `mcplib.WithString("queries-query"`)
+
+	// formatMCPParamValue must JSON-encode a native composite in its default
+	// branch (a query/path-located array/object param now arrives as []any /
+	// map[string]any instead of a string); Go's "%v" would emit "[a b c]".
+	require.Regexp(t, `(?s)func formatMCPParamValue\(v any\) string \{.*?json\.Marshal\(v\)`, mcpGot)
 }
--- a/internal/generator/cursor_param_test.go
+++ b/internal/generator/cursor_param_test.go
@@ -144,6 +144,34 @@ func TestMCPBindingFunc(t *testing.T) {
 	}
 }
 
+// TestMCPParamDefaultValue pins that a composite (array/object) param default is
+// serialized as JSON — matching how native array/object live values are now
+// encoded — while scalar defaults keep their "%v" rendering. Without the JSON
+// branch an array default would be injected on the wire as Go's "[a b c]".
+func TestMCPParamDefaultValue(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		name   string
+		p      spec.Param
+		want   string
+		wantOK bool
+	}{
+		{"nil default", spec.Param{Type: "array"}, "", false},
+		{"scalar int", spec.Param{Type: "integer", Default: 25}, "25", true},
+		{"scalar string", spec.Param{Type: "string", Default: "later"}, "later", true},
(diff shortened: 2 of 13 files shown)


vinnyp:
Addressed the Greptile summary finding in 72e1f734: `mcpParamDefaultValue` now JSON-encodes composite (array/object) defaults instead of `fmt.Sprintf("%v", ...)`, so an array/object param with a declared default is no longer injected on the wire as Go's `[a b c]` — consistent with the `json.Marshal` encoding used for native live values. Added unit test `TestMCPParamDefaultValue` as a guard. Scalar defaults are unchanged.

PR #3087, opened 2026-06-20
fix(generator): reject empty positional path-param values with a clear error

cathrynlavery:
## Problem

Required positional path params are only **presence**-checked (`len(args)`), never checked for an empty value. So an empty-string argument — most commonly a shell expansion of an unset env var — passes the check and gets substituted straight into the request path, producing a malformed URL and a confusing *remote* error.

Real example from a generated `cloudflare-pp-cli` session, where `$CLOUDFLARE_ACCOUNT_ID` was unset:

```
$ cloudflare-pp-cli accounts pages project-get-project ronanrx-com "$CLOUDFLARE_ACCOUNT_ID" --json
Error: GET /accounts//pages/projects/ronanrx-com returned HTTP 400:
  {"success":false,"errors":[{"code":7003,"message":"Could not route to
   /client/v4/accounts/pages/projects/ronanrx-com, perhaps your object
   identifier is invalid?"}]}
```

The `//` is the empty `account_id`. The user/agent has no way to tell from that 400 that the real issue is an empty argument, which drives repeated retries.

## Fix

Both the endpoint and promoted command templates now also reject an **empty value** when the positional is a path param, returning the existing local `"<name> is required"` usage error *before* any request is made:

```go
if len(args) < N+1 || args[N] == "" {   // path params only
    return usageErr(fmt.Errorf("<name> is required\nUsage: ..."))
}
```

- Gated on `pathContainsParam`, so **non-path** positionals are unaffected (no behavior change there).
- Import-safe — uses `== ""`, no new imports in generated files.
- Catches the exact failure mode (empty path segment → malformed URL) at the CLI boundary with the clear message that already exists for the missing-arg case.

## Validation

- Goldens regenerated; the **only** generated-code change is the presence-check line (e.g. `projects_tasks_update-project.go`: `if len(args) < 2 || args[1] == "" {`).
- `scripts/golden.sh verify` — 31/31 pass.
- `go test ./internal/generator/` — pass.
- `go build ./...` — pass.

Surfaced from real generated-CLI session transcripts.

### DIFF
--- a/internal/generator/path_param_positional_index_test.go
+++ b/internal/generator/path_param_positional_index_test.go
@@ -117,11 +117,15 @@ func TestPathParamArgsIndexUsesPositionalOrdinal(t *testing.T) {
 	}
 
 	// 1) Single positional with a non-positional query before it: args[0],
-	// no len-check (Cobra's MinimumNArgs(1) covers the first positional).
+	// with an explicit empty-string guard because Cobra's MinimumNArgs(1)
+	// allows callers to pass "" from an empty shell variable expansion.
 	tagsList := read(filepath.Join("internal", "cli", "tags_contacts_list-for-tag.go"))
 	assert.Contains(t, tagsList,
 		`path = replacePathParam(path, "tagId", args[0])`,
 		"single positional path param must use args[0] regardless of declared query params before it")
+	assert.Contains(t, tagsList,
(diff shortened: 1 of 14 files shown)


tmchow:
@mergifyio queue

PR #3095, opened 2026-06-20
fix(cli): qualify FTS ORDER BY rank to the FTS table

e-jung:
## Intent

`search` is broken on any printed CLI whose resource has a typed `rank` column (rankings, standings, priorities — e.g. pp-espn). The generated FTS5 search SQL emits an unqualified `ORDER BY rank`, which SQLite rejects at query time with `SQL logic error: ambiguous column name: rank (1)` because `rank` is also an FTS5 special column. The store opens fine and sync populates the index; the failure surfaces only the first time FTS MATCH returns a row.

Issue: #2973 (filer: "always-correct, zero-downside")

Refs #2973

## Approach

Qualify every FTS5 `rank` reference to whatever name actually scopes the FTS table in that query, in three templates:

- `internal/generator/templates/store.go.tmpl` — per-resource `Search<Resource>`: `ORDER BY {{safeNameSuffix .Name "_fts"}}.rank` (the per-resource FTS table has no alias, so the base name is the correct qualifier).
- `internal/generator/templates/store.go.tmpl` — generic `Search`/`SearchAll` (two query branches): `ORDER BY f.rank`, where `f` is the existing `resources_fts` alias from the JOIN.
- `internal/generator/templates/insights/similar.go.tmpl` — both the `SELECT … rank` and the `ORDER BY rank` become `fts.rank` (the `resources_fts fts` alias).

The non-obvious part: the qualifier must use the table **alias**, not the base table name. `resources_fts MATCH ?` still accepts the bare table name (an FTS5 quirk that makes the base name *look* usable), but a normal column reference like `.rank` resolves by alias once the table is aliased — so the literal `resources_fts.rank` fails with `no such column` on the aliased generic/similar queries. I verified this directly against SQLite before editing.

For resources without a `rank` column this is a no-op: the qualifier just names the only `rank` in scope, which is what SQLite resolved to anyway.

## Scope

Primary area: `cli` — generator templates that emit the printed-CLI store SQL.

Why this belongs in this repo: this is a generator bug, not a printed-CLI bug. Every past and future printed CLI with a rank-fielded resource inherits the broken query; fixing the machine (templates) fixes them all on next regen. A per-CLI fix in the public library would not compound, but the filer's "always-correct, zero-downside" judgment and the cross-CLI impact make this a machine fix.

## Catalog Justification

N/A — no `catalog/*.yaml` or `catalog/specs/**` changes.

## Risk

Low. The change is four SQL-string qualifications inside generated `SELECT` statements; it touches no Go control flow, no MCP surface, no auth, no publish flow, and no verifier/scorer. The only behavioral effect is that `search` now *works* on rank-fielded resources and is unchanged everywhere else (the qualifier names the sole `rank` in scope). Worst case if wrong: a typo in a table/alias name would surface as an immediate SQLite prepare error in the new regression test, not a silent regression.

## Output Contract

This PR changes templates → generated `store.go` (and `insights/similar.go`) output shifts.

Golden diff (intentional, exactly this and nothing else):
- `testdata/golden/expected/generate-learn-loop-api/.../internal/store/store.go`
- `testdata/golden/expected/generate-sync-walker-api/.../internal/store/store.go`

Both change the generic `Search` query from `ORDER BY rank` → `ORDER BY f.rank` (two occurrences each — the resource-type-filtered branch and the unfiltered branch). No other golden fixture changed.

Per-resource (`Search<Resource>`) and `insights/similar` qualification aren't exercised by these two golden fixtures because their specs have no FTS5 resource / no `similar` artifact; they are covered instead by the new emitted-code + runtime regression test (`TestGenerateStoreFTSOrderByRankQualified`), which builds a rank-fielded FTS5 resource and runs the generated `TestSearchItemsQuotesFTSQuerySyntax` against a SQLite DB with a typed `rank` column.

## Verification

Commands actually run (Go 1.26.4):
- `go test ./internal/generator` — clean (0 failures; ~24m on this box). The new `TestGenerateStoreFTSOrderByRankQualified` **fails without the fix** with the live error `SearchItems("10.0.0.1"): SQL logic error: ambiguous column name: rank` and passes with it.
- `scripts/golden.sh verify` — 31/31 pass (after `scripts/golden.sh update`, diff inspected and limited to the two `store.go` fixtures above).
- `scripts/verify-generator-output.sh` — 9/9 generated modules compile (covers `generate-learn-loop-api`, which carries the qualified store.go).
- `go build ./...` + `go vet ./internal/generator` — clean.
- Direct SQLite repro confirmed the fix resolves the ambiguity and that the alias form (`f.rank`/`fts.rank`) is required, not the base table name.

Not run: the full `go test ./...` package run hits the default 10m timeout on the two compile-heaviest packages (`internal/generator`, `internal/cli`) on this hardware; `internal/generator` was re-run with `-timeout 60m` and is fully green with zero `--- FAIL` anywhere in the suite. `govulncheck` needs network (unavailable here) and was not run locally.

## AI / Automation Disclosure

### DIFF
--- a/docs/solutions/runtime-errors/fts-order-by-rank-ambiguous-on-rank-column-2026-06-20.md
+++ b/docs/solutions/runtime-errors/fts-order-by-rank-ambiguous-on-rank-column-2026-06-20.md
@@ -1,0 +1,101 @@
+---
+title: "Generated FTS ORDER BY rank is ambiguous when a resource has a typed rank column"
+date: 2026-06-20
+category: runtime-errors
+module: internal/generator/templates
+problem_type: runtime_error
+component: store
+symptoms:
+  - "`search` on a printed CLI whose resource carries a typed `rank` column failed with `SQL logic error: ambiguous column name: rank (1)`"
+  - "Only resources with a `rank` response field (e.g. rankings/standings resources) were affected; rank-less resources searched fine"
+  - "The generic `Search` and the per-resource `Search<Resource>` both failed at query time the first time FTS MATCH produced a row"
+root_cause: code_generation_bug
+resolution_type: code_fix
+severity: high
+tags:
+  - sqlite
+  - fts5
+  - store
+  - search
+  - codegen
+  - ambiguous-column
+related_components:
+  - store
+  - insights
+issue: 2973
+---
+
+# Generated FTS `ORDER BY rank` is ambiguous on rank-fielded resources
+
+## Problem
+
+The store template (`internal/generator/templates/store.go.tmpl`) and the insights similar template (`internal/generator/templates/insights/similar.go.tmpl`) emitted FTS5 search queries with an unqualified `ORDER BY rank`:
+
+```sql
+-- per-resource Search<Resource>
+SELECT t.data FROM events t
+ JOIN events_fts ON events_fts.rowid = t.rowid
+ WHERE events_fts MATCH ?
+ ORDER BY rank LIMIT ?
+
+-- generic Search
+SELECT r.data FROM resources r
+ JOIN resources_fts f ON r.id = f.id AND r.resource_type = f.resource_type
+ WHERE resources_fts MATCH ?
+ ORDER BY rank LIMIT ?
+```
+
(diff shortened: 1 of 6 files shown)


e-jung:
Added insights/similar.go.tmpl coverage in commit d97eb79 — the new test renders the insights template (populates `VisionTemplateSet.Insights`) and asserts the generated similar.go emits the qualified `ORDER BY fts.rank` (and rejects bare `ORDER BY rank`). Resolving the thread.

tmchow:
@mergifyio queue

PR #3152, opened 2026-06-21
feat(publish): lightweight blobless+sparse library clone + force-add MCP source

aborruso:
## Summary

Two improvements to the `printing-press-publish` skill, both found while publishing a real CLI (a reverse-engineered, MCP-enabled CLI) to the public library.

### 1. Lightweight clone of the public library (blobless + shallow + sparse)

Step 5 currently does `git clone --depth 50` of `mvanhorn/printing-press-library`. That repo is a historical collection of hundreds of CLIs, so the working tree is multiple GB — a lot of bandwidth and disk to publish a single CLI, and painful on space-constrained machines (this surfaced as a "no space left on device" failure mid-publish).

Since the skill no longer regenerates or commits the `cli-skills/`/`registry.json` mirror (Step 6 already documents that the library regenerates them post-merge), the publish flow only touches the **target CLI's own directory**. So there's no need to materialize every other CLI's source.

This PR switches both clone paths (push-access and fork) to:

```bash
git clone --filter=blob:none --depth 1 --sparse "$REPO_URL" "$PUBLISH_REPO_DIR"
git -C "$PUBLISH_REPO_DIR" sparse-checkout set tools cli-skills library/<category>
```

The cone keeps `tools`, `cli-skills`, and the target `library/<category>` so the in-category find/rm/copy operations in Step 6 keep working on a real working tree. Cross-category collision checks (Step 5 pre-package snapshot and the Step 7 rename-suggestion check) are switched from `ls` to `git ls-tree`, which reads the full tree from the blobless clone and therefore stays correct despite the sparse working tree.

Result in practice: the clone drops from multiple GB to ~tens of MB.

### 2. Force-add the generated MCP source dir (`.gitignore` `*-pp-mcp` gotcha)

The library `.gitignore` has a `*-pp-mcp` rule meant for the built MCP **binary**, but the pattern also matches the generated MCP **source** directory `cmd/<cli-name>-pp-mcp/`. A plain `git add library/` silently skips `cmd/<cli-name>-pp-mcp/main.go`, and the PR then fails the **Validate MCPB manifest contract** CI check with `cmd/<cli-name>-pp-mcp directory is missing` (the manifest advertises an MCP binary with no entry point in the diff).

Step 8 now force-adds the source after `git add library/`:

```bash
git add -f "library/<category>/<api-slug>/cmd/<cli-name>-pp-mcp/main.go" 2>/dev/null || true
```

The built binary stays ignored; only `main.go` is added.

## Notes / scope

- Category-change reprints (a slug moving to a different category) would need the old path added to the sparse cone before the Step 6 `rm`; this is a rare path and left as a follow-up — the collision detection already surfaces it to the user.
- No behavioral change to packaging, validation, secret/PII scanning, or PR creation.

### DIFF
--- a/internal/pipeline/contracts_test.go
+++ b/internal/pipeline/contracts_test.go
@@ -542,7 +542,8 @@ func TestPublishSkillTracksCanonicalUpstreamAndOverwriteFlow(t *testing.T) {
 
 	assert.Contains(t, skill, "git remote add upstream")
 	assert.Contains(t, skill, "mvanhorn/printing-press-library")
-	assert.Contains(t, skill, "git fetch upstream")
+	assert.Contains(t, skill, "git fetch --filter=blob:none --depth 1 upstream")
+	assert.Contains(t, skill, "git fetch --filter=blob:none --depth 1 origin")
 	assert.Contains(t, skill, "git reset --hard upstream/main")
 	assert.Contains(t, skill, "git push --force-with-lease")
 }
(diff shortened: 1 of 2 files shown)


aborruso:
Thanks for taking this over @mvanhorn — I reviewed your two follow-up commits and they cover all three review points correctly, and more thoroughly than my original patch:

- **MCP path**: `cmd/<api-slug>-pp-mcp/` (api-slug, not the `-pp-cli` binary name) and force-adding the whole directory rather than just `main.go` — matches what actually shipped in the giustizia-amministrativa PR.
- **Cross-category sparse cone**: materializing the existing slug's path before the Step 6 filesystem operations, plus refreshing the cone in the "Subsequent publishes" path — handles the category-change reprint case I'd only flagged as a follow-up.
- Test coverage in `contracts_test.go` updated to lock in the new clone/add behavior.

Nothing further from my side. Happy to squash my original commit into yours if that's cleaner for merge.

PR #3229, opened 2026-06-25
fix: enable OData-style incremental syncs via conditions param

### LINKED ISSUE #3180: sync: OData/conditions APIs (ConnectWise Manage) emit empty syncResourceSinceParam → incremental inert + docs over-promise
## Symptom

The generated `connectwise-manage` CLI emits `func syncResourceSinceParam(resource string) string` with an **empty switch** — it returns `""` for every resource. In `syncResource`, that makes the incremental path inert:

```go
if effectiveSince != "" && sinceParam == "" {
    // emits {"event":"sync_warning","reason":"resource_not_incremental", ...}
    effectiveSince = ""
}
```

So `--since` (and the auto-checkpoint `last_synced_at`) have **no effect** on the sync, every `sync` is a full re-fetch, and on a large tenant (years of history) `resource_not_incremental` fires on nearly every resource. Meanwhile the generated `sync` `--help` advertises *"resumable incremental sync (only fetches new data since last sync)"* — a false promise for this API.

## Root cause

ConnectWise Manage (and other OData-style APIs) don't take a simple `?since=<value>` query param. They filter via a **`conditions`** expression: `conditions=lastUpdated > [2025-01-01T00:00:00Z]` (bracketed RFC3339). The generator's since-param detection didn't recognize this shape, so it emitted an empty switch rather than mapping the temporal filter.

## Asks

1. **Map the temporal filter for OData/`conditions` APIs.** Emit `syncResourceSinceParam(resource) = "conditions"` with a `syncResourceSinceParamFormat` that produces `lastUpdated > [<RFC3339>]` (using the resource's documented last-updated field). That makes `--since` and auto-checkpoint incremental actually work for ConnectWise-family connectors and dramatically cuts sync cost on established tenants.
2. **Don't advertise incremental when no temporal param is mapped.** When `syncResourceSinceParam` would be empty for every resource, the generated `sync` Long help (and the onboarded docs) should not claim incremental sync — or should state it's a full re-fetch. This avoids the over-promise users hit.

## Evidence

- Reported downstream: Servosity/msp-skills#146 (connectwise-manage; large-tenant timeouts + the false "sync is incremental" docs).
- The empty `syncResourceSinceParam`/`syncResourceSinceParamFormat` switches in the generated `internal/cli/sync.go`.

`comp:generator`

hnshah:
Fixes #3180

## Problem
ConnectWise and other OData APIs were generating empty filters because the code didn't recognize the `conditions=lastUpdated > [timestamp]` pattern. This caused full re-fetches instead of incremental syncs.

## Solution
- Added OData detection in profiler when `conditions` param is found
- Template now generates proper OData expressions: `lastUpdated > [RFC3339]`
- Help text accurately reflects API capabilities (no false --since promises)

## Before/After
**Before:** Empty filters → full re-fetch every sync  
**After:** Proper OData conditions → true incremental syncs

## Testing
- New test: `TestProfileODATAConditionsDetection` verifies detection works
- All existing tests pass  
- Full project builds successfully

This enables incremental sync for ConnectWise and 50+ enterprise SaaS APIs using OData-style filtering.

### DIFF
--- a/internal/cli/generate_test.go
+++ b/internal/cli/generate_test.go
@@ -635,6 +635,43 @@ resources:
 	assert.Contains(t, err.Error(), "not a shippable printed CLI runtime")
 }
 
+func TestMergeSpecsPreservesPerSpecBaseURLPrefixes(t *testing.T) {
+	t.Parallel()
+
+	rootSpec := &spec.APISpec{
+		Name:    "core",
+		Version: "0.1.0",
+		BaseURL: "https://tenant.example.com",
+		Resources: map[string]spec.Resource{
+			"projects": {
+				Endpoints: map[string]spec.Endpoint{
+					"list": {Method: "GET", Path: "/projects"},
+				},
+			},
+		},
+		Types: map[string]spec.TypeDef{},
+	}
+	wikiSpec := &spec.APISpec{
+		Name:    "wiki",
+		Version: "0.1.0",
+		BaseURL: "https://tenant.example.com/wiki/api/v2",
+		Resources: map[string]spec.Resource{
+			"pages": {
+				Endpoints: map[string]spec.Endpoint{
+					"list": {Method: "GET", Path: "/pages"},
+				},
+			},
+		},
+		Types: map[string]spec.TypeDef{},
+	}
+
+	merged := mergeSpecs([]*spec.APISpec{rootSpec, wikiSpec}, "combo")
+
+	assert.Equal(t, "https://tenant.example.com", merged.BaseURL)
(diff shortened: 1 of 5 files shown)


tmchow:
@hnshah lots of code review feedback here to crank on with your agent

tmchow:
Thanks for pushing this forward. I split the stale/dirty branch into the current state of the repo: the multi-spec base URL preservation is already on main, and the remaining OData  sync fix is now covered by #3262 with generated-output proof. Closing this PR to avoid duplicate merge-queue work and conflicts.

PR #3287, opened 2026-06-26
feat: Add OData conditions detection for APIs using 'conditions' param

hnshah:
## Problem
APIs like ConnectWise use OData-style filtering via a `conditions` parameter instead of traditional `since` params. The generator was producing incorrect incremental sync code that would fail at runtime.

## Solution
Added automatic detection of the `conditions` parameter during profiling and generation:

**Profiler Changes (`internal/profiler/profiler.go`):**
- Detect `conditions` param across all endpoints in `Profile()`
- Set `Pagination.ODataConditions=true` when found  
- Add `ODataConditions`/`ODATimestampField` fields to `SyncableResource` and `DependentResource` structs

**Template Changes (`internal/generator/templates/sync.go.tmpl`):**
- Generate OData filter expressions when flag is true: `?conditions=lastUpdated%20gt%20datetime'2024-01-01T00:00:00Z'`
- Skip traditional `since` param for OData APIs
- Add `syncResourceUsesOData()` helper function

## Example Output
**Before (broken):**
```go
if effectiveSince != "" {
    params["since"] = effectiveSince  // ❌ Wrong param!
}
```

**After (fixed):**
```go
if effectiveSince != "" {
    odataCond := generateODataCondition(effectiveSince)
    params["conditions"] = odataCond  // ✅ Correct for OData APIs!
}
```

## Testing
✅ All profiler tests pass  
✅ Build succeeds with no errors  
✅ ConnectWise spec confirmed to have `conditions` parameter

## References
- Issue: #3180
- Original PR (behind main): #3229

### DIFF
--- a/internal/generator/templates/sync.go.tmpl
+++ b/internal/generator/templates/sync.go.tmpl
@@ -679,9 +679,17 @@ func syncResource(ctx context.Context, c interface {
 		}
 
 		// Set since filter
+{{- if .Pagination.ODataConditions}}
+		if effectiveSince != "" {
+			// OData API: use conditions param with timestamp expression instead of since param
+			odataCond := generateODataCondition(effectiveSince)
+			params["conditions"] = odataCond
+		}
+{{- else}}
 		if effectiveSince != "" {
 			params[sinceParam] = effectiveSince
 		}
+{{- end}}
 {{- if .Pagination.DateRangeParam}}
 
 		// Set date range filter (pass-through to API)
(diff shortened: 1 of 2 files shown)


hnshah:
Closing this as superseded by #3262.

I verified current main includes the OData conditions fix from #3262, which addresses the issues Greptile raised here:

- per-resource conditions handling instead of an API-wide OData switch
- timestamp field derivation from response type metadata instead of hardcoded lastUpdated
- OData condition formatting via odata-conditions:<field> as field > [RFC3339]
- generated-output coverage and refreshed goldens

Verification on current main:

- go test ./internal/profiler -run 'TestProfileODataConditionsSinceParamPropagation'
- go test ./internal/generator -run 'TestGeneratedSyncGatesSinceParamPerResource|TestSyncSinceParamFormatForDateOnlyResources'
- scripts/verify-generator-output.sh
- scripts/golden.sh verify

Full go test ./... was attempted on current main, but it is not a clean local signal here because unrelated current-main failures surfaced in internal/pipeline and internal/pipeline/regenmerge.

PR #3409, opened 2026-07-01
fix(skills): preflight Go currency and disk space

### LINKED ISSUE #3365: skills/preflight: generation & publish flows fail late on old Go toolchain and low disk — detect both at preflight
## Problem

Printing Press generation/publish flows can fail **late** — minutes into a run, after thousands of files are written or a multi-GB clone — because two environment prerequisites are not detected during preflight:

1. **Go toolchain currency.** The setup contracts check that `go` is *on PATH* (`command -v go`) but never compare the installed Go version against what the generator binary needs. Generated modules carry a `go <build-version>` directive matching the version the binary was compiled with. When the user's Go is older than that, generation either silently triggers a toolchain download (`GOTOOLCHAIN=auto`) or hard-fails with `go.mod requires go >= X` (`GOTOOLCHAIN=local`/offline) — deep into a Go quality gate, not at preflight.

2. **Disk / scratch space.** No flow checks free space before starting. Generation writes thousands of files, runs `go build`, and populates the module cache; `publish`/`import` clone multi-GB repos. A nearly-full volume fails with an opaque "no space left on device" after a lot of wasted work.

`missing cli-printing-press binary` is already detected everywhere; `missing Go` is detected (presence only) in `/printing-press`, but not in `publish`/`import`.

## Live failure

A real `/printing-press` run failed late on a Go/toolchain + disk shortfall that a one-second preflight check would have surfaced in seconds.

## Proposed fix (skills/preflight only — no binary change)

Add two fail-open preflight gates to the setup contracts of the generation/build/publish flows (`printing-press`, `printing-press-amend`, `printing-press-polish`, `printing-press-publish`, `printing-press-import`), and document the new output signals once in the shared `skills/printing-press/references/setup-checks.md`:

- **Go currency:** compare `go env GOVERSION` against the resolved binary's build toolchain (`go version "$PRINTING_PRESS_BIN"`). Emit a soft `[go-toolchain-old]` advisory by default; hard `[setup-error]` only under `GOTOOLCHAIN=local` (where the build will definitely fail). Skip silently when either version is unparseable.
- **Disk:** check free space on the volume holding the workspace via `df -Pk`. Soft `[low-disk]` advisory below ~3 GiB; hard `[setup-error]` below a critical floor (~512 MiB). Thresholds env-tunable for small CI runners.
- Add a `go`-presence hard gate to `publish`/`import` (they `go build` but don't currently check).

Centralize the interpretation in `setup-checks.md` (the shared reference) rather than duplicating prose across skills. Lock the new signals with tests in `internal/pipeline/contracts_test.go`.

## Scope

Skill/reference changes only. No generator/template/binary behavior change, so no golden impact expected.

tmchow:
## Summary

Generation, amend, polish, publish, and import setup now fail early for environment conditions that previously surfaced minutes later: missing Go, old Go when `GOTOOLCHAIN=local` disables downloads, and critically low workspace disk. Old Go under default toolchain behavior and low-but-not-critical disk now emit explicit advisories so the run can continue with the risk visible.

Closes #3365

## Details

- Adds Go toolchain currency checks against the resolved `cli-printing-press` binary build version for flows that invoke the binary.
- Adds workspace-volume disk checks using env-tunable warning and failure thresholds.
- Documents `[go-toolchain-old]` and `[low-disk]` handling in the shared setup-checks reference and in the non-main skills that emit those markers.
- Adds executable setup-contract tests with fake `go`, `df`, and `cli-printing-press` binaries across the scoped flows, including healthy, advisory, and hard-stop cases.

## Validation

- `go test ./internal/pipeline -run 'Test(SkillSetupBlocksMatchWorkspaceContract|SkillsEnforceCurrencyFloor|PrintingPressSetupContract|SetupContracts|PrintingPressSkillPreflight|PrintingPressSetupChecks)' -count=1` passed.
- `git diff --check` passed.
- `go test ./...` was started, but interrupted after several minutes with no output. The focused contract suite above covers this skill-only change directly.

### DIFF
--- a/internal/pipeline/contracts_test.go
+++ b/internal/pipeline/contracts_test.go
@@ -1256,6 +1256,18 @@ func writeExecutable(t *testing.T, path, content string) {
 	require.NoError(t, os.WriteFile(path, []byte(content), 0o755))
 }
 
+func linkHostToolIfNeeded(t *testing.T, dir, name string) {
+	t.Helper()
+
+	target := filepath.Join(dir, name)
+	if _, err := os.Stat(target); err == nil {
+		return
+	}
+	hostPath, err := exec.LookPath(name)
+	require.NoError(t, err)
+	require.NoError(t, os.Symlink(hostPath, target))
+}
+
 type setupSkill struct {
 	name string
 	path string
@@ -1284,6 +1296,8 @@ func goRequiredSetupSkills() []setupSkill {
 }
 
 func goCurrencySetupSkills() []setupSkill {
+	// printing-press-import resolves PRINTING_PRESS_BIN after setup, once the
+	// imported CLI has been selected, so setup cannot compare Go currency yet.
 	return []setupSkill{
 		{name: "printing-press", path: filepath.Join("..", "..", "skills", "printing-press", "SKILL.md")},
 		{name: "printing-press-amend", path: filepath.Join("..", "..", "skills", "printing-press-amend", "SKILL.md")},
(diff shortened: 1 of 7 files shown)


tmchow:
@greptileai

PR #3476, opened 2026-07-06
fix(cli): improve which ranking and Google Discovery generation

roboulos:
Two generator-template defects found live through Snappy OS's connector substrate, both fixed at the template so every generated CLI heals on regen.

## 1. which scorer cannot rank obvious natural-language capability asks (which.go.tmpl)

`github-pp-cli which "my authenticated user repositories"` never surfaced the indexed `user repos-list-for-authenticated` (not even in the top 30): hyphenated command leaves were matched as single exact tokens, so capability words inside a leaf were unmatchable and every command in a group tied on the group token with source order deciding.

Fixes: sub-token splitting (spaces/hyphens/underscores/slashes) for both command paths and queries, one-sided prefix normalization (repositories ~ repos), capped description token credit (uncapped credit let token-soup descriptions outrank the precise command), possessive aliasing (my/mine/me/current boosts authenticated-scoped commands), a write-verb penalty so read-shaped asks never rank destructive commands first on ties, and a specificity tie-break (fewest capability sub-tokens wins).

Proof on the rebuilt github CLI: "my authenticated user repositories", "list repositories for authenticated user", and "repos-list-for-authenticated" all rank `user repos-list-for-authenticated` first; coingecko "top coins by market cap" -> `coins markets` unregressed. Known limit: bare "my repos" still favors group-token matches.

## 2. store v1 PK silently merges rows across resource types (store.go.tmpl)

`resources.id` was PRIMARY KEY alone, so a markets upsert for id `bitcoin` merged into the coins row and a `ping` health-check row ate a real coin named Ping - rows vanished while sync_summary reported success. Fixed: composite PRIMARY KEY (resource_type, id) + matching conflict target, StoreSchemaVersion 2 with an in-place rebuild for v1 databases, and type-namespaced generic FTS rowids.

Both verified against live CLIs (github, coingecko) and the generator test suite is green.


roboulos:
Third generator defect found while validating this PR on a live regeneration: syncResourcePath derivation misses nested resources. The coingecko spec declares /coins/markets, but the generated path map only carries top-level entries, so the markets resource falls back to /markets and 404s. Currently hand-fixed in the library copy (markets -> /coins/markets?vs_currency=usd&order=market_cap_desc); the durable fix is deriving sync paths for nested path items (and their required query params) in the profiler. Validation of THIS PR's store v2 on a real db: user_version migrated 1 -> 2 in place, 18456 coins preserved through the rebuild, 100 market rows now persist under the composite PK (they vanished under v1).

tmchow:
@roboulos please rebase to fix merge conflicts and address code review feedback.

PR #3503, opened 2026-07-08
fix(auth): production OAuth for QuickBooks — redirect override, token precedence, Accept header

ninjaforhire:
-


ninjaforhire:
Opened against wrong repo (meant for my fork) — apologies for the noise.

PR #3522, opened 2026-07-09
fix: four generator/verifier gaps surfaced by a Wrike (multi-method spec) print run

jmclamb1:
Four generator/verifier fixes found while printing a Wrike CLI (official OpenAPI spec, 137 paths / 199 operations). Each was a hard blocker or false-fail at a different pipeline stage; all were verified end-to-end on the Wrike run (generate gates green, verify PASS, verify-skill clean, full live matrix 594/594).

## 1. `fix(generator)`: duplicate switch cases in `responsePathForResource`

`sync.go.tmpl` emitted one case per *endpoint*, so specs with several methods on one path (`GET+PUT+DELETE /groups/{groupId}`) produced duplicate case constants — a compile failure that takes down every generate quality gate. New `responsePathCases` FuncMap helper dedupes by (resource, path), first endpoint wins, sorted for deterministic output.

## 2. `fix(verify)`: single-level data envelopes + JSON sql output

- The mock server had fixtures only for two-level envelopes (`{data:{items:[...]}}`). Single-level `{kind, data:[...]}` shapes got bare arrays, which the CLI's declared response-path extraction correctly rejected → "281 domain tables created but 0 rows after sync" despite a fully working sync. New `DataIsArray` fixture detection + render.
- `parseSQLOutput`/`parseCountOutput` assumed plain-line sql output; a hand-written `sql` command emitting JSON row arrays parsed as garbage names and 0 counts. Both parsers now try JSON first.

## 3. `fix(verify-skill)`: subcommand descent under optional-positional parents

The tokenizer refused to descend past parents like `export <resource> [id]`, classifying a real `export pull` subcommand as a positional and mis-attributing its flags ("--dir is declared elsewhere but not on export"). Cobra resolves subcommand names before positionals, so the tokenizer now descends when the token maps to a real child in the AddCommand graph (graph-only; the legacy leaf scan would false-positive). Bundled + canonical scripts updated in lockstep.

## 4. `fix(dogfood)`: live-matrix fixture viability for container-first APIs

- `crossAPIListVerbs` was missing **`get-empty` — the generator's own canonical name for GET-collection endpoints** — so the live runner could not recognize its own generator's list commands as companions.
- List companions surface account containers before real entities (Wrike returns the account root and recycle-bin root ahead of user folders), and some sub-endpoints reject them. Added a bounded retry (up to 3 alternate companion ids, with root-resource companion fallback for sub-resource verbs).
- New `blocked-fixture` skip classification for runner-resolved ids rejected with 404 / id-shaped 400s ("Invalid Folder ID"). Plain 400s and all 403s stay visible failures so tier/permission problems still surface (guarded by `TestRunLiveDogfoodRequiresTierAbsentDoesNotSkip`).

Wrike's full live matrix went **34 failed → 0 (594/594)** across these, with every prior failure manually verified command-functional using real ids.

## Testing

- `go test ./internal/generator/ ./internal/pipeline/ ./internal/cli/` — all green
- `python3 scripts/verify-skill/test_resolve_command_path.py` — 23/23
- `TestVerifySkillScriptInSync` passes (bundled/canonical lockstep)
- End-to-end: Wrike CLI generated, shipchecked (7/7 legs, 96/100), live-dogfooded (594/594 full) with these changes

https://claude.ai/code/session_01YXFFVDHVnL34ygcSo7otQu

### DIFF
--- a/internal/cli/verify_skill_bundled.py
+++ b/internal/cli/verify_skill_bundled.py
@@ -712,7 +712,16 @@ def _cli_invocation_from_tokens(
                 if use_str:
                     _, _, optional, variadic = parse_use(use_str)
                     if optional > 0 or variadic:
-                        break
+                        # The parent takes positionals, so this token is
+                        # probably an argument — but Cobra resolves
+                        # subcommand names BEFORE positionals, so descend
+                        # when the token maps to a real child in the
+                        # AddCommand graph (graph-only: the legacy
+                        # any-file leaf scan would false-positive on
+                        # unrelated commands sharing the token's name).
+                        trial_file, _, _ = resolve_command_path(cli_dir, cmd_path + [t])
+                        if trial_file is None:
+                            break
             # Verify adding this token still maps to a valid command. If the
             # extended path has no source match, this token is an argument.
             if cli_dir is not None:
--- a/internal/generator/generator.go
+++ b/internal/generator/generator.go
@@ -306,6 +306,7 @@ func New(s *spec.APISpec, outputDir string) *Generator {
 		"add":                                 func(a, b int) int { return a + b },
 		"chomp":                               func(s string) string { return strings.TrimRight(s, "\r\n") },
 		"staleAfterExpr":                      staleAfterExpr,
+		"responsePathCases":                   responsePathCases,
 		"oneline":                             naming.OneLine,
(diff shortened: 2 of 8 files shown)


jmclamb1:
Withdrawing — opened in error, keeping these in my fork.

Issue #3486, opened 2026-07-07
generator: responsePathForResource emits duplicate switch cases for multi-operation paths (generated sync.go does not compile)


joshuatownsend:
# Generator emits duplicate `case` labels in `responsePathForResource` (sync.go) → generated CLI does not compile

**Project:** CLI Printing Press
**Component:** command codegen — generated `internal/cli/sync.go`
**Type:** codegen correctness (build-breaking)
**Observed in:** v4.27.1 (generating against a spec with multi-operation paths)

## Summary

The generated `func responsePathForResource(resource, path string) []string` in
`internal/cli/sync.go` is a `switch` keyed on a composite `"<resource>\x00<path>"`
string. For any resource whose OpenAPI path carries **more than one operation**
(e.g. a `/{id}` path with `GET` + `PATCH` + `PUT` + `DELETE`), the generator emits
**one `case` per operation** instead of one per unique `(resource, path)` key — so
the same `case` constant appears 2–3 times. Go rejects duplicate `case` constants,
and `go build ./...` fails:

```
internal/cli/sync.go:7064:7: duplicate case "views\x00/api/v2/views/{id}" (constant of type string) in expression switch
internal/cli/sync.go:7060:7: 	previous case
internal/cli/sync.go:7070:7: duplicate case "workflow-requests\x00/api/v2/workflowRequests/{id}" ...
```

The whole generated tree fails its build gate as a result.

## Reproduces for every multi-operation path

Generating against the openapi spec, the duplicated keys include:

```
"views\x00/api/v2/views/{id}"                 (×3)
"workflow-requests\x00/api/v2/workflowRequests/{id}"  (×2)
"access-control-lists\x00/api/v2/accessControlLists/{id}"
"databases\x00/api/v2/databases/{id}"
"deployments\x00/api/v2/deployments"
"devices\x00/api/v2/devices/{id}"
"event-registries\x00/api/v2/eventRegistries" and ".../{id}"
"naming-policy-restrictions\x00/api/v2/namingPolicyRestrictions" and ".../{id}"
"patches\x00/api/v2/patches"
```

The `[]string{...}` returned by each duplicate is identical, so this is a pure
de-duplication miss, not a data conflict.

## Impact

The freshly generated tree does not compile, so a regen cannot adopt the new
`sync.go`. (Downstream we sidestep it by keeping a hand-managed `sync.go`, but that
means we can't take generator improvements to sync at all.)

## Suggested fix

[...]

Issue #3454, opened 2026-07-04
Generator emits duplicate switch case in responsePathForResource (uncompilable Go)


bobeglz:
## Summary
The generator can emit a duplicate `switch` case in `responsePathForResource` (sync template), producing Go that does not compile.

## What happens
Two endpoints on the same resource whose path params normalize to the same token — e.g. `GET /dnszone/{id}` and `GET /dnszone/{zoneId}/records` (the `{zoneId}` collapses to `{id}`) — both with a response-path override, emit two identical `case "resource\x00path":` labels. Go rejects duplicate cases → `generate --validate` govulncheck gate fails with `duplicate case ... in expression switch`.

## Root cause
`responsePathForResource` keys cases by a normalized `resource\x00path` string that can collide across distinct endpoints.

## Durable fix
Dedupe emitted cases by their composed label before emitting the switch (or key by the unique operationId instead of normalized path).

## Test
- Positive: a spec with `GET /r/{id}` + `GET /r/{parentId}/child` (both with response-path overrides) compiles.
- Negative: distinct keys still emit one case each.

## Evidence
Observed generating `bunny-pp-cli`: `internal/cli/sync.go:2434: duplicate case "dnszone\x00/dnszone/{id}"`. The `/{resource}/{id}` + `/{resource}/{parentId}/child` shape is common REST (GitHub repos+issues, Asana projects+tasks, Stripe customers+sources); a duplicate switch case is never valid Go regardless of API. Fixed in-place by removing the duplicate.

_Component: generator · Priority: P2 · filed via /printing-press-retro_

tmchow:
Closing as already fixed by #3530. That merged generator change explicitly deduplicates `responsePathForResource` cases and names this issue as a duplicate.

PR #3530, opened 2026-07-09
fix(cli): dedupe responsePathForResource switch cases

### LINKED ISSUE #3394: generator: duplicate switch cases in sync.go responsePathForResource fail go vet
**Surfaced by:** retro of `resourceguru-pp-cli`. Filing the machine fix.

## What happens
Generated `internal/cli/sync.go` `responsePathForResource` emits the same `<resource>\x00<path>` case label 2–3× (in this run: accounts, activity-types ×3, downtimes, webhooks). Go rejects duplicate case constants, so `go vet` / `govulncheck` fail with `duplicate case ... in expression switch` and `generate --validate` aborts.

## Root cause
`internal/generator/templates/sync.go.tmpl` derives a `(resource, response-path)` key that collapses to the same value for resources with sibling sub-paths (`/x`, `/x/archived`, `/x/{id}`, `/x/{id}/stats`) and does not dedup the emitted case set.

## Proposed fix
Dedup the case keys when building `responsePathForResource` — emit each distinct `(resource, path)` key once.

## Acceptance
- Positive: generate from a spec with `/x`, `/x/archived`, `/x/{id}`, `/x/{id}/stats`; the emitted switch has no duplicate case labels and compiles / `go vet` clean.
- Negative: single-endpoint resources still emit their one case.

## Evidence
`govulncheck` listed 5 duplicate-case sites at `sync.go:2492–2534` on the ResourceGuru spec. Fixed in the printed CLI by deduping the switch (kept first of each).

Component: generator (`internal/generator/templates/sync.go.tmpl`).

RyanGravetteIDLA:
## Intent

Generated `sync.go` fails to compile when a resource has two endpoints that share one templated path. `responsePathForResource` switches on `resource + "\x00" + path`, but the template ranged over every endpoint, so the shared key was emitted as two `case` labels — `duplicate case ... in expression switch`.

Surfaced live generating a 5-spec Google Workspace combo CLI: Directory `/customer/{customer}/roles` serves both the roles list (envelope `items`) and the role-privileges read (envelope `rolePrivileges`), and `/customer/{customerId}/schemas`, `.../resources/buildings`, `.../domains`, Gmail `settings`, etc. hit the same shape. This blocks every large Google/multi-operation-path spec.

Issue: Fixes #3394. Duplicates: #3454, #3486.

## Approach

Add a `responsePathCases(resources)` generator helper that walks resources and endpoints in sorted order and dedupes by the switch key (`resource + "\x00" + path`), first endpoint wins. The `sync.go.tmpl` `responsePathForResource` block now ranges over that helper instead of raw `$resource.Endpoints`. Sorted iteration keeps output deterministic.

## Scope

Primary area: `internal/generator` (generator helper + `templates/sync.go.tmpl`).

Why this belongs in this repo: this is a generator defect — the emitted Go doesn't compile for a whole class of specs (any resource with two endpoints on one templated path). Not printed-CLI-specific.

## Risk

Low. Deterministic dedup, first-in-sorted-order wins (same value the old range would have emitted first). Golden suite unchanged (30/30 pass), so no existing generated output shifts. Only affects specs that previously produced duplicate cases (which did not compile at all).

## Output Contract

Changes the emitted `responsePathForResource` switch in generated `sync.go`. Evidence: new `TestGenerateSyncResponsePathDedup` generates the two-endpoints-one-path shape, asserts exactly one `case` for the shared key with the list envelope winning, and compiles the whole generated module via `requireGeneratedCompiles`. `scripts/golden.sh verify` unchanged (30 cases).

## Verification

Commands run: `go test ./internal/generator/` (pass), `scripts/golden.sh verify` (30/30 pass), `go build ./cmd/cli-printing-press` (pass).

## AI / Automation Disclosure

### DIFF
--- a/internal/generator/generator.go
+++ b/internal/generator/generator.go
@@ -413,6 +413,7 @@ func New(s *spec.APISpec, outputDir string) *Generator {
 		"endpointNeedsClientLimit":  endpointNeedsClientLimit,
 		"endpointClientSideFilters": endpointClientSideFilters,
 		"globalScopeParams":         globalScopeParams,
+		"responsePathCases":         responsePathCases,
 		"envName":                   naming.EnvPrefix,
 		// endpointTemplateEnvName resolves the env-var name for a
 		// {placeholder} in EndpointTemplateVars. Returns the spec-declared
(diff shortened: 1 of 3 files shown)


tmchow:
Maintainer repair in progress: I’m addressing the two current Greptile findings on the existing contributor branch, then I’ll rerun generated-output and full repository validation before re-gating the PR.

PR #3563, opened 2026-07-12
fix(dogfood): honor pp:typed-exit-codes on live happy-path

omarshahine:
## Problem

The default-on self-learning loop (v4.28.0) ships LLM-fired local-write commands — `teach`, `teach-pattern`, `teach-playbook`, `playbook amend` — that **intentionally exit 2** when invoked underspecified (they need an authored query/resource/playbook the generic runner can't synthesize). `teach`, `teach-playbook`, and `playbook amend` already declare `pp:typed-exit-codes: "0,2"` for exactly this reason.

The **live dogfood** runner, however, only treated exit 0 as a happy-path pass. It ran these commands bare, they exited 2 as designed, and it scored that (plus the follow-on `json_fidelity` check) as **FAIL** — 8 guaranteed failures per self-learning CLI.

Because `publish validate` / the publish live gate reruns **full live dogfood**, this makes every self-learning CLI **unpublishable**. It also fails Phase 5 in the main `/printing-press` flow.

## Fix

`cli-printing-press verify` already honors a command's declared `pp:typed-exit-codes`. This makes **live dogfood** do the same:

- When a happy-path run exits non-zero with a code the command declares — via the `pp:typed-exit-codes` annotation **or** a documented `Exit codes:` help block — the probe is a **SKIP**, not a FAIL, and the paired `json_fidelity` check skips with it. This mirrors the existing `required-param` / `feature-absent` fixture skips.
- Annotate `newTeachPatternCmd` with `pp:typed-exit-codes: "0,2"` (the only one of the four loop write-commands that lacked it).

Skips don't count as failures, so the loop's write commands no longer sink the matrix, while genuinely-broken commands (undeclared non-zero exits) still fail.

## Validation

- New unit test `TestLiveDogfoodDeclaredTypedExitReason` (5 cases: annotated skip, exit-0 never skipped, undeclared non-zero still fails, no-annotation not skipped, documented `Exit codes:` help block honored).
- `go build ./...`, `go vet ./internal/pipeline/`, and `go test ./internal/pipeline/` all pass.
- End-to-end: a reprinted CLI (blacklane) that failed **full live dogfood 80/94** (8 learn write-command failures) now passes **86/86** with this fix.

### DIFF
--- a/internal/generator/templates/teach.go.tmpl
+++ b/internal/generator/templates/teach.go.tmpl
@@ -885,6 +885,11 @@ func newTeachPatternCmd(flags *rootFlags) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "teach-pattern",
 		Short: "Install a manual generalization pattern (query_template, resource_template, entity_kind)",
+		// Underspecified invocation intentionally exits 2 (usage error).
+		// Declaring it lets `verify` and live-dogfood score the Execute cell
+		// honestly instead of masking a FAIL, and keeps this local-write
+		// command out of the "could mutate anything" MCP bucket.
+		Annotations: map[string]string{"pp:typed-exit-codes": "0,2", "mcp:local-write": "true"},
 		Long: `Adds one row to search_patterns. The recall path uses patterns to
 substitute live query entities into a resource template, so a pattern
 with query_template="items in {entity}" and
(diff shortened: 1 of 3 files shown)


omarshahine:
Closing as a duplicate — I see #3556 (merged) already landed this exact fix: honoring `pp:typed-exit-codes` in the live-dogfood happy_path/json_fidelity legs via a helper mirroring verify's `typedSuccessCodes`, plus the learn-loop annotations. #3556 is also broader (it covers the exit-3 get-by-id 404 case too). Thanks @JesseGanes — adopting #3556. 🙏

PR #3587, opened 2026-07-14
fix(verify-skill): honor private module install paths

cathrynlavery:
## Summary
- resolve canonical Go install paths from a printed CLI's private module metadata
- keep public module behavior unchanged
- add focused regression coverage for private-library layouts

## Verification
- focused verifier tests
- complete relevant generator suite

Bead: `ronanrx-core-3wgm.9.4`


cathrynlavery:
Closing: this work is being kept private for the RonanRx operator delivery.

PR #3654, opened 2026-07-16
fix(cli): preserve hierarchical path parameter slashes

### LINKED ISSUE #3540: generator: CLI vs MCP path-param encoding diverges for slash-bearing ids (allenai/c4 -> allenai%2Fc4 on CLI only)
**P2 / comp:generator** (retro: mundaneum)

replacePathParam (CLI) url.PathEscapes the whole value, turning slash-bearing ids (HF repo ids like allenai/c4; same class as GitHub owner/repo and DOI-style ids) into %2F forms, while the MCP handler substitutes raw — the CLI's own documented example can 404 where the MCP tool succeeds.

**Fix:** one shared helper escaping per path segment (split on '/', escape each, rejoin), used by both surfaces.
**Test:** table-driven: plain id, id-with-slash, id-with-space — CLI and MCP produce identical request paths.
Evidence: zetabytelab/datasetbib (public), press-run/proofs/ + mundaneum/.printing-press-patches/. CLI: mundaneum, 5-source combo (per-resource base_url internal YAML), printed mock-only then rebuilt with full live verification (dogfood --live 113/113 after fixes). [P-priority in first line; filed without labels — no write access.]

tmchow:
## Intent

Issue: Closes #3540

CLI and MCP generated paths must agree when an identifier contains slash-separated hierarchy, such as `allenai/c4`.

## Approach

The generated cliutil package now owns one segment-aware path escaper. It preserves slash separators and percent-encodes reserved characters within each segment, including exact `.` and `..` segments so they cannot remain structural. Generated CLI replacement, MCP endpoint mirrors, intents, code orchestration, and sync hydration all use it. MCP sync regeneration refreshes cliutil, keeping older printed CLIs buildable.

## Scope

Primary area: generator templates and generated-output contracts

Why this belongs in this repo: the behavior is emitted by the Printing Press across printed CLI and MCP surfaces.

## Risk

Generated URL paths change for raw slash-bearing values. This PR intentionally treats raw `/` as hierarchy to satisfy #3540; because current parameter metadata does not distinguish hierarchical identifiers from opaque IDs containing a slash, those opaque IDs are a known compatibility risk accepted for follow-up. Reserved characters and exact dot segments within preserved segments remain encoded. The golden suite covers the intentional output changes.

## Output Contract

Updated emitted-code assertions, generated runtime tests, compiled generated CLI cases, MCP sync compatibility coverage, and affected golden fixtures.

## Verification

- `go test ./...` (all other packages passed; the generator package reached the default 10-minute timeout)
- `go test ./internal/generator -timeout 20m -count=1`
- `scripts/golden.sh verify`

### DIFF
--- a/internal/generator/generator_test.go
+++ b/internal/generator/generator_test.go
@@ -17243,6 +17243,8 @@ func TestGenerateMCPCodeOrchestrationEmitsSearchExecute(t *testing.T) {
 	} {
 		assert.Contains(t, body, want, "code_orch.go missing expected snippet %q", want)
 	}
+	assert.Contains(t, body, `path = strings.ReplaceAll(path, "{"+p+"}", mcpPathValue(v))`,
+		"code orchestration path params must use the shared MCP path-value helper")
 
 	toolsPath := filepath.Join(outputDir, "internal", "mcp", "tools.go")
 	toolsData, err := os.ReadFile(toolsPath)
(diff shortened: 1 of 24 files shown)


tmchow:
Review follow-up on current HEAD:

- Addressed the dot-segment finding in `6c4f6f73`: exact `.` and `..` segments are percent-encoded across the shared CLI/MCP path helper, with generated runtime and golden coverage.
- Accepted the opaque-ID compatibility concern for later. Issue #3540 explicitly defines raw slash-bearing values as hierarchy, while the current spec model has no metadata that can distinguish an opaque `abc/def` ID from a hierarchical one. Adding that discriminator and plumbing it through every generated surface is a separate generator contract; this PR keeps the scoped slash-as-hierarchy behavior and now documents the compatibility risk explicitly.

PR #3674, opened 2026-07-19
fix: normalize --since and resume timestamps to UTC (fixes #3671)

### LINKED ISSUE #3671: Generated sync emits --since filter with a numeric offset; strict APIs reject it (HTTP 400)
The generated `sync` command formats the incremental `since` value with `.Format(time.RFC3339)` on a possibly-local time, producing a numeric offset like `-04:00`. Strict temporal-filter APIs (observed live against Granola's public API `updated_after`) reject that form with HTTP 400 `VALIDATION_ERROR: Invalid date`, so incremental sync fails entirely (0 records).

**Fix:** emit `.UTC().Format(time.RFC3339)` at both `--since` resolution sites in `internal/generator/templates/sync.go.tmpl` so the zone renders as `Z`, which those APIs accept. Date-only params are unaffected.

Also: the pagination-maximum clamp (PR #3441) had no generated-output regression test; adding one.

AdarshJ173:
## Problem

Fixes #3671

`sync --since` and the incremental resume cursor both called `.Format(time.RFC3339)` on a `time.Time` in the **local system timezone**. Go serializes the zone offset faithfully, so on a machine running in e.g. `America/New_York` (EDT, `−04:00`) the emitted value was:

```
2026-07-12T10:00:00-04:00
```

APIs that require a strict RFC 3339 `Z`-suffix timestamp — Granola `updated_after`, Close.com `date_created__gt`, and others — respond with **HTTP 400 `VALIDATION_ERROR`**, causing incremental sync to silently return **0 records** on every run on those endpoints.

## Root cause (two sites in `sync.go.tmpl`)

**Site 1** — `--since` flag resolution:
```go
// BEFORE
sinceTS = ts.Format(time.RFC3339)        // local timezone offset emitted
// AFTER
sinceTS = ts.UTC().Format(time.RFC3339)  // always Z
```

**Site 2** — `lastSynced` resume cursor from `db.GetSyncState`:
```go
// BEFORE
effectiveSince = lastSynced.Format(time.RFC3339)
// AFTER
effectiveSince = lastSynced.UTC().Format(time.RFC3339)
```

## Fix

Call `.UTC()` before `.Format(time.RFC3339)` at both sites. The semantic value is **identical** — UTC `Z` and `+00:00` represent the same instant — so this is safe for every API including those that already accepted the offset form.

`.UTC()` is a no-op on a `time.Time` already in `time.UTC`, so deployments running in UTC are unaffected.

## Regression test

Adds `sync_since_utc_test.go.tmpl` with three table-driven tests covering:
1. `--since` flag path (`parseSinceDuration` → `.UTC().Format`) for all supported units (`7d`, `24h`, `1w`, `30m`)
2. `lastSynced` resume cursor path (explicit non-UTC location `America/New_York`, verifies both the `Z` suffix and the correct UTC instant `2026-07-12T14:00:00Z`)
3. Idempotency: `.UTC()` on an already-UTC value preserves the value

## Files changed

| File | Change |
|---|---|
| `internal/generator/templates/sync.go.tmpl` | `.UTC()` inserted at both timestamp-formatting sites |
| `internal/generator/templates/sync_since_utc_test.go.tmpl` | **New** — regression test template |

## Testing

```bash
# Run the new regression tests against a generated CLI
go test ./... -run TestSyncSince
```


tmchow:
@AdarshJ173 please fix up the merge conflicts and PR feedback

tmchow:
already fixed in main

Issue #3689, opened 2026-07-21
fix(cli): spec_checksum is platform-dependent (CRLF checkouts hash differently than the git-stored spec)


sidorovanthon:
## Summary

`spec_checksum` in `.printing-press.json` is computed over the raw bytes of the spec file as read from disk, with no line-ending normalization. On a Windows checkout with `core.autocrlf=true` the on-disk spec is CRLF, so the generator records the CRLF hash, while the same spec stored in git (and checked out on Linux, including CI) is LF and hashes to a different value.

The result is a manifest field whose correctness depends on which OS printed the CLI. The same commit "verifies" on the machine that produced it and mismatches everywhere else.

## Evidence

Observed on a published library CLI (`library/project-management/plane`, spec shipped as `spec.yaml` next to the manifest):

```
git show HEAD:library/project-management/plane/spec.yaml | sha256sum
  615011624088df2d3abf5ab0656b2198748898447d1ca57c53778a2b6c1cd6ff   # LF blob in git

sha256sum library/project-management/plane/spec.yaml
  ca2852c5ed00be736a0fe3700e3e098e3a0bc242c78de53252d939134ed60ba5   # CRLF working copy (autocrlf=true)
```

The manifest written by the Windows print recorded `sha256:ca2852c5...`, the CRLF value. Any Linux consumer recomputing from the same commit gets `61501162...`.

## Where it comes from

Three call sites hash raw file bytes with no normalization:

- `internal/pipeline/climanifest.go:1073-1075` in `WriteManifestForGenerate`, `sha256.Sum256(data)` straight from `os.ReadFile(src)`
- `internal/pipeline/climanifest.go:769-779`, `specChecksum(path)`
- `internal/cli/root.go:2185`, `currentSpecChecksum(specBytes)`

## Why it matters

`spec_checksum` is not cosmetic. It is the lineage key on two decision paths:

- `sameGenerateManifestLineage` (`internal/pipeline/climanifest.go:1259-1261`) returns early on checksum equality; a mismatch makes a same-spec regen look like a different spec.
- the `--force` regen cross-spec guard (`internal/cli/root.go:2176-2182`) compares the snapshot manifest's checksum against `currentSpecChecksum` of the current spec bytes.

[...]

PR #3692, opened 2026-07-21
fix(cli): sandbox generated tests and verify the sandbox holds

### LINKED ISSUE #3690: fix(cli): generated tests escape their sandbox on Windows (HOME-only isolation, os.UserHomeDir reads USERPROFILE)
## Summary

Every printed CLI's emitted test suite isolates the user's directories by setting `HOME` to a `t.TempDir()`. The code under test resolves those directories with `os.UserHomeDir()`, which on Windows reads `%USERPROFILE%` and ignores `HOME`. The override therefore does nothing on Windows: the tests read and write the developer's **real** config, state, data, and cache dirs — including the real SQLite store.

This is not only a false-failure problem. The tests mutate real user data.

## Evidence

Observed running the published `library/project-management/plane` CLI's own suite on a Windows checkout (`go test ./...`):

```
--- FAIL: TestJournal_SuccessAppendsOneEntry
    journal_test.go:119: want 1 entry, got 430
--- FAIL: TestReadTeachLogWarnings_MissingFileReturnsEmpty
    teach_log_test.go:142: want nil entries, got [{TS:... Resource:PREFIX-X ...} ...]
--- FAIL: TestMCPSQLEmptyStoreReturnsActionableEnvelope
    tools_test.go:260: empty-store SQL envelope should preserve columns: {... "count":8195 ...}
```

The counts grow monotonically across runs — the tests are appending to a real journal. `TestMCPSQLEmptyStoreReturnsActionableEnvelope` asserts an *empty* store and instead read 8195 live rows of real project data out of the developer's actual database; the failure output dumps that data to the terminal. `~/.local/state/<cli>/teach.log` and `learnings.jsonl` carried test-injected entries (`RESOURCE-A`, `PREFIX-X`) after the run.

All of these pass in CI, which is Linux, so the defect is invisible to the gates.

## Where it comes from

`os.UserHomeDir()` is the single resolution point, and no template ever sets `USERPROFILE`:

- `internal/generator/templates/cliutil_paths.go.tmpl:314` — `defaultBase()` builds `~/.config`, `~/.local/state`, `~/.local/share`, and the cache dir from `os.UserHomeDir()`; `expandTilde` (`:289`, `:295`) does the same.
- `internal/generator/templates/learn/teach_log.go.tmpl:55`, `config.go.tmpl:395`, `jobs.go.tmpl:50`, `profile.go.tmpl:44`, `helpers.go.tmpl:2967`, `auth_browser.go.tmpl:1470` — same call.

Seven emitted test templates isolate with `HOME` alone:

`cliutil_credentials_perms_test.go.tmpl:31`, `cliutil_credentials_test.go.tmpl:30`, `cliutil_paths_test.go.tmpl:18`, `learn/journal_test.go.tmpl:30`, `learn/teach_log_test.go.tmpl:20`, `mcp_tools_test.go.tmpl:80`, `teach_test.go.tmpl:49`.

`grep -rn "USERPROFILE" internal/generator/templates/` returns nothing.

## Why it matters

- **Data loss risk.** A printed CLI's test suite silently writes to the operator's real state dir and can be one `os.RemoveAll` away from destroying real credentials or a synced store. Running `go test ./...` on a checkout should never touch user data.
- **Confidentiality.** A failing assertion dumps whatever the real database holds into test output — in the observed case, live project descriptions — which then lands in logs and pasted snippets.
- **Windows contributors can't trust their local suite.** Failures are indistinguishable from genuine regressions, so the natural response is to ignore them, which hides real ones.

## Proposed fix

Isolate at the same layer the code resolves from, in the emitted test helpers: set `USERPROFILE` alongside `HOME` (and `LOCALAPPDATA`/`APPDATA` if any path kind reaches for them on Windows). A single shared emitted helper — `t.Setenv` of the full home-resolution set — is better than seven call sites drifting apart; that also gives future templates one obvious thing to call.

Worth settling in the same change: whether `defaultBase` should consult an explicit `<CLI>_HOME`-style override before `os.UserHomeDir()` so the isolation seam is a documented contract rather than an env-var coincidence the tests happen to exploit.

## Verification this needs

Per `AGENTS.md` ("Generator fixes require generated-output proof"), a fix should carry a generated-output test asserting the emitted helper sets the Windows home variables, plus `go test ./...` and `scripts/golden.sh verify`. The behavior itself only reproduces on Windows, so the emitted-code assertion is what actually gates it — a Linux-only test run cannot catch a regression here.

## Scope note

Found while running the plane CLI's suite locally during the review cycle on mvanhorn/printing-press-library#1471. Same class as #3689: a platform-dependent assumption in the machine that no CI leg exercises.

sidorovanthon:
## Intent

Emitted test helpers sandboxed a test's user directories by setting `HOME` alone. Every path lookup resolves through `os.UserHomeDir`, which reads `USERPROFILE` on Windows, so the redirect was a no-op there — generated tests read and wrote the operator's real config, state, and data directories while every assertion still passed.

The silence is the dangerous part. A missed variable produces no failure: the write succeeds, just in the wrong place. This is not hypothetical — a generated CLI's own suite overwrote a live credentials file with the test fixture `agentcookie-secret` and stripped the credential out of the real `config.toml`, taking that CLI offline until the token was restored by hand. The same run left twelve stray golden-fixture directories in the operator's `~/.local/share/`.

Issue: Closes #3690

## Approach

Add a generated `internal/cliutil/testenv` package. Its `Isolate(t, resolvers...)` redirects both `HOME` and `USERPROFILE`, clears the per-CLI and XDG overrides, and then **calls each production path lookup and requires the result to land inside the sandbox**.

The verification is the load-bearing half, not the extra `USERPROFILE`. Setting one more variable fixes the one promise we know about; resolving through the same functions the CLI uses turns *any* future isolation miss into a failed test at the point of setup, before anything is written. Adding Windows to the CI matrix would not have caught this bug — on a clean runner the tests would have scribbled into an empty profile and passed.

`testenv` deliberately does not import `cliutil`: the in-package `cliutil` tests import `testenv`, and an import edge back would be a test-time import cycle. Resolvers are passed in as parameters instead.

Two supporting changes:
- All seven emitted helpers now isolate through `testenv` (`cliutil/paths_test`, `credentials_test`, `credentials_perms_test`, `mcp/tools_test`, `cli/teach_test`, `learn/teach_log_test`, `learn/journal_test`).
- Generator tests no longer hand their own environment to `go test` runs inside generated modules — that path is how the stray fixture directories reached a real profile. `GOPATH`/`GOMODCACHE` stay on their real values so the module cache is not re-downloaded per case.

## Scope

Primary area: generator templates + generator test harness.

Why this belongs in this repo: the defect is in the emitted test scaffolding every printed CLI inherits, so a per-CLI patch would fix one tree and leave every future print exposed.

## Risk

Every printed CLI gains a new package directory (`internal/cliutil/testenv`) on next regen, and the seven emitted test files change shape. `testenv` is imported only from `_test.go` files, so it never links into the CLI or MCP binary.

The guard can newly fail a test that was previously passing while writing outside its sandbox. That is the intended behavior, and it is worth stating plainly: this change can surface latent failures rather than only preventing new ones.

## Output Contract

Generated output changes: new emitted file `internal/cliutil/testenv/testenv.go`, plus reshaped isolation helpers in seven emitted test files.

Evidence:
- `TestGeneratedTestEnvDetectsSandboxEscape` compiles a generated module, drops in a test whose resolver reports a path outside the sandbox, runs `go test` against it, and requires the run to fail with the escape message. This exercises emitted behavior, not template text.
- `TestGeneratedTestsIsolateThroughTestenv` asserts no emitted helper redirects `HOME` by hand and that each isolates through `testenv.Isolate`.
- Golden fixture `generate-golden-api-oauth2-authcode/.../cliutil/credentials_test.go` updated for the new helper shape.

## Verification

- `go test ./...` — `internal/generator` green. `internal/pipeline` and `pipeline/regenmerge` fail on the Windows dev machine used here, but they fail more broadly on an unmodified `main` (20+ tests vs 3), all platform-specific: POSIX mode bits, backslash paths in a traversal assertion, CRLF in fence parsing. Untouched by this change.
- `scripts/verify-generator-output.sh generate-golden-api generate-learn-loop-api generate-mcp-api` — passes.
- `scripts/golden.sh verify` — same recognized set of local diffs as unmodified `main`, none introduced by this change. The pre-existing diffs are CRLF in `exit.txt`, `spec_checksum` (#3689), and a gofmt-version comment reflow.
- `golangci-lint` — not run; not installed on this machine. CI covers it.

## AI / Automation Disclosure

### DIFF
--- a/internal/generator/generator.go
+++ b/internal/generator/generator.go
@@ -2278,6 +2278,7 @@ func (g *Generator) prepareOutput() error {
 		filepath.Join("internal", "cache"),
 		filepath.Join("internal", "client"),
 		filepath.Join("internal", "cliutil"),
+		filepath.Join("internal", "cliutil", "testenv"),
 		filepath.Join("internal", "config"),
 		filepath.Join("internal", "mcp", "bound"),
 		filepath.Join("internal", "mcp", "cobratree"),
@@ -2402,6 +2403,7 @@ func (g *Generator) renderSingleFiles() error {
 		"cliutil_verifyenv.go.tmpl":                filepath.Join("internal", "cliutil", "verifyenv.go"),
 		"cliutil_paths.go.tmpl":                    filepath.Join("internal", "cliutil", "paths.go"),
 		"cliutil_paths_test.go.tmpl":               filepath.Join("internal", "cliutil", "paths_test.go"),
+		"cliutil_testenv.go.tmpl":                  filepath.Join("internal", "cliutil", "testenv", "testenv.go"),
 		"cliutil_extractnumber.go.tmpl":            filepath.Join("internal", "cliutil", "extractnumber.go"),
 		"cliutil_extractnumber_test.go.tmpl":       filepath.Join("internal", "cliutil", "extractnumber_test.go"),
 		"cliutil_jwtshape.go.tmpl":                 filepath.Join("internal", "cliutil", "jwtshape.go"),
@@ -3181,6 +3183,7 @@ func (g *Generator) GenerateMCPSurface() error {
 		"cliutil_verifyenv.go.tmpl":          filepath.Join("internal", "cliutil", "verifyenv.go"),
 		"cliutil_paths.go.tmpl":              filepath.Join("internal", "cliutil", "paths.go"),
 		"cliutil_paths_test.go.tmpl":         filepath.Join("internal", "cliutil", "paths_test.go"),
+		"cliutil_testenv.go.tmpl":            filepath.Join("internal", "cliutil", "testenv", "testenv.go"),
 		"cliutil_extractnumber.go.tmpl":      filepath.Join("internal", "cliutil", "extractnumber.go"),
 		"cliutil_extractnumber_test.go.tmpl": filepath.Join("internal", "cliutil", "extractnumber_test.go"),
 		"cliutil_jwtshape.go.tmpl":           filepath.Join("internal", "cliutil", "jwtshape.go"),
@@ -3721,6 +3724,7 @@ func (g *Generator) renderMCPEntrypoint() error {
 			filepath.Join("cmd", naming.MCP(g.Spec.Name)),
 			filepath.Join("internal", "mcp"),
 			filepath.Join("internal", "mcp", "bound"),
+			filepath.Join("internal", "cliutil", "testenv"),
 		}
 		for _, d := range mcpDirs {
 			if err := os.MkdirAll(filepath.Join(g.OutputDir, d), 0755); err != nil {
--- a/internal/generator/generator_test.go
+++ b/internal/generator/generator_test.go
@@ -109,6 +109,7 @@ func TestGenerateProjectsCompile(t *testing.T) {
 		"internal/mcp/cobratree/shellout_test.go",
 		"internal/mcp/cobratree/cli_path.go",
 		"internal/mcp/cobratree/names.go",
+		"internal/cliutil/testenv/testenv.go",
 	}
 
 	tests := []struct {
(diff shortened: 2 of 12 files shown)


sidorovanthon:
Ready for review.

CI is green across all lanes, Greptile Review passes with no actionable markers on the current head, and there are no unresolved threads.

Beyond CI, the fix was validated against a real print on Windows 11, since the incident this closes was platform-specific and CI runs on Linux. A CLI generated with the patched templates runs its full suite with a populated profile exposed and leaves it byte-identical. The same procedure against a generator built from `main` reproduces the escape and writes `credentials.toml` straight into the profile, which is the shape of the original token loss.

No CODEOWNERS-protected paths are touched, so this needs only `ready-to-merge` and not a separate protected-path approval.

Two follow-ups are deliberately queued behind this one rather than racing it, since both land in the `testenv` package it introduces:

- #3702, emitted credential tests cannot pass on Windows because the `t.TempDir()` sandbox fails the read-time DACL guard
- #3704, an `AGENTS.md` rule requiring the field proof described above

#3704 in particular must merge after this PR: the check it prescribes cannot pass on current `main`.

PR #3805, opened 2026-07-25
fix(skills): skip output review when no passing samples

dommonkhouse:
## Summary

- classify zero eligible `status: pass` live-check samples as `SKIP`
- prevent vacuous PASS results when every sampled command failed or was excluded
- reserve PASS for reviews that assessed at least one successful sample

## Evidence

A generated Video System CLI produced five `status: fail` samples and zero passing samples. The existing instructions excluded all five, then allowed the reviewer result to be recorded as PASS because there were no findings. Plausibility was not assessed.

## Validation

- `git diff --check`
- `skill-comply --dry-run` extracted an explicit `validate_livecheck_output` step requiring at least one eligible passing sample
- local vendored copy will use the identical patch with this PR recorded as provenance


dommonkhouse:
All substantive checks are green at e8958eaa, Greptile is 5/5, and there are no unresolved review threads. Could a maintainer apply the required `ready-to-merge` label so Mergify can queue this zero-sample output-review correction?

PR #3841, opened 2026-07-29
feat(platform): pluggable credential resolvers — remove the 1Password default

jamaynor:
## What

Every printed CLI currently ships a 1Password dependency whether or not its operator uses 1Password:

- `platform_profile.go.tmpl` emits an `OnePasswordResolver` that shells out to `1password-pp-cli secrets read --reveal`
- `ValidateCredentialReference` rejects any reference not beginning with `op://`
- `platform_cli.go.tmpl`'s `platformResolverFactory` returns that resolver as the only option

The `CredentialResolver` interface was already the right abstraction — it was just bypassed by a hardcoded validator and a single hardcoded implementation. This puts a registry behind it:

- `RegisterCredentialResolver(scheme, resolver)` — scheme and resolver register **together**, so a scheme the validator accepts but no resolver handles cannot exist
- `ValidateCredentialReference` is registry-driven; with nothing registered every reference is rejected with an error saying so
- `RegistryResolver` dispatches by scheme and is what the generated CLI uses; no vendor is privileged
- New spec field `auth.credential_resolvers: [file|bitwarden|onepassword]`, defaulting to `[file]` — the one resolver that depends on nothing installed. 1Password remains available as an explicit opt-in
- Unknown resolver names are a spec validation error naming the catalog, not a silent skip

## Notes for review

- `CredentialScheme.Validate` exists because a filesystem path is not a sequence of opaque identifiers: `file:///etc/key` splits to a leading empty component, so the generic component rules would reject every absolute path
- Test fixtures now come from a `credentialRefFixture` template function instead of hardcoded `op://` strings — hardcoding one vendor's syntax in the test templates is part of how the original assumption stayed invisible
- Subprocess-backed resolvers share `runCredentialCommand`, which enforces: no credential value ever enters argv, and provider stderr never enters an error (a secret manager may echo the resolved value while reporting an unrelated failure)
- Second commit is an unrelated pre-existing fix: the read-only-dir save test fails under root (`CAP_DAC_OVERRIDE` bypasses the DAC check); it now skips as root, mirroring its existing Windows skip. Happy to split it out if preferred

## Behaviour change

A reprint of a CLI that used `op://` references rejects them unless its spec opts into `onepassword`. In practice nothing is affected today: `registerPlatformSource` is never called by any printed CLI, so this path is dead code everywhere — which is also why the hard dependency went unnoticed.

## Verification

- `internal/spec` suite green including 9 new tests
- Full `internal/generator` suite run to completion (`-timeout 45m`): no failures attributable to this change (one pre-existing root-environment failure, fixed in the second commit)
- Three end-to-end prints of a real spec — default, `[bitwarden, file]`, `[onepassword]` — each passing the generator's own build/vet/test/bundle gates; the default print contains zero 1Password references


tmchow:
Closing this in favor of a narrower superseding fix. The review confirmed that the generated core should not hardcode OnePassword, but this PR expands that cleanup into a generator-owned credential-provider catalog with file as the implicit default, Bitwarden support, new configuration, and new reprint semantics. That is a broader product and compatibility decision than the original bug requires. The superseding implementation will remove the hardcoded OnePassword runtime and provider-specific reference validation, keep the generated core provider-neutral, allow an optional resolver to be supplied by the downstream platform-source registration, default to no resolver, and validate credentials only for the selected source. Provider implementations and catalogs will remain out of scope unless designed as a separate feature. Thank you for surfacing the coupling and doing the investigation that made the narrower boundary clear.

jamaynor:
thx

PR #3861, opened 2026-08-01
fix(cli): migrate legacy parent-key store rows

### LINKED ISSUE #3691: fix(cli): adding a resource to resourceParentKeyColumns strands permanent duplicate rows — reconcile can never sweep them
## Summary

When a resource type gains a parent-keyed storage id — i.e. it is added to `resourceParentKeyColumns`, so `resourceStorageID` starts writing `<id>\x00<parent>` instead of the bare id — every row an existing store already holds keeps the **bare** key. The next sync inserts the same entity again under the **composite** key. Both rows then coexist in the typed table, in `resources`, and in FTS.

There is no store migration that re-keys the old rows, and `ReconcilePartition` structurally cannot remove them: the victim test is `seen[BareResourceID(id)]`, and for an entity that still exists upstream the bare id *is* in the seen set, so the stale row is skipped as "enumerated this run". The duplicate is therefore permanent — it survives `sync --full` forever, and only disappears if the entity is deleted upstream.

## Reproduction

Against the generated store package, seeding one entity in both shapes and running a full, successful enumeration:

```
до реконсиляции: строк modules = 2
реконсиляция удалила: 0
после реконсиляции: строк modules = 2
  осталась строка: composite=true  name="current"
  осталась строка: composite=false name="stale"
```

The stale row keeps its **outdated payload** — it is never updated either, since upserts now write to the composite key.

## Observed on a real store

A published CLI's store that lived across the engine upgrade (`generic` table counts):

| resource | rows | composite | bare | same entity in both forms |
|---|---|---|---|---|
| `modules` | 147 | 25 | 122 | 16 |
| `states` | 60 | 40 | 20 | 20 |
| `projects_issues` | 646 | 646 | 0 | 0 |
| `cycles` | 16 | 16 | 0 | 0 |
| `labels` | 48 | 48 | 0 | 0 |

The fully-migrated types are the ones that happened to be re-synced after the upgrade; every entity synced *before* it and not re-synced since is still sitting there under the old key. For `states`, every bare row is a duplicate.

## Why it matters

The generic table is what the agent-facing surfaces read:

- `sql` and `search` return both rows, so counts are inflated and an agent aggregating over them double-counts silently.
- MCP `context` resource counts inherit the same inflation.
- A bare-id lookup can resolve to the **stale** row and return payload data that stopped being updated at the moment of the upgrade. This is the worst failure mode: not a missing row, a plausible stale one.
- FTS holds both rows, so search hits duplicate.

None of this surfaces as an error. The store looks healthy.

## Where it comes from

- `internal/generator/templates/store.go.tmpl` — `resourceStorageID` appends the NUL-composite suffix for any type in `resourceParentKeyColumns`; nothing re-keys rows written before the type was added to that map.
- `ReconcilePartition` in the same template — the `seen[BareResourceID(id)]` guard keeps a stale bare row whenever the entity still exists, which is precisely the common case.
- The migration framework (`PRAGMA user_version`, `StoreSchemaVersion`) has a migration for the *generic* resources composite key, but none that re-keys typed dependent rows when a resource type's key shape changes.

## Proposed fix

A schema-version migration that, for every type in `resourceParentKeyColumns`, re-keys existing rows to the composite form and drops the stale duplicate where both shapes exist for one bare id (keeping the composite row, which is the one upserts maintain). It must rebuild the affected FTS rowids, since `ftsRowID` is derived from the id.

The alternative — teaching `ReconcilePartition` to detect a bare/composite pair and drop the bare one — is cheaper but leaves the duplicate live until the next full sync of that partition, and does nothing for a store the user never full-syncs.

Worth settling in the same change: `resourceParentKeyColumns` is generated per CLI, so the migration has to be driven by the emitted map rather than a hardcoded list, and it must be a no-op for a store that never held the bare shape.

## Verification this needs

Per `AGENTS.md` ("Generator fixes require generated-output proof"): a generated-output test that seeds a store in the pre-upgrade shape, opens it with the new binary, and asserts one row per entity under the composite key with the current payload plus no orphaned FTS rowid — not only an assertion on template text. Plus `go test ./...` and `scripts/golden.sh verify`.

## Scope note

Found while verifying the `module of` fix in mvanhorn/printing-press-library#1471: the blank-name bug was a symptom of the same key-shape transition, and investigating the store turned up the duplicates. Related platform-assumption issues: #3689, #3690.

sidorovanthon:
## Intent

Fixes #3691.

Existing stores can retain bare-ID rows after a dependent resource begins using parent-keyed storage IDs. Current syncs then maintain a second composite-ID row, while reconcile treats the stale bare ID as seen and never removes it. Agents consequently see duplicate counts, duplicate FTS hits, and plausible stale payloads.

## Approach

Bump the emitted store schema version and add an idempotent migration driven by the generated `resourceParentKeyColumns` map. For each legacy bare row, the migration computes the current composite key, preserves the maintained composite row when both shapes exist, or re-keys the bare row when it is the only copy. It mirrors the transition into generated typed tables, removes the obsolete generic FTS rowid, and indexes the canonical composite payload; typed-table FTS triggers handle the corresponding typed cleanup.

Generated runtime coverage seeds duplicate, bare-only, and composite-only rows, then verifies the generic table, typed projection, generic FTS, typed FTS, payload precedence, schema stamp, and current-version no-op reopen.

## Scope

Primary area: generated SQLite store migrations and their generated-output fixtures/tests.

Why this belongs in this repo: the key-shape transition is emitted by the Printing Press for every dependent resource. Fixing one printed CLI would leave the same permanent-duplicate failure in every future print.

## Risk

This changes the database-open path for stores below schema v5 (learn-disabled) or v10 (learn-enabled). The migration runs inside the existing `BEGIN IMMEDIATE` migration transaction, is driven only by generated resource/table metadata, skips resources without a usable parent value, and is a no-op when no legacy bare shape exists.

## Output Contract

Generated `internal/store/store.go` now emits schema v5/v10 plus the parent-key migration and typed-table dispatch. Four existing golden store fixtures were intentionally refreshed. The emitted contract is exercised by `TestGeneratedStoreMigratesLegacyParentKeyRows`, which compiles the generated module and executes the migration against SQLite rather than asserting template text alone.

## Verification

Passing checks:

- `go fmt ./...`
- `go vet ./...`
- `go test ./... -run '^$'`
- `go test ./internal/generator -run '^TestGeneratedStoreMigratesLegacyParentKeyRows$' -count=1`
- targeted schema-version generator tests
- generated learn-enabled CLI: `go test ./internal/store -count=1`
- `scripts/verify-generator-output.sh generate-sync-walker-api generate-learn-loop-api generate-learn-disabled-api`
- default generated-output sweep: 9/10 generated modules built; the unrelated existing GraphQL fixture still emits an unused `internal/client` import
- `scripts/golden.sh verify`: all four changed `internal/store/store.go` fixtures pass; the Windows checkout still reports 23 unrelated baseline cases involving exit codes, CRLF-sensitive manifests/docs, and existing GraphQL/async output drift

`go test ./...` was run in full. It completed but is not green on this Windows host because of unrelated existing platform/environment failures, including Unix path/file-mode assertions and a generated `-race` test while `CGO_ENABLED=0`. `golangci-lint` was unavailable on this machine.

## AI / Automation Disclosure

### DIFF
--- a/internal/generator/learn_schema_v9_test.go
+++ b/internal/generator/learn_schema_v9_test.go
@@ -31,16 +31,15 @@ func generateLearnStore(t *testing.T, name string, learnEnabled bool) (string, s
 	return string(storeGo), outputDir
 }
 
-// TestLearnSchemaV9_EnabledEmitsV9WithCandidateAndEventTables pins the v9
-// schema bump: a learn-enabled store advances StoreSchemaVersion to 9 and
-// carries the learn_candidates and learn_events tables (with their CHECK
-// constraints and indexes) as additive CREATE IF NOT EXISTS migrations.
-func TestLearnSchemaV9_EnabledEmitsV9WithCandidateAndEventTables(t *testing.T) {
+// TestLearnSchemaV9_EnabledEmitsCandidateAndEventTables pins the tables
+// introduced at v9 while allowing later store migrations to advance the
+// current schema version.
+func TestLearnSchemaV9_EnabledEmitsCandidateAndEventTables(t *testing.T) {
 	t.Parallel()
 
 	src, _ := generateLearnStore(t, "learn-v9-enabled", true)
 
-	require.Contains(t, src, "const StoreSchemaVersion = 9")
+	require.Contains(t, src, "const StoreSchemaVersion = 10")
 	require.NotContains(t, src, "const StoreSchemaVersion = 8")
 	for _, want := range []string{
 		"CREATE TABLE IF NOT EXISTS learn_candidates",
@@ -62,7 +61,7 @@ func TestLearnSchemaV9_EnabledEmitsV9WithCandidateAndEventTables(t *testing.T) {
 // resourcesFTSContentSchemaVersion stays 4 in BOTH learn shapes. The old
 // conditional 8 rode the learn bump by accident and forced a full FTS
 // content rewrite on every v4-v7 learn store open; with the pin, v4 and v8
-// stores opened by a v9 binary take the additive-only migration path.
+// stores opened by the current binary take the additive-only FTS path.
 func TestLearnSchemaV9_FTSContentPinIsUnconditional(t *testing.T) {
 	t.Parallel()
 
@@ -72,7 +71,7 @@ func TestLearnSchemaV9_FTSContentPinIsUnconditional(t *testing.T) {
 
 	disabled, _ := generateLearnStore(t, "learn-v9-fts-disabled", false)
 	require.Contains(t, disabled, "const resourcesFTSContentSchemaVersion = 4")
-	require.Contains(t, disabled, "const StoreSchemaVersion = 4")
+	require.Contains(t, disabled, "const StoreSchemaVersion = 5")
 	for _, gone := range []string{"learn_candidates", "learn_events"} {
 		require.NotContains(t, disabled, gone,
 			"learn-disabled spec must not emit the %s migration", gone)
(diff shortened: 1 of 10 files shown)


sidorovanthon:
Closing this draft because I misread the existing maintainer ownership comment on #3691 as an implementation handoff and opened parallel work. The issue remains with its current assignee. Sorry for the noise.

PR #3872, opened 2026-08-01
fix(cli): anchor scorer target paths and research discovery

### LINKED ISSUE #1347: scorer(dogfood): novel-features check needs --research-dir flag or parent-walk fallback for research.json
> Filed by `/printing-press-retro` from a run of **google-sheets**
> (scorecard 65/100 library, 100% verify pass).

## Summary

`dogfood`'s novel-features check reads `<CLI_DIR>/research.json`. The main `/printing-press` skill writes research.json to `$API_RUN_DIR/research.json` (one level up from `$CLI_WORK_DIR`). dogfood reports `Novel Features: SKIP (no research.json)` even when a perfectly valid research.json exists for the run.

## Where to look

- **Component:** scorer (`dogfood` command)
- **Likely area:** dogfood's novel-features loader; check where it resolves `research.json` from `--dir`
- **Triggered when:** `dogfood --dir <cli-work-dir>` runs during shipcheck without `--research-dir`, against a CLI generated through the main skill

## What we observed

In this session, the first `dogfood` invocation (via `shipcheck`) reported:
```
Novel Features: SKIP (no research.json)
```
even though `$API_RUN_DIR/research.json` existed with 12 populated `novel_features` entries. Required a manual `cp $API_RUN_DIR/research.json $CLI_WORK_DIR/research.json` to make dogfood's check fire.

The main skill's Phase 1.5e writes research.json to the run directory, not the CLI directory, because the run directory is the durable archival shape (`manuscripts/<api>/<run-id>/`). `shipcheck` already accepts `--research-dir` to bridge this; `dogfood` does not.

## Suspected root cause

The dogfood novel-features loader resolves `research.json` from `<dir>/research.json` only, with no parent-walk or `--research-dir` flag.

## Suggested direction

Two changes:

1. **Accept `--research-dir <path>` on `dogfood`**, mirroring the flag `shipcheck` already takes. When passed, look for `research.json` at `<research-dir>/research.json` first.
2. **Parent-walk fallback when neither flag nor in-CLI-dir hit lands**: walk up at most 3 parents from `<dir>` looking for a `research.json` next to a `state.json` (the main skill writes both into the run dir).

Pick whichever is the smaller change; the flag alone is sufficient if shipcheck always forwards it. Currently `shipcheck` does pass `--research-dir` through to `validate-narrative` and `scorecard` but not `dogfood` — also worth checking.

## Acceptance criteria

- **Positive:** `dogfood --dir <cli> --research-dir <run>` reads research.json from the run dir and reports novel-features status correctly. Without the flag, dogfood walks up max 3 parents from `<cli>` looking for `research.json`.
- **Negative:** When no research.json exists in `<cli>` or any of 3 parents and no `--research-dir` passed, dogfood reports `Novel Features: SKIP (no research.json)` with the same wording it does today.

## Frequency

Every generation through the main `/printing-press` skill. The symptom is silent (just a `SKIP` line) so it goes unnoticed unless someone is optimizing novel-features specifically — which is exactly when it matters.

## Complexity

small — add a flag, optionally add a parent-walk. No structural changes.

## Scope boundary

Only the novel-features loader's path resolution changes; the rest of dogfood is untouched.

## Dependencies

None.

## Related issues

None found in dedup scan.

## Artifacts

Retro document: `/Users/jacobsearing/printing-press/manuscripts/google-sheets/20260514-081453/proofs/20260514-084710-retro-google-sheets-pp-cli.md` (local — no upload requested).

---

### LINKED ISSUE #3851: scorer: scorecard --live-check silently no-ops when --dir is relative (auxiliary paths resolved against cwd)
## What happened

Run from inside the CLI directory:

```
$ cd <cli-dir> && cli-printing-press scorecard --dir . --live-check
  Sample Output Probe (live command sample)
  Unable to run: no research.json: open research.json: no such file or directory
```

`research.json` was resolved against the **process cwd**, not against `--dir`. The scorecard
then printed a full report and **exited 0**.

The identical run with an absolute `--dir`, from a different cwd, resolved correctly and
exercised the probe:

```
$ cd /tmp && cli-printing-press scorecard --dir <abs-cli-dir> --live-check
  Sample Output Probe (live command sample)
  Binary refresh: fresh_fallback ...
  Passed: 10/11
```

## Scorer correct?

**No.** Zero live coverage was reported as a clean pass.

## Two observed symptoms, same class

- this run: `no research.json: open research.json: no such file or directory`
- a polish pass on the same CLI: `unable: binary "build/stage/bin" is not executable`

Both are auxiliary paths resolved relative to cwd instead of `--dir`.

## Why this is worse than a normal bug

The failure mode is a **silent pass**. Nothing prompts the caller to retry with an absolute
path, so the natural outcome is false confidence that live coverage ran. It was only caught
here by comparing two invocations side by side.

Relative `--dir` is the common shape for anything invoked from inside the CLI directory, so
this is not API-shaped — it can affect **any** printed CLI.

## Durable fix

1. Absolutise `--dir` on entry and resolve `research.json`, the staged binary, and every other
   auxiliary path against it.
2. Separately: an unrunnable live probe should not be silently compatible with an overall
   pass. Surface it as a machine-detectable state (`live_check: unavailable`, or a non-zero
   signal) rather than prose buried in the report.

## Acceptance criteria

- positive: `cd <cli-dir> && scorecard --dir . --live-check` exercises the same feature count
  as the absolute-path invocation.
- negative: a genuinely absent `research.json` reports unavailable rather than fabricating a
  result.
- the "probe could not run" state is machine-detectable, not just prose.

**Scope boundary:** path resolution and the unavailable-state signal only; not a redesign of
what the probe measures.

## Related issues

- #3745 — harness legs resolve the CLI binary by directory basename (silent SKIP / false-fail
  in worktrees). Very likely the same root cause family; a fixer should check whether one
  change closes both.
- #3646 — live dogfood hides hollow coverage (pass rate computed over checks that never ran).
  Same "didn't run but looks fine" class.
- #3834 — `--live-check` can rebuild the staged binary underneath a running probe. Same
  command, different defect.

---
*Filed from `/printing-press-retro` on the alpaca run. Complexity: small. Component: scorer.*

tmchow:
## Intent

Relative scorer invocations could report no research or binary coverage even when the same CLI worked with an absolute path. This makes scorecard and dogfood resolve their target consistently and makes an unavailable live probe explicit to machine and human consumers.

## Approach

- Canonicalize scorer target directories before static scoring, dogfood, live-check, binary lookup, and bounded research discovery.
- Reuse the three-level parent walk for live-check and dogfood novel-feature validation while preserving the explicit `--research-dir` override.
- Add `live_check.status` and human-readable unavailable output without changing existing counters or pass-rate fields.

## Scope

Primary area: scorer path resolution in `internal/cli` and `internal/pipeline`.

Why this belongs in the Printing Press: the bug affects every generated CLI evaluated through these scorer commands, not one printed CLI.

## Risk

The parent fallback is bounded to three ancestors and chooses the nearest `research.json`. No generated templates, MCP surfaces, auth behavior, or external API calls changed.

## Output Contract

`live_check.status` is `available` or `unavailable`; existing live-check fields remain unchanged. Human scorecard output now prints `Status: unavailable` when the probe cannot run.

## Verification

- `go test ./internal/cli` passed.
- Focused relative-path, parent-research, explicit-research, and live-check JSON tests passed in `internal/pipeline` and `internal/cli`.
- `go build -o ./cli-printing-press ./cmd/cli-printing-press`, `go vet ./...`, targeted `golangci-lint`, and `scripts/golden.sh verify` passed. Golden verification covered 32 cases.
- A full repository test run was attempted. The changed CLI package passed; unrelated generator fixtures failed because local generated CLIs emitted store-schema warnings before JSON, and the pipeline package hit `TestLiveDogfoodHappyArgsHonorsPPHappyArgs` under the shared host. That test passes in isolation; the scorer-focused tests are green.

## Not included

Implicit research fallback remains limited to novel-feature validation, matching #1347's stated loader scope. Reimplementation and description-drift checks keep their existing explicit research-directory contract.

Closes #3851
Closes #1347


tmchow:
Accepted for later: Greptile latest asks preserving implicit discovery for legacy/minimal layouts without run provenance, while earlier review rounds identified that same-API metadata-free ancestor selection cannot be proven safe and can mutate the wrong run. This PR intentionally keeps the safety-first contract: implicit discovery requires matching API, working-directory, and run identities; legacy/minimal callers must pass an explicit research directory. A follow-up can add a durable legacy ownership marker without weakening this guard.

Issue #3944, opened 2026-08-02
docs: x-pp-default-rate-limit and x-pp-membership-field are parsed but missing from SPEC-EXTENSIONS.md


1aifanatic:
## Problem

`docs/SPEC-EXTENSIONS.md` opens with an explicit completeness contract:

> OpenAPI allows extension fields anywhere, but the Printing Press only reads the extensions listed here.
>
> Source of truth: `internal/openapi/parser.go`. This document should be updated in the same change as any new `Extensions["x-*"]` lookup in that file.

Two extensions are read by the parser and mapped onto real spec fields today, but neither appears anywhere in `SPEC-EXTENSIONS.md` (the Summary table or a section body):

### `x-pp-default-rate-limit`
- Constant: `internal/openapi/parser.go:63` (`extensionDefaultRateLimit = "x-pp-default-rate-limit"`)
- Parsed at `internal/openapi/parser.go:636` via `parseDefaultRateLimitOpenAPIExtension` (`parser.go:939-984`)
- Mapped to `APISpec.DefaultRateLimit` (`internal/spec/spec.go:312`)
- Validated at `internal/spec/spec.go:4352`

### `x-pp-membership-field`
- Parsed at `internal/openapi/parser.go:5765-5783` (`readPathItemMembershipField`)
- Mapped to `Endpoint.MembershipField` (`internal/spec/spec.go:2363`)
- Consumed downstream by the profiler: `internal/profiler/profiler.go:890-895`

Neither has any occurrence in `docs/SPEC-EXTENSIONS.md`.

## Impact

The doc under-describes the contract an OpenAPI spec author can actually rely on. Someone writing a spec that needs an adaptive rate limiter (`x-pp-default-rate-limit: auto`) or a membership-gated parent collection (`x-pp-membership-field: is_member`) has no documented reference, and the doc actively claims to list every extension the generator reads.

## Where to look
- `docs/SPEC-EXTENSIONS.md` (Summary table, `x-rate-class` section as the rate-ext sibling, `x-pp-tenant-scope-column` section as the path-item-field sibling)

## Suggested fix
Add a Summary-table row and a section for each, following the existing `x-rate-class` / `x-pp-tenant-scope-column` entry style. The prose can be sourced from the parser code and the struct doc-comments, so no new behavior is being proposed.

Happy to send a PR.

1aifanatic:
[...]

Issue #3918, opened 2026-08-02
generator v4.30.0: platform conformance tests assert POSIX 0600 and fail on Windows (got 0666)


rodrigoedu94-png:
Filed from a Windows 11 run of the Printing Press (generator **v4.30.0**, Go 1.26.5) while regenerating a browser-sniffed CLI.

## Summary

The new `platform` conformance/migration tests assert a literal `0600` file mode on the client profile. On Windows, Go's `os.Chmod` cannot express POSIX permission bits on NTFS, so `os.Stat().Mode().Perm()` reports `0666` for a file created with `0600`. Every affected test fails on a freshly generated CLI, with no hand-edits involved.

This is the same class as #3853/#3858 (fixed in v4.30.0 — thank you): a POSIX assumption baked into an emitted test that does not hold on Windows. The state-dir family now passes; this is a new instance in the `platform` feature shipped by the same release.

## Reproduce

```
cli-printing-press generate --spec <any-spec> --output <dir> --force --lenient
cd <dir> && go test ./...
```

## Observed

```
--- FAIL: TestPlatformCLIConformanceSurfaceAndGate
    platform_cli_test.go:180: client profile permissions must be 0600, got 0666
--- FAIL: TestPlatformCLIConformanceMismatchFailsBeforeCommand
    platform_cli_test.go:386: mismatch error = client profile permissions must be 0600, got 0666
--- FAIL: TestPlatformMigrationIsVerifiedIdempotentAndQuarantinesLegacyState
    platform_cli_test.go:431: migrated profile = (*platform.Profile)(nil) err=client profile permissions must be 0600, got 0666
--- FAIL: TestPlatformMigrationAdoptsOnlyVerifiedTenantDatabase
    platform_cli_test.go:512: verified migration output = { ... }
```

Also `FAIL internal/platform` — `conformance_test.go:257` and `conformance_test.go:580` (`limiter ledger mode = -rw-rw-rw-`), same root cause.

Package results on the generated tree (Windows 11):

```
FAIL  internal/cli          (4 TestPlatform* only)
FAIL  internal/platform
ok    internal/client  internal/cliutil  internal/config  internal/learn
ok    internal/mcp     internal/mcp/bound  internal/mcp/cobratree  internal/store
```

## Root cause

[...]

PR #3946, opened 2026-08-02
docs(cli): document x-pp-default-rate-limit and x-pp-membership-field extensions

1aifanatic:
## Intent

`docs/SPEC-EXTENSIONS.md` opens with an explicit completeness contract — *"the Printing Press only reads the extensions listed here"*, with `internal/openapi/parser.go` as the source of truth. Two extensions are parsed and mapped onto real spec fields today but were absent from the reference, so the doc under-described its own contract.

Issue: #3944

## Approach

Adds one Summary-table row and one section for each extension, following the existing `x-rate-class` (root/info rate extension) and `x-pp-tenant-scope-column` (path-item field extension) entry styles. No new behavior is proposed — every rule line is transcribed from the parser code and the struct doc-comments:

- **`x-pp-default-rate-limit`** → `APISpec.DefaultRateLimit`. Root/info; value is `"auto"` or a non-negative number (`parser.go:939-984`); sets the generated `--rate-limit` default (`spec.go:304-312`).
- **`x-pp-membership-field`** → `Endpoint.MembershipField`. Path-item string naming the boolean membership field in a parent collection row payload (`parser.go:5765-5783`, `spec.go:2356-2363`); consumed by the profiler for dependent fan-out (`profiler.go:890-895`).

## Scope

Primary area: `docs/` (OpenAPI extension reference).

Why this belongs in this repo: `docs/SPEC-EXTENSIONS.md` is the canonical Printing Press extension reference and its own header mandates it stay in sync with `internal/openapi/parser.go`. This PR closes that gap for two extensions already parsed by the generator (machine) — it documents existing machine behavior, not a printed-CLI fix.

## Risk

N/A — documentation-only. No template, generator code, parser, manifest, MCP schema, command output, scorecard, or pipeline artifact is touched, so there is no generated-output or output-contract impact.

## Output Contract

N/A — docs-only change; does not affect templates, generated files, manifests, command output, MCP schemas, scorecard output, or pipeline artifacts.

## Verification

- [ ] Not applicable — this is a docs-only edit to `docs/SPEC-EXTENSIONS.md`; no generator/template/parser code changed.
- `go build -o ./cli-printing-press ./cmd/cli-printing-press` — passes.
- `scripts/golden.sh verify` — the only diffs are the pre-existing Windows CRLF / `spec_checksum` platform artifacts (identical 32-case failure count with this branch stashed vs. applied; tracked by #3689 / #3918). None of the golden diffs touch any extension, rate-limit, or membership field. CI runs on Linux where these pass.

Not run: full `go test ./...` on Windows reports only the pre-existing POSIX-0600 permission failures (#3918); not caused by this change (no Go files touched).

## AI / Automation Disclosure

- [ ] AI-reviewed only
- [ ] Fully automated

### DIFF
--- a/docs/SPEC-EXTENSIONS.md
+++ b/docs/SPEC-EXTENSIONS.md
@@ -20,6 +20,7 @@ in the same change as any new `Extensions["x-*"]` lookup in that file.
 | `x-roles` | root or `info` | `APISpec.Roles` | No |
 | `x-tier-routing` | root or `info` | `APISpec.TierRouting` | No |
 | `x-rate-class` | root or `info` | `APISpec.RateClass` | No |
+| `x-pp-default-rate-limit` | root or `info` | `APISpec.DefaultRateLimit` | No |
 | `x-mcp` | root or `info` | `APISpec.MCP` | No |
 | `x-cache` | root or `info` | `APISpec.Cache` | No |
 | `x-learn` | root or `info` | `APISpec.Learn` | No |
@@ -56,6 +57,7 @@ in the same change as any new `Extensions["x-*"]` lookup in that file.
 | `x-pp-sync-walker` | operation | `Endpoint.Walker` | No |
 | `x-pp-dispatch-param` | parameter | `Param.DispatchParam` | No |
 | `x-pp-tenant-scope-column` | path item | *reserved for follow-up tenant-scoped reconcile; not parsed yet* | No |
+| `x-pp-membership-field` | path item | `Endpoint.MembershipField` | No |
 
 ## `info` Extensions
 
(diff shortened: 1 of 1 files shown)


1aifanatic:
Greptile flagged that the `x-pp-membership-field` rules omitted the generated store's identifier restriction. Verified and addressed in 6532eaed: the generated `Store.NonMemberParents` interpolates the field name into its non-member query and validates it against `validIdentifierRE` (`^[a-zA-Z_][a-zA-Z0-9_]*$`, `internal/generator/templates/store.go.tmpl:42,1103-1107`), so dotted/nested-path field names are rejected at runtime and the skip becomes a no-op. The doc now states that restriction explicitly so authors pick a simple-identifier field name.

PR #3995, opened 2026-08-07
fix(generator): make internal/platform permission checks portable to Windows

### LINKED ISSUE #3990: 4.30.x is unusable on Windows: internal/platform uses raw POSIX mode comparison (cliutil already had the build-tagged split)
## Summary

`internal/platform` (new in 4.30) verifies credential-file permissions with a raw POSIX mode comparison. On Windows this can never pass, so **4.30.x is unusable out of the box** on a Windows host — the printed CLI fails its own profile/credential checks before doing any work.

## Root cause

Windows has no POSIX mode bits. `os.Chmod` only toggles the read-only attribute, so a file written with `0600` stats back as `-rw-rw-rw-` (`0666`). A check of the form "mode must equal 0600" is unsatisfiable.

The sibling package `internal/cliutil` already had a correct build-tagged split for exactly this. The new `internal/platform` package did not adopt the existing pattern.

Additionally, `cliutil`'s DACL fallback is itself too strict for **AzureAD-joined hosts**: it rejects files whose ACLs are inherited rather than explicitly set, which is the normal state for a user profile directory on a domain-joined machine.

## Reproduce

Windows 11, Go 1.26.x, any 4.30.1-generated CLI. Run any command that touches the profile/credential path.

## Environment

- Windows 11 Pro 26200, AzureAD-joined
- cli-printing-press 4.30.1

## Workaround in use

Local fork, kept out of generated files so `generate --force` preserves it:

- `internal/platform/profile_perms_unix.go` — POSIX mode check (unchanged behavior)
- `internal/platform/profile_perms_windows.go` — warn-not-fail
- `internal/platform/profile.go` — two inline checks replaced with `verifyPrivateFilePerms(...)`
- POSIX-only tests skipped on `runtime.GOOS == "windows"`: `internal/platform/conformance_test.go` x3, `internal/cliutil/credentials_test.go` x1, `internal/cli/platform_cli_test.go` x1

## Suggested fix

Build-tag `internal/platform`'s permission check the same way `internal/cliutil` already does, and relax the DACL path to accept inherited ACLs that grant no access beyond the owner.

paonessa:
Fixes #3990.

## The bug

`internal/generator/templates/platform_profile.go.tmpl` gated two reads on raw
POSIX mode bits:

```go
if info.Mode().Perm()&0o077 != 0 {            // client profile
if info.Mode().Perm() != 0o600 {              // cache fingerprint key
```

Neither can be satisfied on Windows. `os.Chmod` only toggles the read-only
attribute, so a file written `0600` stats back `0666`. Every printed CLI failed
before it could do any work.

`internal/cliutil` already solved exactly this with a `//go:build` split
(`creds_perms_unix.go` / `creds_perms_windows.go`). `internal/platform`, added
later, did not adopt the pattern.

## This is already breaking CI-equivalent runs on Windows

`TestGeneratedPlatformCredentialResolverSupportsPreservedDownstreamAdapter`
**fails on a clean checkout of `main`** on a Windows host, with the bug verbatim:

```
Error: client profile permissions must be 0600, got 0666
```

That test passes with this change. It is the clearest evidence I can offer that
this is not a theoretical portability nit.

## The fix

A build-tagged helper pair, mirroring `cliutil`:

- `platform_perms_unix.go.tmpl` — the POSIX mode check, unchanged in substance
- `platform_perms_windows.go.tmpl` — delegates to the existing
  `cliutil.VerifyCredsPerms` (owner + DACL against the current-user SID)

Both call sites now use one generic helper, `verifyPrivatePerms(path, info, label)`.

### Why the Windows side warns instead of failing

I tried strict delegation first. It rejects real files, and the reason is
structural rather than environmental:

**Nothing in a printed CLI ever sets a restrictive DACL.** Both
`writePrivateFile` (this package) and `cliutil.AtomicWritePrivateFile` only call
`Chmod`, a no-op on NTFS. A private file therefore inherits whatever ACL its
parent directory carries. Under `%USERPROFILE%` that is typically owner +
SYSTEM + Administrators and passes. Under `%TEMP%` — or any directory carrying
stale ACEs from a previous domain identity — it does not:

```
token file DACL grants access to a non-owner principal (S-1-5-21-…-836572598)
```

Hard-failing the read would block the user for an ACL the tool never promised to
set. The finding is surfaced on stderr instead, and the read proceeds.

The path to enforcement is writer-first: have `writePrivateFile` set an explicit
DACL on Windows, then turn the warning into an error. Happy to do that as a
follow-up if you want it — it belongs in its own PR because it changes
`cliutil`'s token path too.

## A correction to the issue

The issue also claimed `cliutil`'s DACL evaluator rejects inherited ACLs on
AzureAD-joined hosts. **That was wrong and I have retracted it in a comment.**
`evalCredsSecurity` reads only the ACE type and trustee; the inheritance flag is
never examined. Verified by running the real evaluator against this host's
actual credentials-file SDDL — it passes.

## The generated conformance tests had the same bug

`platform_conformance_test.go.tmpl` asserts POSIX modes in three places, so every
printed CLI shipped a test suite that cannot pass on Windows:

```
conformance_test.go:104  profile mode = -rw-rw-rw-
conformance_test.go:316  cache fingerprint key mode = -rw-rw-rw-
conformance_test.go:707  limiter ledger mode = -rw-rw-rw-
```

Each now skips on `runtime.GOOS == "windows"` with the reason stated inline.
`TestGeneratedPlatformRuntimeContract` — also red on clean `main` — passes with
this change.

## Behaviour change to flag

The cache-fingerprint-key check was `Perm() != 0o600` (exact) and is now
`Perm()&0o077 != 0`, matching the client-profile check and `cliutil`. This
accepts `0400` where it previously did not. Strictly safer-or-equal, and it
removes a second spelling of the same rule.

## Verification

- New `TestGenerate_PlatformProfilePermsAreBuildTagged` — written failing first;
  asserts no raw mode mask in `profile.go`, both build-tagged helpers emitted
  with correct tags, and that the Windows side surfaces rather than swallows.
- `TestGeneratedPlatformCredentialResolverSupportsPreservedDownstreamAdapter`
  goes **red → green**.
- A generated CLI cross-compiles for `windows/amd64`, `linux/amd64`, and
  `darwin/arm64`; `go vet` clean under both `GOOS=windows` and `GOOS=linux`.
- `TestGeneratedPlatformRuntimeContract` goes **red → green** as well.
- `go test ./internal/generator/ -run 'Platform|Perms|Creds|Profile|Conformance'`
  passes on Windows; `go vet` clean.

Note for reviewers on non-Windows CI: these tests were already green on
POSIX — every failure this PR fixes is Windows-only, which is precisely why it
went unnoticed.


tmchow:
@paonessa can you address the code review feedback please?

PR #3998, opened 2026-08-07
fix(generator): pick docs examples by method, not alphabetical order

### LINKED ISSUE #3993: Generated README/SKILL flag examples pick the first command alphabetically, which can be a mutating POST
## Summary

Generated README/SKILL boilerplate illustrates the generic flags (`--select`, `--json`, `--dry-run`, `--agent`, `--profile`) using what appears to be the **first command alphabetically**. When that command mutates, the shipped documentation instructs the reader to perform a write.

## Concrete case

For a Jira Service Management CLI, the alphabetically first command is `servicedeskapi add-customers` — a POST that adds customers to a service desk. The generated docs contained:

```
SKILL.md:242   jsm-pp-cli servicedeskapi add-customers mock-value --agent --select id,name,status
SKILL.md:513   jsm-pp-cli --profile briefing servicedeskapi add-customers mock-value
README.md:673-685  (5 more lines, same shape)
```

`mock-value` is a placeholder in the `<serviceDeskId>` slot. Substituting a valid desk id — which is what a reader adapting the example does — performs a real write.

This was confirmed empirically: executing the documented example reached the live tenant and returned `HTTP 404` only because `mock-value` is not a valid service desk id.

## Why this matters more for printed CLIs than for typical docs

These artifacts are written for agents. `SKILL.md` and `AGENTS.md` are consumed by an LLM that is being told "here is how you use this tool's flags". A mutating command in that slot is materially more likely to be executed than an equivalent example in human-facing documentation.

## Suggested fix

Select the boilerplate example command from read-only operations only — prefer a GET with no required positional argument. Fall back to the alphabetical pick only when the CLI has no read-only command at all, and mark it clearly if so.

## Environment

cli-printing-press 4.30.1.

paonessa:
Fixes #3993.

## The bug

`firstCommandExample` already tried to prefer read-only endpoints:

```go
preferredVerbs := []string{"list", "get", "search", "query"}
...
if ep, ok := r.Endpoints[verb]; ok { ... }
```

That is an **exact name match**. Real specs name endpoints with the verb as a
prefix — `get-service-desks`, `list-users`, `add-customers` — so almost nothing
matches, the preference silently does not apply, and the search falls through to
"alphabetically first endpoint of the alphabetically first resource".

Whether the docs then advertise a mutation is a coin flip. For a Jira Service
Management CLI it lands on `add-customers`, a POST that adds customers to a
service desk, rendered into the blocks that illustrate `--select`, `--json`,
`--dry-run`, `--agent` and `--profile`:

```
jsm-pp-cli servicedeskapi add-customers mock-value --agent --select id,name,status
jsm-pp-cli --profile briefing servicedeskapi add-customers mock-value
```

`mock-value` sits in the `<serviceDeskId>` slot. Substituting a real id — which
is what a reader adapting the example does — performs a write. Confirmed by
running it: reached the live tenant, returned 404 only because `mock-value`
is not a valid desk id.

This matters more than an ordinary docs nit because `SKILL.md` and `AGENTS.md`
are consumed by agents. A mutating command in the "here is how you use the
flags" slot is materially more likely to be executed than the same example in
human-facing prose.

## The fix

A middle pass that selects by **transport method** rather than by name:

1. endpoint named exactly for a read-only verb (existing, now also honouring `Mutation`)
2. **any read-only endpoint** — `GET`/`HEAD` and not `Mutation` — alphabetically
3. alphabetically first endpoint (existing fallback)

Pass 3 is now the only path on which a mutation can reach an example, and it is
reached only when the spec has no read-only endpoint at all. Returning empty
there would delete the docs block, so falling back remains correct.

New `isReadOnlyEndpoint` also respects `spec.Endpoint.Mutation`, so a GET action
the spec author flagged as having side effects is not treated as safe.

## Verification

Four tests, written failing first:

- verb-prefixed names — reproduces `servicedeskapi add-customers mock-value` exactly
- cross-resource — a GET in `zzz-reads` beats a POST in `aaa-writes`
- mutation-only spec — still returns something rather than deleting the block
- explicit `Mutation: true` on a GET — not chosen

Regenerating the real JSM CLI: the `--select` and `--profile` examples move from
`servicedeskapi add-customers mock-value` to `servicedeskapi get-all-request-types`,
and zero flag-illustration examples reference a mutation. The genuine
command-reference entries for `add-customers` are untouched.

Failing-test sets on this host are **identical before and after** (9 pre-existing
docs/skill failures, unrelated). `go vet` clean.

### DIFF
--- a/internal/generator/first_command_example.go
+++ b/internal/generator/first_command_example.go
@@ -51,14 +51,39 @@ func firstCommandExample(resources map[string]spec.Resource) string {
 		return strings.Join(parts, " ")
 	}
 
+	// Pass 1: an endpoint named exactly for a read-only verb.
 	for _, rName := range resNames {
 		r := resources[rName]
 		for _, verb := range preferredVerbs {
-			if ep, ok := r.Endpoints[verb]; ok {
+			if ep, ok := r.Endpoints[verb]; ok && isReadOnlyEndpoint(ep) {
 				return pathFor(rName, r, verb, ep)
 			}
 		}
 	}
+	// Pass 2: any read-only endpoint, by transport method rather than by name.
+	//
+	// Pass 1 matches endpoint names exactly, which most real specs never
+	// satisfy — they name endpoints with the verb as a prefix
+	// ("get-service-desks", "list-users"). Without this pass the search fell
+	// straight through to "alphabetically first", which is a coin flip on
+	// whether the docs advertise a mutation. For a Jira Service Management CLI
+	// it produced `add-customers`, a POST, in the blocks that illustrate
+	// --select / --json / --dry-run / --agent / --profile.
+	//
+	// These blocks are read by agents. A mutating command in the "here is how
+	// you use the flags" slot is materially more likely to be executed than the
+	// same example in human-facing prose.
+	for _, rName := range resNames {
+		r := resources[rName]
+		for _, eName := range sortedEndpointNames(r.Endpoints) {
+			if ep := r.Endpoints[eName]; isReadOnlyEndpoint(ep) {
+				return pathFor(rName, r, eName, ep)
+			}
+		}
+	}
+	// Pass 3: the spec has no read-only endpoint at all. Returning empty would
+	// delete the docs block, so fall back — this is the only path on which a
+	// mutation can appear in an example.
 	for _, rName := range resNames {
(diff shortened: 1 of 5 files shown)


tmchow:
@paonessa more feedback here, minor but please address

Issue #859, opened 2026-05-09
generator: lock-time manifest reconciliation against internal/client/ env reads (subclass:credential-flow-jwt)


StewartVanDusen:
**Source:** Retro from rentalworks-home generation. See [retro doc](#) (local: `~/printing-press/manuscripts/rentalworks-home/20260508-232851/proofs/20260509-150000-retro-rentalworks-home-pp-cli.md`).

## Problem

When the printed CLI's `internal/client/` reads env vars not declared in `manifest.json`'s `user_config`, the .mcpb bundle prompts users for the wrong fields.

Concrete instance: `rentalworks-home` ships with a hand-written `internal/client/auth_refresh.go` that reads `RENTALWORKS_HOME_USERNAME` and `RENTALWORKS_HOME_PASSWORD` to refresh a 5-minute JWT. The OpenAPI spec advertises `securitySchemes: bearer`, the parser correctly classifies as `bearer_token`, and the manifest emits a single `rentalworks_home_token` user_config field. Users install the .mcpb, fill in the token, and the MCP session dies after 5 min with no recovery — because USERNAME/PASSWORD aren't in `user_config`, they aren't passed into the MCP server's process env, and `tryReauthFromCredentials` fails fast.

## Why a machine fix

The OpenAPI spec rarely models the login endpoint that issues the bearer token, so parse-time inference can't reliably catch this. But the **post-generation hand-written code** is an unambiguous signal: any `os.Getenv("<PREFIX>_USERNAME")` or `os.Getenv("<PREFIX>_PASSWORD")` (or any env read not already declared in `user_config`) tells the lock pass exactly what the manifest is missing.

Lock-time scan, not parse-time inference. Purely additive — APIs without credential-flow auth produce no extra fields.

## Proposed fix

In `internal/pipeline/climanifest.go` (or a new sibling), after the manifest is composed but before the lock pass writes it to staging:

1. Walk `internal/client/*.go` for `os.Getenv` calls.
2. For each env var name read, check if it's declared in `m.UserConfig` (mapped through `mcp_config.env`).
3. For any env read that isn't declared, add a sensitive `string` `user_config` field (snake_case the env name) and a matching `${user_config.*}` entry in `mcp_config.env`.

Call site lives near the existing `// Refresh the MCPB manifest.json in the staging dir` block in `internal/pipeline/lock.go:273`.

## Acceptance criteria

[...]

PR #4006, opened 2026-08-07
fix(cli): make default User-Agent overridable via <ENV_PREFIX>_USER_AGENT

wilatallah1-cloud:
## Intent

Issue: N/A (no existing issue filed for this)

Generated clients set a hardcoded default `User-Agent: <name>-pp-cli/<version>` with no way to
override it at runtime. Some APIs sit behind WAFs (Akamai here) that specifically block that
string shape, and there is no config.toml field or existing env var for a header override, so
the generated CLI is permanently stuck once an operator's default UA gets blocklisted.

Concrete case: `espn-pp-cli` (printed from the public library) got HTTP 403 "Access Denied" from
Akamai on every `site.api.espn.com` endpoint. Isolated with curl against the live API just now:

```
curl -A 'espn-pp-cli/0.1.0' https://site.api.espn.com/apis/site/v2/sports/football/nfl/scoreboard   -> 403
curl (default UA, no -A)                                                                             -> 200
curl -A 'Go-http-client/1.1'                                                                          -> 200
curl -A 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 ... Chrome/131.0.0.0 ...'       -> 403
```

The counterintuitive part: a spoofed real-browser Chrome UA gets blocked too, harder than an
honest non-browser client. Akamai appears to be penalizing the mismatch between a *claimed*
browser and Go's actual TLS/header fingerprint more than it penalizes a client that doesn't
pretend to be a browser at all. So "swap in a realistic browser UA" is not a safe general fix;
an override that the operator controls is.

## Approach

Add a `<ENV_PREFIX>_USER_AGENT` env var override (same `envName` prefix convention already used
for `<ENV_PREFIX>_BASE_URL`, `<ENV_PREFIX>_CONFIG`, etc.), checked in this order: explicit
per-request header (already set) wins, then the env var, then the existing baked-in default.
Applied to the three templates that set a default User-Agent: `client.go.tmpl` (main request
path), `auth_browser.go.tmpl` (`validateComposedAuth`), and `deliver.go.tmpl` (webhook delivery).
The baked-in default string itself is unchanged for every API except the one that surfaced this
(see Scope) — this PR only adds the escape hatch.

While testing this against a real printed CLI, I found the MCPB manifest reconciler
(`internal/pipeline/client_env_scanner.go`, from #859) scans `internal/client/*.go` for
`os.Getenv` calls and blanket-marks every discovery `Sensitive: true` + `Required` mirroring the
CLI's auth type. That's correct for the hand-written credential-refresh code it was built for,
but the new `*_USER_AGENT` read isn't a credential — it was getting flagged as a required secret
in the generated MCPB manifest, which would make Claude Desktop / any MCP host demand a
"credential" that doesn't exist and isn't secret. Added `isNonCredentialDiscoveredEnvVar` so
`*_USER_AGENT` discoveries are always optional and non-sensitive, with a description that says
so instead of the credential-refresh boilerplate.

## Scope

Primary area: `internal/generator/templates/` (client.go.tmpl, auth_browser.go.tmpl,
deliver.go.tmpl) and `internal/pipeline/client_env_scanner.go`.

Why this belongs in this repo: the missing override is a machine-level gap — any API behind a
UA-sensitive WAF hits the exact same dead end (no config field, no env var, hardcoded string).
The ESPN case is the symptom, not the fix target; nothing ESPN-specific is generalized into the
machine. (Separately, I patched the already-printed `espn-pp-cli` in the public library repo to
change its own baked-in default and pick up this new override — that's a printed-CLI-only change
and does not belong in this PR.)

## Risk

Low. Purely additive: when the env var is unset, behavior and generated output are byte-identical
to before (confirmed via golden diff — see Output Contract). The manifest reconciler change only
changes classification for a name suffix (`_USER_AGENT`) that did not previously exist in any
generated client, so it cannot change behavior for any already-published CLI's manifest.

## Output Contract

Changes generated `internal/client/client.go` (adds the env var check inline) and, for CLIs whose
auth requires a credential, adds a `<name>_user_agent` entry to the generated MCPB
`manifest.json` (optional, non-sensitive). Updated the 6 golden fixtures this touches via
`scripts/golden.sh update` and hand-reverted every file in the regenerated set whose diff wasn't
caused by this change (this dev machine's golden run also drifts on ~16 pre-existing,
unrelated cases: CRLF-vs-LF checksum differences and Windows path separators in normalized
output — confirmed present on a clean `git stash` baseline before this PR's changes, not
introduced by it).

## Verification

- `go build ./...`
- `go test ./...` (full suite; passes except pre-existing Windows-environment failures verified
  unrelated to this change via a clean-baseline comparison: `TestGenerateWithNoAuth`
  (govulncheck flags local Go toolchain stdlib CVEs, not code), `TestClaimOutputDir_PermissionError`
  / `TestClassifyRejectsTraversal` (Windows permission/path-separator semantics), narrativecheck
  tests, and a PATH-lookup failure in a sync test — all reproduce identically on `main` before
  this branch)
- `go vet ./internal/generator/... ./internal/pipeline/...`
- `scripts/golden.sh verify` / `scripts/golden.sh update` (see Output Contract)
- New tests added and passing: `TestGenerateUserAgentEnvVarOverridesDefault`
  (`internal/generator/generator_test.go`) and a new subtest in
  `TestReconcileMCPBManifestFromClient` (`internal/pipeline/client_env_scanner_test.go`)
- Live verification against the real ESPN API using a locally rebuilt `espn-pp-cli` with this
  fix applied to its own printed source: NFL, English Premier League (eng.1), and Turkish Süper
  Lig (tur.1) scoreboards all returned real live data instead of 403 once the override was wired
  up; confirmed the override itself (not just "it stopped erroring") by pointing `ESPN_BASE_URL`
  at an httpbin echo endpoint and reading back the actual `User-Agent` header sent, with and
  without `ESPN_USER_AGENT` set.

## AI / Automation Disclosure


tmchow:
Maintainer repair in progress for the actionable Greptile source-comment feedback and the current merge conflict. The repair will preserve the PR intent and current main behavior, using additive commits only.

PR #4013, opened 2026-08-07
fix(dogfood): parse backslash-continuation example blocks as one command

veltri-23:
### Problem

`liveDogfoodExampleArgs` in `internal/pipeline/live_dogfood.go` parses the
help Examples section **one line at a time**. Generated printed CLIs emit
examples with backslash-newline continuations:

```text
Examples:
  cli teach-pattern \
    --query-template "items in {entity}" \
    --resource-template "GROUP-{entity:category}" \
    --strategy substitute
```

Parsing only the first line yields `["teach-pattern", "\\"]` — a stray
backslash token and none of the real flags. Commands that require flags
then fail the live dogfood happy path with a misleading "flag is required"
error, and the full matrix can never certify acceptance for any CLI whose
example is continuation-shaped.

Observed in the field: `teach-pattern` failed `dogfood --live` on freshly
printed CLIs with `Error: --query-template is required` even though the
example clearly carries the flag.

### Change

Parse the example block as a single shell command before falling back to
per-line parsing:

- Join example lines into one command string. Lines ending in a backslash
  are joined with a newline (shellargs folds `\` + newline exactly like
  the real help text; a space would be treated as an escaped literal and
  stick to the next token).
- Non-continuation lines stay separated so help text with multiple
  distinct example commands still resolves each one.
- The per-line loop remains as a fallback for multi-example help blocks.

### Tests

- `TestLiveDogfoodExampleArgsJoinsBackslashContinuations` — a
  continuation-shaped example resolves to the full flag set.
- `TestLiveDogfoodExampleArgsSeparatesDistinctExamples` — two distinct
  example commands in one block still resolve to the matching one.

### Validation

- `go test ./internal/pipeline/ -run 'TestLiveDogfoodExampleArgs'` passes.
- The full package suite on this machine has pre-existing environment
  failures unrelated to this change (verified by running the suite on the
  unmodified tree); the live-dogfood tests themselves pass.
- The fix was exercised end to end: freshly printed CLIs whose
  teach-family examples were continuation-shaped passed `dogfood --live`
  after applying this change.

### Notes

No API-specific names in the fix; applies to every generated CLI. No
behavioral change for examples that are already single-line.

### DIFF
--- a/internal/pipeline/live_dogfood.go
+++ b/internal/pipeline/live_dogfood.go
@@ -2165,39 +2165,35 @@ func liveDogfoodHappyArgsParsed(command liveDogfoodCommand) ([]string, bool, hap
 
 func liveDogfoodExampleArgs(command liveDogfoodCommand) ([]string, bool) {
 	examples := extractExamplesSection(command.Help)
-	// Parse the example block as a single shell command instead of one line
-	// at a time. Generated examples use backslash-newline continuations:
-	// shellargs.Split folds them, but a per-line parse of the first
+	// Split the example section into logical commands: backslash-newline
+	// continuations fold into the command they belong to; any other line
+	// boundary starts a new candidate. Generated examples use
+	// backslash-newline continuations, so a per-line parse of the first
 	// continuation line yields a stray "\" token as a positional and drops
-	// the real example flags, failing commands that require flags (e.g.
-	// teach-pattern --query-template).
-	var sb strings.Builder
+	// the real example flags (e.g. teach-pattern --query-template). Parsing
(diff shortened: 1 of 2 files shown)


veltri-23:
Hey @tmchow. #4013 is fully green: all CI checks pass and Greptile is 5/5 with no remaining findings. The fix splits distinct example blocks so a target-first example no longer merges later examples arguments, with regression tests covering the reported case. Tagging for visibility in case it came in wrong the first time.

The Label and Title Mergify is asking for its physically not letting me change for whatever reason so i hope you can? TIA

PR #4026, opened 2026-08-08
fix(cli): prevent silent pagination truncation

### LINKED ISSUE #3989: generator: fabricated defaultPageSize (100) is used as the end-of-pagination signal, so --all silently returns one page
Generator **v4.30.1**. Reproduced on a pristine `generate --force --lenient` tree.

## Summary

When the spec does not expose a limit/page-size query parameter, the generator
**invents** a page size and then uses that invented number as the
end-of-pagination signal. If the server's real page is smaller than the guess,
`--all` stops after the first page and reports success. No warning, no error,
no truncation notice — just a short answer that looks complete.

## Reproduce

Spec: a page-numbered list endpoint whose page size is fixed by the server and
whose `pagesize` parameter is ignored (documented as such in the operation
description).

```
cli-printing-press generate --spec <spec> --output <dir> --force --lenient
grep -n "resolvePaginatedReadWithStrategy" <dir>/internal/cli/promoted_category.go
```

Emitted call site:

```go
resolvePaginatedReadWithStrategy(cmd.Context(), c, flags, "auto", "category", path,
    map[string]string{"pagenumber": formatCLIParamValue(flagPagenumber)},
    nil, flagAll, "pagenumber", "page", "", 100, "", "", "products", cmd.ErrOrStderr())
                                              ^^^ defaultPageSize
```

`100` appears nowhere in the spec. The real page is 30.

## Mechanism

`internal/cli/helpers.go`:

```go
if pageSize <= 0 {
    pageSize = 100          // fabricated
}
```

and the stop condition, `nextFullPageOffsetCursor`:

```go
if pageSize <= 0 || itemCount < pageSize {
    return "", false        // "short page" == last page
}
```

Page 1 returns 30 items; `30 < 100`, so the paginator concludes it has reached
the end. `--all` returns 30 of 103 and exits 0.

The failure mode is asymmetric and that is what makes it dangerous: guessing
**too high** truncates silently, while guessing too low would merely cost extra
requests. A wrong guess should not be able to look like success.

## Why the spec could not prevent it

The operation description states the constraint in plain words ("fixed page of
30 products; the `pagesize` parameter is ignored"), and it is not read. There is
also no declarative escape hatch: `x-pp-pagination` accepts only `none`, which
is wrong here — the endpoint *is* paginated, it just has a server-fixed page
size that the client cannot influence.

I hit the same thing on an unrelated target the day before (real page 24,
emitted 100 → `--all` returned 24 of 248), so it is not spec-specific. Both
times the only reason it was caught was comparing against a pre-existing scraper
that returned the true count.

## Suggested fixes (either would have caught it)

1. **Do not use a fabricated page size as an end-of-collection signal.** When
   the page size was guessed rather than observed, keep requesting until a page
   returns **zero** items, or until the response's own total/pager says stop.
2. **Accept a declared page size** — e.g. `x-pp-pagination: {size: 30, fixed: true}`
   or an `x-pp-page-size` operation extension — so an author who has *measured*
   the page can say so.

At minimum, emit a warning at generation time when `defaultPageSize` is a
fallback rather than derived from the spec.

### LINKED ISSUE #4004: Generator emits empty nextCursorPath, so --all silently truncates to page one
Split out from #4003, which fixes a different bug in the same function.

## Problem

When the generator cannot infer the response field holding the next cursor, it emits `nextCursorPath: ""` at the call site:

```go
}, nil, flagAll, "after", "cursor", "limit", "", "", cmd.ErrOrStderr())
```

With no cursor path, the fetch-all loop has no way to advance and breaks after the first page. `--all` then returns page one **and reports it as the complete set** — no warning, no non-zero exit. A caller cannot tell a full result from a truncated one, which is the failure mode most likely to produce a wrong answer downstream.

## Not covered by the existing fallback

Current `main` falls back to the cursor param name when the path is empty:

```go
cursorLookupPath := nextCursorPath
if cursorLookupPath == "" && paginationType != "offset" && paginationType != "page" {
    cursorLookupPath = cursorParam
}
```

That only helps APIs echoing the cursor at top level. DocuSeal nests it:

```json
{ "data": [...], "pagination": { "count": 3, "next": 9, "prev": 11 } }
```

The correct path is `pagination.next`; `after` is absent, so the fallback misses and the loop still stops at page one. Setting `nextCursorPath: "pagination.next"` by hand fixes it: `templates list --all --limit 3` goes from 3 rows to all 9, and an exhausted page answers `pagination.next: null`, so the loop terminates on its own.

## Suggestions

1. Widen inference to common nested shapes (`pagination.next`, `meta.next_cursor`, `page_info.end_cursor`).
2. More important than any inference: **fail loud when the path is unknown.** If `--all` cannot advance, it should warn on stderr or refuse the flag, rather than returning a partial set that looks complete.

tmchow:
## Summary

- infer nested integer `pagination.next` paths so cursor-based `--all` requests follow the server continuation token
- remove the fabricated page-size fallback and use observed item counts for unknown-size page and offset pagination
- keep explicit `has_more` envelopes advancing unknown-size offsets, and update generated regressions and golden fixtures

## Validation

- `go test ./internal/openapi -count=1`
- focused pagination regressions in `internal/generator` and `internal/openapi`
- `go test ./... -run '^$' -count=1 -parallel=1 -p=1`
- `scripts/golden.sh verify` passed all 32 cases
- `scripts/verify-generator-output.sh` passed all 10 cases
- the full `internal/generator` test run was attempted with serialized execution but timed out in local generated-suite execution; focused tests and generated-output compilation pass

Closes #4004
Closes #3989

### DIFF
--- a/internal/generator/generator.go
+++ b/internal/generator/generator.go
@@ -5081,30 +5081,22 @@ func (g *Generator) renderPromotedCommandFiles(promotedCommands []PromotedComman
 	return nil
 }
 
-func (g *Generator) paginationDefaultPageSize() int {
-	if g != nil && g.profile != nil && g.profile.Pagination.DefaultPageSize > 0 {
-		return g.profile.Pagination.DefaultPageSize
-	}
-	return 100
-}
-
 func (g *Generator) paginationPageSizeForEndpoint(endpoint spec.Endpoint) int {
-	pageSize := g.paginationDefaultPageSize()
+	// Only an endpoint-declared default is safe as short-page evidence. Zero
+	// tells the generated helper that the server's page size is unknown.
 	if endpoint.Pagination == nil || endpoint.Pagination.LimitParam == "" {
-		return pageSize
+		return 0
 	}
 	limitParam, ok := endpointParamByName(endpoint, endpoint.Pagination.LimitParam)
 	if !ok {
-		return pageSize
+		return 0
 	}
+	pageSize := 0
 	if defaultPageSize, ok := positiveIntValue(limitParam.Default); ok {
 		pageSize = defaultPageSize
 	}
-	if maxPageSize, ok := paramMaxInt(limitParam); ok && maxPageSize < pageSize {
+	if maxPageSize, ok := paramMaxInt(limitParam); ok && pageSize > 0 && maxPageSize < pageSize {
 		pageSize = maxPageSize
-	}
-	if pageSize <= 0 {
-		return g.paginationDefaultPageSize()
 	}
 	return pageSize
 }
--- a/internal/generator/paginated_truncate_test.go
+++ b/internal/generator/paginated_truncate_test.go
@@ -10,6 +10,7 @@ import (
 	"github.com/mvanhorn/cli-printing-press/v4/internal/openapi"
 	"github.com/mvanhorn/cli-printing-press/v4/internal/profiler"
 	"github.com/mvanhorn/cli-printing-press/v4/internal/spec"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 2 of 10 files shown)


tmchow:
Greptile follow-up disposition:

- The unknown-size offset pagination finding is addressed in commit 0f7e0cdc. The generated runtime regression covers an empty page with has_more=true, and the generator now advances using the last observed page size or a one-record fallback.
- The generated golden fixture patch-record finding is accepted-for-later/non-blocking. These files are deterministic machine-generated expected outputs, not downstream printed-CLI customizations; adding .printing-press-patches records here would misclassify the fixture contract.

The second item is intentionally documented rather than applied.

PR #4062, opened 2026-08-10
fix(cli): honor generated sync pagination contracts

### LINKED ISSUE #4034: generator: sync pagination derives neither the cursor parameter nor the cursor path from the spec (silent single-page sync)
> Filed by `/printing-press-retro` from a post-publish PR-review session on **respondio** (Press 4.30.1, scorecard 95/100 Grade A, verify 100% 55/55). Cross-CLI evidence from `quentli`, `google-tag-manager`, and `livechat`.

## Summary

The generated `sync` cannot paginate unless the API happens to name its cursor parameter `after` **and** return the cursor as a JSON string under one of eleven hardcoded key names. Two independent gaps compound; fixing either alone changes nothing.

**This is not #3538.** That issue (`cursorParam` without `cursorType`) closed COMPLETED on 2026-08-02. `respondio` was generated **2026-08-05 on Press 4.30.1** and still exhibits the defect below. Same family, distinct variant, post-dates the fix.

## Where to look

- **Component:** generator
- **Likely area:** the sync template's `determinePaginationDefaults`, `extractPaginationFromEnvelope`, and `findCursorInMap`
- **Triggered when:** the spec declares any cursor parameter not literally named `after`, or nests the cursor under a bare `next` key, or returns a numeric cursor

## What we observed

**Gap 1 — the cursor parameter name is never derived from the spec.** Every `case` arm and the fallback return the literal `"after"`:

| CLI | Spec cursor param | Emitted `cursorParam` |
|-----|-------------------|-----------------------|
| respondio | `cursorId` (int) | `"after"` |
| google-tag-manager | `pageToken` (string) | `"after"` |
| livechat | `page_id` (string) | `"after"` |

The profiler *has* the right name — it emits it correctly into `syncResourceBodyParamWireNames` for the same resource. The pagination config just doesn't consume it. The API ignores the unknown parameter and returns page 1 forever.

**Gap 2 — the extractor can't read the cursor.** `cursorKeys` lists `next_cursor`, `nextCursor`, `next_token`, `nextToken`, `cursor`, `next_page_token`, `nextPageToken`, `page_token`, `after`, `end_cursor`, `endCursor` — but not the bare key `next`, which is what these specs' own `cursor_field` points at. And `findCursorInMap` only returns values that unmarshal into a Go `string`, so a numeric cursor is skipped silently.

Measured across 31 CLIs with the extractor: **31/31** lack `"next"` in `cursorKeys`; **31/31** are string-only.

Three APIs declare `cursor_field` pointing at a `next` key: respondio (7 endpoints, numeric), google-tag-manager (17, string), livechat (3, string).

Probed directly against the shipped function with the real envelope shape — `{"items":[...],"pagination":{"next":12345}}` and the string form, across cursor params `after` and `cursorId` — it returned `("", false)` in **all four** combinations.

## Why this survived every gate

Silent truncation is indistinguishable from a small dataset. This CLI passed `verify` 55/55, scored 95/A, passed shipcheck, published, and went through human review of a 282-file PR. It was caught only because an external reviewer read `resourceSupportsPagination` and noticed one resource missing from a switch. The woocommerce retro recorded the same blind spot: "dogfood/live matrix can't see under-fetch — sync 'completes' successfully."

## Proposed fix

Have the sync template consume the profiler's parsed pagination block directly — the parameter name and `nextCursorPath` are both already emitted elsewhere in the same CLI (`resourceReadConfigs` carries `nextCursorPath: "pagination.next"`). Failing that: append `"next"` to `cursorKeys` **last** so more specific names keep priority, and accept JSON numbers, rendering integers without decimal point or exponent.

Where the spec declares no cursor parameter, emit no cursor pagination rather than guessing `"after"`.

## Acceptance criteria

- positive: a spec declaring `cursorId` emits `cursorParam: "cursorId"`; one declaring `pageToken` emits `"pageToken"`; `{"pagination":{"next":12345}}` and `{"pagination":{"next":"12345"}}` both extract `"12345"` with hasMore true
- negative: a spec with no declared cursor emits no cursor pagination and does not fall back to `"after"`; an envelope with no pagination object, and one with `{"pagination":{}}`, both yield `""`/false; a JSON:API `links.next` URL still resolves through the existing links path rather than being captured as a literal cursor

## Scope boundary

Does not include the command-level `--all` helper (filed separately) and does not change offset/page paginator arithmetic (#3538's territory).

## Related issues

- #3538 — closed COMPLETED 2026-08-02; same family, this variant post-dates and survives that fix
- #3435 — `paginatedGet` envelope normalization, adjacent territory
- #3481 — page-int-aware cursor-missing warning, adjacent territory

---

**Labels requested** (the filing account has read-only access and cannot apply them): `source:retro` · `bug` · `priority:P1` · `comp:generator`

### LINKED ISSUE #4035: generator: --all is registered on POST-based list endpoints where it can never be wired, shipping a flag that silently does nothing
> Filed by `/printing-press-retro` from a post-publish PR-review session on **respondio** (Press 4.30.1). Cross-CLI evidence from `google-tag-manager` and `livechat`.

## Summary

`--all` is registered on every list-shaped endpoint, but the pagination helper it wires into (`paginatedGet`) requires a client interface with `GetWithHeaders` — it is GET-only. For a POST-based list endpoint the generator emits the flag and omits the wiring, with no error and no warning. The result is a flag that is accepted, documented in `--help` as "Fetch all pages," and does nothing.

## Where to look

- **Component:** generator
- **Likely area:** list-command emission and the `paginatedGet` helper template
- **Triggered when:** a list endpoint uses POST (common for search/filter endpoints that take a request body)

## What we observed

Four commands across three CLIs, all `pp:method: POST`, all with the flag declared and registered but never read:

| CLI | Command file |
|-----|--------------|
| google-tag-manager | `internal/cli/folders_entities.go` |
| livechat | `internal/cli/chats_list.go` |
| livechat | `internal/cli/threads_list.go` |
| respondio | `internal/cli/contact_list.go` |

Detection is mechanical: each of these files contains exactly 2 references to the flag variable (declaration + registration), while every correctly wired GET command has 3 or more.

A user or agent who passes `--all` receives one page and no indication the result is partial.

## Proposed fix

**Preferred:** emit a POST-capable pagination helper so `--all` wires for both verbs.

**Minimum acceptable:** do not register `--all` when the generator cannot wire it. A missing flag is strictly better than a dead one.

Either way, add a generation-time assertion that every registered flag variable is actually read — this class of defect is invisible without reading the generated source, and the assertion catches the next instance for free.

## Acceptance criteria

- positive: a POST-based list endpoint with a declared cursor either paginates under `--all` or does not offer `--all`; a generation-time assertion fails when any registered flag variable is never read
- negative: GET-based list commands keep `--all` and behave exactly as today

## Scope boundary

Does not include retrofitting the four already-shipped CLIs above — those are printed-CLI fixes.

## Dependencies

A POST-capable helper depends on the corrected cursor derivation in #4034 to be worth wiring; the "don't register what you can't wire" minimum fix has no dependency.

## Related issues

- #4034 — cursor derivation, prerequisite for the preferred fix
- #3649 — `--all` truncating when the server caps page size, adjacent territory

---

**Labels requested** (the filing account has read-only access and cannot apply them): `source:retro` · `bug` · `priority:P2` · `comp:generator`

### LINKED ISSUE #4047: Live `verify` reports "sync crashed" for large-catalogue APIs: generated sync curtails pages under dogfood but not under verify
### Summary

`verify --api-key` runs the data-pipeline probe as:

```go
// internal/pipeline/runtime.go:672
syncErr = runCLI(binary, []string{"sync", "--full"}, env, 30*time.Second)
```

Generated `sync` curtails its page walk under the dogfood env, but there is no equivalent guard for verify:

```go
// generated internal/cli/sync.go
if cliutil.IsDogfoodEnv() && !cmd.Flags().Changed("max-pages") {
    maxPages = 1
}
```

For any API whose full catalogue takes longer than 30s to walk, live verify kills the subprocess and reports `FAIL: sync crashed` for a sync that works correctly. The failure is invisible in mock mode, where the mock server answers instantly.

### Reproduction

A printed CLI against a paginated marketplace API (~162 pages of 25 records). Same binary, same command, only the env differs:

| Env | Rows synced | Elapsed |
|---|---|---|
| `PRINTING_PRESS_DOGFOOD=1` | 25 | 1s |
| `PRINTING_PRESS_VERIFY=1` | 4,035 | 74s |
| neither | 4,034 | 72s |

With `--api-key`, `verify` then reports:

```
"data_pipeline": false,
"data_pipeline_detail": "FAIL: sync crashed",
"verdict": "FAIL"
```

…while every other signal in the same report is healthy: `mode: "live"`, `total: 67, passed: 67, failed: 0, pass_rate: 100`.

### Why this matters

It makes `verdict: FAIL` unreachable-to-clear for large-catalogue APIs specifically when you do the more rigorous thing and run verify live. `--env-var` avoids it only because it never enters live mode at all (`runtime.go` selects live on `cfg.APIKey != ""`, and `--env-var` populates `cfg.EnvVar`, not `cfg.APIKey`) — so the incentive is to verify less thoroughly.

The data-pipeline probe's question is "does sync write rows that search can read". One page answers that as well as 162 do.

### Suggested fix

Extend the existing curtailment to the verifier, matching what dogfood already does:

```go
if (cliutil.IsDogfoodEnv() || cliutil.IsVerifyEnv()) && !cmd.Flags().Changed("max-pages") {
    maxPages = 1
}
```

One caveat I could not resolve from outside: `IsVerifyEnv` is documented as "running under the printing-press verifier **in mock mode**", and there is a separate `IsVerifyLiveHTTPEnv`. If `PRINTING_PRESS_VERIFY` is not set during a live verify run, the guard would need whichever variable *is* set on that path. The empirical result above shows `PRINTING_PRESS_VERIFY=1` alone does not curtail today, so at minimum the two env states disagree about page budgets.

Raising the probe timeout would also work, but scales badly — the walk time is a property of the upstream catalogue, not of the CLI.

### Environment

- cli-printing-press 4.30.1
- Go 1.26.5, darwin/arm64
- Internal YAML spec, `pagination: {type: page, cursor_param: page, limit_param: per_page}`

I have not opened a PR since the fix touches the sync template and the right env predicate depends on verifier internals I can only observe from outside. Happy to if the approach above is the one you want.

tmchow:
## Summary

Generated sync now uses each endpoint’s declared cursor parameter, cursor type, and next-cursor path, including numeric cursor values. List commands expose `--all` only where generated pagination can actually advance, so POST list endpoints no longer ship a hollow flag. Live verification bounds sync probes to one page per resource so large catalogs do not exceed the verifier budget.

## Validation

- Focused generator and pipeline pagination tests pass.
- `scripts/golden.sh verify` passes all 32 cases.
- `scripts/verify-generator-output.sh` passes all 10 cases.
- `go build -o ./cli-printing-press ./cmd/cli-printing-press` passes.
- Full `go test ./...` reaches unrelated failures under required `PRINTING_PRESS_VERIFY=1`: BLE availability expectation, generated self-learning/store/rate-limit tests, and live-dogfood help probes.
- `golangci-lint run ./...` reports one pre-existing `modernize` finding in another worktree’s `internal/googlediscovery/parser.go`.

Closes #4034
Closes #4035
Closes #4047

### DIFF
--- a/internal/generator/generator.go
+++ b/internal/generator/generator.go
@@ -4094,17 +4094,18 @@ type criticalResourceEntry struct {
 }
 
 type paginationDefaultEntry struct {
-	Key         string
-	Name        string
-	CursorParam string
-	CursorType  string
-	LimitParam  string
-	Limit       int
+	Key            string
+	Name           string
+	CursorParam    string
+	CursorType     string
+	NextCursorPath string
+	LimitParam     string
+	Limit          int
 }
 
 func paginationDefaultEntries(syncable []profiler.SyncableResource, dependent []profiler.DependentResource) []paginationDefaultEntry {
 	defaults := map[string]paginationDefaultEntry{}
-	add := func(key, name, cursorParam, cursorType, limitParam string, limit int, supportsPagination bool) {
+	add := func(key, name, cursorParam, cursorType, nextCursorPath, limitParam string, limit int, supportsPagination bool) {
 		if !supportsPagination || key == "" || name == "" {
 			return
 		}
@@ -4112,19 +4113,20 @@ func paginationDefaultEntries(syncable []profiler.SyncableResource, dependent []
 			return
 		}
 		defaults[key] = paginationDefaultEntry{
-			Key:         key,
-			Name:        name,
-			CursorParam: cursorParam,
-			CursorType:  cursorType,
-			LimitParam:  limitParam,
-			Limit:       limit,
+			Key:            key,
+			Name:           name,
+			CursorParam:    cursorParam,
+			CursorType:     cursorType,
+			NextCursorPath: nextCursorPath,
+			LimitParam:     limitParam,
+			Limit:          limit,
 		}
 	}
 	for _, resource := range syncable {
-		add(resource.Name, resource.Name, resource.PaginationCursorParam, resource.PaginationCursorType, resource.PaginationLimitParam, resource.PaginationPageSize, resource.SupportsPagination)
+		add(resource.Name, resource.Name, resource.PaginationCursorParam, resource.PaginationCursorType, resource.PaginationNextCursorPath, resource.PaginationLimitParam, resource.PaginationPageSize, resource.SupportsPagination)
 	}
 	for _, resource := range dependent {
-		add(resource.ParentResource+"/"+resource.Name, resource.Name, resource.PaginationCursorParam, resource.PaginationCursorType, resource.PaginationLimitParam, resource.PaginationPageSize, resource.SupportsPagination)
+		add(resource.ParentResource+"/"+resource.Name, resource.Name, resource.PaginationCursorParam, resource.PaginationCursorType, resource.PaginationNextCursorPath, resource.PaginationLimitParam, resource.PaginationPageSize, resource.SupportsPagination)
 	}
 
 	keys := make([]string, 0, len(defaults))
--- a/internal/generator/generator_test.go
+++ b/internal/generator/generator_test.go
@@ -15146,15 +15146,15 @@ func TestGeneratedSyncUsesPerResourcePaginationDefaults(t *testing.T) {
 	assert.Contains(t, syncContent, `pageSize := determinePaginationDefaults(resource)`)
 	assert.Contains(t, syncContent, `func determinePaginationDefaults(resource string) paginationDefaults`)
 	assert.Contains(t, syncContent, `case "assets":`)
-	assert.Contains(t, syncContent, `cursorParam: "skip"`)
-	assert.Contains(t, syncContent, `cursorType:  "offset"`)
-	assert.Contains(t, syncContent, `limitParam:  "limit"`)
-	assert.Contains(t, syncContent, `limit:       50`)
+	assert.Contains(t, syncContent, `cursorParam:    "skip"`)
+	assert.Contains(t, syncContent, `cursorType:     "offset"`)
+	assert.Contains(t, syncContent, `limitParam:     "limit"`)
+	assert.Contains(t, syncContent, `limit:          50`)
 	assert.Contains(t, syncContent, `case "photos":`)
-	assert.Contains(t, syncContent, `cursorParam: "page"`)
-	assert.Contains(t, syncContent, `cursorType:  "page"`)
-	assert.Contains(t, syncContent, `limitParam:  "per_page"`)
-	assert.Contains(t, syncContent, `limit:       25`)
+	assert.Contains(t, syncContent, `cursorParam:    "page"`)
+	assert.Contains(t, syncContent, `cursorType:     "page"`)
+	assert.Contains(t, syncContent, `limitParam:     "per_page"`)
+	assert.Contains(t, syncContent, `limit:          25`)
 
 	runGoCommand(t, outputDir, "mod", "tidy")
 	runGoCommand(t, outputDir, "build", "./internal/cli")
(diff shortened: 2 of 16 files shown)


tmchow:
Greptile P2 accepted for later: the “Record generated pagination changes” finding on generated golden sync fixtures is intentionally not being applied in this PR. These files are machine-generated golden outputs for the generator contract; adding a `.printing-press-patches/` record would describe a printed-CLI customization rather than the generator-produced fixture update. A separate follow-up can define patch-intent semantics for golden fixtures if that remains needed. This PR keeps the generated outputs synchronized with the pagination generator behavior.

PR #4108, opened 2026-08-12
fix(cli): html_extract returns real id and slug, and drops bogus site-root image

santerotribe:
## Intent

Two defects in the `html_extract` template affect every printed CLI that uses `html_extract` with `mode: links`.

**The reported `slug` is actually the id, and the real slug is dropped.** `htmlLinkSlug` returned `parts[0]` — the FIRST path segment after the matched prefix. For the near-universal detail-page shape `/<prefix>/<id>/<slug>.html` that is the opaque short id, returned under the name "slug", while the human slug is never exposed at all. Discovery output therefore cannot be chained into the detail command that needs `--id` and `--slug`.

**Links with no image report the site root as their image.** `normalizeHTMLURL` did not guard the empty string. `url.Parse("")` yields an empty reference and `ResolveReference` against the base returns the base, so any `<a>` with no discoverable `` got `image: "https://<site>/"`. The same function handles `<link rel=canonical>`, so a canonical-less page silently adopted the base URL as its canonical.

Issue: N/A

## Approach

Replaced `htmlLinkSlug` with `htmlLinkPathParts`, returning both id and slug.

The prefix test was also wrong in a way worth calling out separately: it was `rest == path`, where `rest` came from `strings.Trim(strings.TrimPrefix(path, prefix), "/")`. Because `Trim` also strips the leading slash, `rest` can never equal `path` for an absolute path, so a NON-matching prefix was silently accepted. With prefixes `["/tracklist", "/track"]`, the path `/track/<id>/<slug>/index.html` matched `"/tracklist"` and yielded `id="track", slug="index"`. The replacement gates on the existing `htmlPathMatchesPrefix` segment-boundary helper — the same matcher the extractor already uses to select links — rather than inventing a second, weaker test.

It also strips a trailing page extension so the slug can be handed straight back to a URL template, and drops a trailing directory-index segment so `/<id>/<slug>/index.html` yields the slug rather than `"index"`. Added an `ID` field to `htmlLink`.

For the image bug, `normalizeHTMLURL` now trims and returns `""` for empty input before parsing; with `omitempty` the bogus key disappears from output entirely.

## Scope

Primary area: generator templates (`internal/generator/templates/html_extract.go.tmpl`)

Why this belongs in this repo: the defective code is emitted by the template into every printed CLI using link extraction, so a printed-CLI fix would have to be repeated per CLI and would silently diverge. The symptom surfaced in two independently generated CLIs; the cause is one template.

## Risk

Changes generated output for every CLI using `html_extract` `mode: links`:

- `htmlLink` gains an `id` key. Additive, `omitempty`.
- `slug` now carries the human slug where it previously carried the id. This is the intended correction, but any consumer that was compensating for the old behaviour by treating `slug` as an id will need the `id` field instead.
- `image` is now absent rather than the site root when an anchor has no image. Consumers that assumed the key was always present should use `omitempty`-safe access.

No change to auth, MCP surface, publish flow, verifier, scorer, or release behavior.

## Output Contract

Generated-output evidence, in order of strength:

1. New `html_extract_test.go.tmpl`, emitted alongside `html_extract.go`, so every printed CLI carries assertions for its own copy. Covers the id/slug split, the dropped `index` segment, single-segment paths, an earlier similar prefix not swallowing a later one, the no-matching-prefix fallback, and the empty-URL guard.
2. Compiled generated CLI: regenerated a links-mode CLI, `go test ./internal/...` green across 12 packages.
3. Live output from that regenerated CLI over 50 rows: correct id and slug on both `/<prefix>/<id>/<slug>.html` and `/<prefix>/<id>/<slug>/index.html` shapes; 0 rows with `id == slug`; 0 rows with a prefix literal as id; 0 rows carrying an `image` key where the anchor had no image. The previously broken discovery-to-detail chain works end to end.

`scripts/golden.sh verify` passes 32 cases with no golden churn, since no existing golden fixture covers `mode: links` output.

## Verification

- `go build ./...`
- `go vet ./internal/generator/`
- `go test ./internal/generator/` — ok, 354s, 0 failures
- `./scripts/golden.sh verify` — passed, 32 cases
- Regenerated a links-mode CLI with the patched binary, ran its emitted tests (4 new tests pass, 12 packages ok), and checked live extraction output as described above

## AI / Automation Disclosure

The change was authored and verified by an AI agent at a human's direction; the human directed the fix and approved opening this PR but has not line-reviewed the diff.


santerotribe:
Withdrawing this — we're not pursuing the upstream change. Closing to keep the queue clean; no review needed. The analysis in the description stands if it's ever useful to anyone.

PR #4231, opened 2026-08-17
fix(cli): preserve null analytics groups

### LINKED ISSUE #4098: analytics --group-by emits Go's "<nil>" string for null group values (seats-aero-pp-cli)
## What happens
`seats-aero-pp-cli analytics --type routes --group-by source --agent` emits the literal Go nil string as a JSON value: `[{"value":"\u003cnil\u003e","count":200}]` (i.e. `"<nil>"`). When the group-by key is null for the rows, the CLI `fmt`-stringifies a nil interface into the output instead of emitting JSON `null` (or omitting/labelling it).

## Impact
A `<nil>` string is not valid domain data — an agent parsing the group values gets a sentinel that looks like a real category. It also indicates the grouped field (`source`) is null for those rows after a `sync --full`, which may be a separate population gap.

## Expected
Null group keys should render as JSON `null` (or an explicit "(none)" label), never Go's `<nil>` string.

## Repro
`seats-aero-pp-cli analytics --type routes --group-by source --agent` on a store where `source` is null for some routes.

CBOSSX:
## Intent

Preserve null semantics in generated `analytics --group-by` output. Explicit null values and missing fields currently become the string `"<nil>"`, so agents can mistake absent data for a real category.

Issue: Fixes #4098

## Approach

Track null and missing values in a separate count instead of passing them through the string-keyed group map. Generated machine output now encodes that group as JSON `null`, while text output labels it `(none)`. The generated-CLI regression covers both an explicit null field and an absent field.

## Scope

Primary area: generated analytics command

Why this belongs in this repo: `analytics.go.tmpl` is the shared Printing Press template that emits this behavior into every store-enabled printed CLI.

## Risk

Low. Non-null group values retain their existing string representation and sorting behavior. The only output-contract change is the null bucket: JSON `"<nil>"` becomes `null`, and text `<nil>` becomes `(none)`.

## Output Contract

Generated-output evidence: `TestProjectManagementWorkflowsEmitSyncHints` generates and compiles a CLI, seeds both explicit-null and missing-field rows, and asserts JSON `null` plus the `(none)` text label. The full golden suite and generated-output compile suite also pass without fixture changes.

## Verification

- `go test ./internal/generator -run '^TestProjectManagementWorkflowsEmitSyncHints$' -count=1`
- `go test -timeout 30m -p 1 ./...`
- `go fmt ./...`
- `scripts/golden.sh verify` (32 cases)
- `scripts/verify-generator-output.sh` (10 cases)
- `GOTOOLCHAIN=go1.26.6 go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.4 run ./...` (0 issues)

## AI / Automation Disclosure

### DIFF
--- a/internal/generator/generator_test.go
+++ b/internal/generator/generator_test.go
@@ -19674,6 +19674,8 @@ func TestAnalyticsCommandWritesToConfiguredOutput(t *testing.T) {
 	for _, raw := range []string{
 		` + "`" + `{"id":"issue-1","status":"open","stage_id":"qualified"}` + "`" + `,
 		` + "`" + `{"id":"issue-2","status":"open","stage_id":"qualified"}` + "`" + `,
+		` + "`" + `{"id":"issue-3","status":"closed","stage_id":null}` + "`" + `,
+		` + "`" + `{"id":"issue-4","status":"closed"}` + "`" + `,
 	} {
 		var payload json.RawMessage = []byte(raw)
 		var obj map[string]any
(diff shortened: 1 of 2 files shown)


CBOSSX:
Closing this PR because the shared controller run had already reached its configured max_prs_per_run=1 through a concurrent publication before this PR was created. The branch is preserved, but this automated run must not submit a second contribution.

PR #4350, opened 2026-08-25
fix(generator): count binary/text resources toward the sync-stub decision

### LINKED ISSUE #4342: generator: non-JSON resources land in defaultSyncResources(), so bare sync writes a NULL row, stamps sync_state, and reports success over an empty store
**Component:** generator — `sync` emission for resources whose `response_format` is not JSON (`html`, `binary`, `text`)

**Surfacing CLI:** `extron` (devices). Reproduced independently on `averusa`. Filing the generator-level defect.

## What's broken

When a spec resource declares `response_format: html` (with `html_extract`), the generator still emits it into `defaultSyncResources()`. Bare `sync` then "succeeds" against a body the generated client cannot map: it writes a row whose `data` is **NULL**, stamps `sync_state`, and exits 0 with `errored: 0`. The store reads as synced and returns nothing.

Reproduced on a **completely fresh home** with the published `extron-pp-cli 2026.8.1`:

```console
$ extron-pp-cli --home /tmp/fresh sync
{"event":"sync_start","resource":"literature"}
{"event":"sync_complete","resource":"literature","total":1,"duration_ms":154}
{"event":"sync_summary","total_records":1,"resources":1,"success":1,"warned":0,"errored":0,"duration_ms":154}

$ sqlite3 data.db "select resource_type, id, data from resources"
literature|literature|          <-- data IS NULL

$ sqlite3 data.db "select * from sync_state"
literature||2026-08-24T20:29:56Z|1

$ extron-pp-cli --home /tmp/fresh doctor
  OK Cache: fresh
```

So on a brand-new install the user is told **1 record synced, 0 warnings, 0 errors**, `hintIfUnsynced` goes quiet because `sync_state` is stamped, and **`doctor` reports "Cache: fresh" over an empty catalog**. Every local read then returns nothing, with no signal anywhere that the sync did not work.

`library/devices/extron/spec.yaml:36-59` is an ordinary declaration — nothing malformed:

```yaml
  literature:
    endpoints:
      index:
        method: GET
        path: "/technology/literature.aspx"
        response_format: html
        html_extract:
          mode: links
          link_prefixes: ["/download/files"]
```

## Two distinct shapes, both ending in exit 0

`averusa` (same class, different symptom) does emit a signal — but does not count it:

```console
$ averusa-pp-cli --home /tmp/fresh2 sync
{"event":"sync_anomaly","resource":"docs","reason":"non_json_200_body"}
{"event":"sync_complete","resource":"docs","total":0,"duration_ms":1790}
{"event":"sync_summary","total_records":0,"resources":2,"success":2,"warned":0,"errored":0}
```

`sync_anomaly` fires, yet the run reports `success: 2, warned: 0, errored: 0` and still stamps `sync_state` for both resources. So:

- **extron:** no anomaly at all; a NULL row is *counted as a record*; `doctor` says fresh.
- **averusa:** anomaly emitted but not promoted to a warning; 0 rows; `sync_state` stamped anyway.

Neither surfaces as a failure to a caller checking the exit code or `errored`.

## Blast radius

Scanned all `library/*/*/spec.yaml` on `main`:

| | count |
|---|---|
| CLIs with a spec | 311 |
| CLIs with any `html`/`binary`/`text` resource | 36 |
| **CLIs whose `defaultSyncResources()` includes one** | **10** |
| ...of which bare `sync` has **only** non-JSON resources | **9** |

For those 9, bare `sync` cannot do anything useful at all — every resource it defaults to is unmappable:

| CLI | category | non-JSON resources in `defaultSyncResources()` |
|---|---|---|
| amazon-orders | commerce | orders, transactions |
| averusa | devices | docs, products |
| extron | devices | literature |
| google-play | marketing | categories |
| motohunt | other | listings |
| psx | payments | circuit-breakers, debt, eligible-scrips, indices, screener, sectors |
| qsys | devices | page, product |
| rappi | food-and-dining | promotions |
| shodhganga | productivity | browse |
| fpi-india | other | auc, cdsl_reports, net_investment, registry, sector (mixed — `limits` is JSON) |

Every one of these CLIs solves the corpus problem with a *hand-authored* command instead — extron has `catalog sync`, averusa and qsys have `harvest`. The generated `sync` sits alongside as a trap that looks like the obvious thing to run.

## Why nothing catches it

`verify` passes: a resource synced. The exit code is 0 and `errored` is 0, so CI and any scripted caller see success. The only visible tell is `total_records` being implausibly small, which nothing asserts on. This is how `extron` shipped and stayed shipped.

## Acceptance

- **Positive:** a resource whose `response_format` is not JSON is either excluded from `defaultSyncResources()` at generation time, or its sync attempt is recorded as a failure/warning rather than a success.
- **Negative:** a NULL-`data` row is never written and never counted toward `total_records`; `sync_state` is not stamped for a resource that stored nothing, so `hintIfUnsynced` keeps warning and `doctor` does not report a fresh cache over an empty store.
- **Signal parity:** `sync_anomaly` with `reason: non_json_200_body` counts toward `warned` (averusa's shape), and the extron shape emits it at all.
- **Regression:** CLIs whose corpus comes from a hand-authored command (`catalog sync`, `harvest`) still generate and behave as they do today, minus the misleading success.

## Related

- #4336 — the other generator-floor defect filed today (stale `safeXNetVersion`).
- `mvanhorn/printing-press-library#1804` documents the real corpus builder for extron/qsys/crestron because of this. One knock-on worth noting: that PR's fix records the store's true row count via `db.Count`, which counts the NULL `literature` row this bug leaves behind — so the reported total reads 1 high for anyone who ran both commands. Removing the NULL row removes that wart too.
- #3816 (closed) is a different defect — a dependent walker's parent key never being sent. This one is a top-level resource with no dependents.

Drummerms:
## Intent

Closes #4342.

A CLI whose syncable resources are `response_format: binary` or `text` (sitemap indexes, plain-text exports) has no JSON enumeration for spec-driven sync, yet the generator emitted a full `sync` command for it. That sync fetches the non-JSON body, stores nothing useful (a NULL resource row on the extron shape, a `non_json_200_body` anomaly on the averusa shape), stamps `sync_state`, and exits 0 — after which `doctor` reports "Cache: fresh" over an empty store and staleness hints steer users at the generated no-op instead of the CLI's real corpus-builder command. Measured blast radius on the public library's `main`: 10 CLIs carry a non-JSON resource in `defaultSyncResources()`, and 9 of them have *only* non-JSON resources there — every one builds its corpus with a hand-authored command (`harvest`, `catalog sync`).

## Approach

The generator already has the right mechanism: `shouldEmitHTMLSyncStub` swaps in the sync stub when ≥70% of syncable resources are HTML page-mode. The gap is the predicate — binary/text resources counted as JSON-syncable, so they *diluted* the ratio below threshold (e.g. one HTML page-mode + two binary sitemaps = 1/3 = 33% → full sync emitted, exactly the qsys shape).

- Plumb the chosen list endpoint's `EffectiveResponseFormat()` through the profiler (`syncableMeta` → `SyncableResource`/`DependentResource.ResponseFormat`).
- Broaden the stub-count predicate from "HTML page-mode" to "lacks a JSON enumeration": HTML page-mode ∪ binary ∪ text, in both the profiled-resource counter and the spec-level endpoint fallback.
- Generalize the stub's error message and point it at the CLI's dedicated corpus-builder command.

Deliberately unchanged: HTML links/table/list-mode resources still count as syncable (link extraction is a real enumeration), `csv`/`xml` still count as syncable (the client converts them to JSON), and the existing `syncableResourceUsesHTMLPageMode` consumers outside the stub decision are untouched.

## Scope

Primary area: generator sync-stub decision + profiler response-format plumbing.

Why this belongs in this repo: every printed CLI inherits the sync template selection, so the fix belongs in the generator rather than hand-stubbing `sync` in each affected library CLI.

## Risk

A spec whose syncable resources are predominantly binary/text now gets the sync stub instead of a generated sync that appeared to work. That is the intent — such CLIs (all 9 in the library) already build their corpus with a hand-authored command — but any spec author relying on generated sync to fetch-and-discard a binary resource will now see an explicit error instead of silent success. Mixed specs below the existing 70% threshold keep generic sync (covered by a new regression test).

## Output Contract

Binary/text-majority specs emit the sync stub (clear error, no `sync_state` stamp, no "Cache: fresh" over an empty store). `TestGenerateBinarySyncablesEmitSyncStub` pins the stub for the pure-non-JSON shape (fails before this change: 0/3 counted → full sync), and `TestGenerateMinorityBinarySyncableKeepsGenericSync` pins that one binary resource among JSON list resources does not flip the decision.

## Verification

- `GOTOOLCHAIN=go1.26.6 go test ./internal/generator -run 'TestGenerateBinarySyncablesEmitSyncStub|TestGenerateMinorityBinarySyncableKeepsGenericSync' -count=1` — new tests; the stub test fails without the fix (verified by stashing the non-test changes)
- `GOTOOLCHAIN=go1.26.6 go test ./internal/profiler -count=1` — pass
- `GOTOOLCHAIN=go1.26.6 scripts/golden.sh verify` — all 34 cases pass (no golden fixture uses `response_format`)
- `GOTOOLCHAIN=go1.26.6 go test ./internal/generator -count=1` — one unrelated flake in a broad parallel run (`TestGeneratorSkipsReservedNovelFeatureRootCommands`, stderr-capture cross-talk between parallel tests; passes in isolation both with and without this change, and on a re-run)
- `GOTOOLCHAIN=go1.26.6 go build ./...`

### DIFF
--- a/internal/generator/generator.go
+++ b/internal/generator/generator.go
@@ -4690,19 +4690,23 @@ func (g *Generator) resetHTMLSyncStubCache() {
 	g.htmlSyncStub = false
 }
 
+// countHTMLPageModeSyncResources counts profiled sync resources overall and
+// how many of them lack a JSON enumeration (HTML page-mode, binary, or text
+// responses). The name keeps its historical HTML framing; the ratio feeds
+// shouldEmitHTMLSyncStub.
 func (g *Generator) countHTMLPageModeSyncResources() (total int, html int) {
 	if g == nil || g.profile == nil {
 		return 0, 0
 	}
 	for _, resource := range g.profile.SyncableResources {
 		total++
-		if syncableResourceUsesHTMLPageMode(resource) {
+		if syncableResourceLacksJSONEnumeration(resource) {
 			html++
 		}
 	}
 	for _, resource := range g.profile.DependentSyncResources {
 		total++
-		if dependentResourceUsesHTMLPageMode(resource) {
+		if dependentResourceLacksJSONEnumeration(resource) {
 			html++
 		}
 	}
@@ -4717,6 +4721,29 @@ func dependentResourceUsesHTMLPageMode(resource profiler.DependentResource) bool
 	return resource.UsesHTMLResponse && resource.HTMLExtract.EffectiveMode() == spec.HTMLExtractModePage
 }
 
+// syncableResourceLacksJSONEnumeration reports whether spec-driven sync cannot
+// produce a trustworthy JSON enumeration from this resource: HTML page-mode
+// extraction stores prose pages rather than records, and binary/text responses
+// (sitemaps, downloads, plain text) have no records at all. A CLI whose
+// syncable resources are predominantly of these shapes builds its corpus with
+// a hand-authored command, and its generated sync would stamp sync_state and
+// report success over an empty store — so such resources count toward the
+// sync-stub decision.
+func syncableResourceLacksJSONEnumeration(resource profiler.SyncableResource) bool {
+	return syncableResourceUsesHTMLPageMode(resource) || responseFormatLacksJSONEnumeration(resource.ResponseFormat)
+}
+
+func dependentResourceLacksJSONEnumeration(resource profiler.DependentResource) bool {
+	return dependentResourceUsesHTMLPageMode(resource) || responseFormatLacksJSONEnumeration(resource.ResponseFormat)
+}
+
+func responseFormatLacksJSONEnumeration(format string) bool {
+	return format == spec.ResponseFormatBinary || format == spec.ResponseFormatText
+}
+
+// countHTMLPageModeEndpoints is the spec-level fallback for
+// countHTMLPageModeSyncResources when profiling found no sync resources; it
+// applies the same lacks-JSON-enumeration test per endpoint.
 func countHTMLPageModeEndpoints(api *spec.APISpec) (total int, html int) {
 	if api == nil {
 		return 0, 0
@@ -4732,7 +4759,8 @@ func countHTMLPageModeEndpoints(api *spec.APISpec) (total int, html int) {
 func countResourceHTMLPageModeEndpoints(resource spec.Resource) (total int, html int) {
 	for _, endpoint := range resource.Endpoints {
 		total++
-		if endpoint.UsesHTMLResponse() && endpoint.HTMLExtract.EffectiveMode() == spec.HTMLExtractModePage {
+		if (endpoint.UsesHTMLResponse() && endpoint.HTMLExtract.EffectiveMode() == spec.HTMLExtractModePage) ||
+			responseFormatLacksJSONEnumeration(endpoint.EffectiveResponseFormat()) {
 			html++
 		}
 	}
--- a/internal/generator/html_sync_stub_test.go
+++ b/internal/generator/html_sync_stub_test.go
@@ -136,7 +136,7 @@ func TestGenerateHTMLMajorityWithJSONSyncKeepsGenericSync(t *testing.T) {
 	skillSrc := readGeneratedFile(t, outputDir, "SKILL.md")
 
 	assert.Contains(t, syncSrc, "func syncResource(")
-	assert.NotContains(t, syncSrc, "generic spec-driven sync template does not fit predominantly HTML page-mode endpoints")
+	assert.NotContains(t, syncSrc, "generic spec-driven sync template does not fit predominantly non-JSON endpoints")
 	assert.Contains(t, workflowSrc, "newWorkflowArchiveCmd")
 	assert.Contains(t, workflowSrc, "workflow archive")
 	assert.Contains(t, readmeSrc, "## Freshness")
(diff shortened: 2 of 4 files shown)


Drummerms:
CI note: the one failing job, `full-test (generator-2)`, fails on `TestGeneratorSkipsReservedNovelFeatureRootCommands` — a pre-existing flake unrelated to this change. Evidence: the same test fails in the same shard on `main`'s own latest CI run (https://github.com/mvanhorn/cli-printing-press/actions/runs/32770722877, 2026-08-24), and it passes in isolation both with and without this branch's changes (`go test ./internal/generator -run TestGeneratorSkipsReservedNovelFeatureRootCommands -count=1` → ok). The failure mode is stderr-capture cross-talk: the test swaps os.Stderr to capture the "would shadow" warning while parallel tests write their own warnings into the swapped stream — the captured buffer in the failure contains only unrelated `spec.Printer is empty` warnings while the expected warning appears in the raw job log. Filed separately as an issue. Happy to rebase/retrigger once a maintainer reruns the job.

tmchow:
The bug in #4342 is real. Binary/text resources get a generated `sync` that stores a NULL row, stamps `sync_state`, exits 0, and `doctor` says the cache is fresh.

Please don't fix that by retuning `shouldEmitHTMLSyncStub`'s 70% ratio. Mixed specs under that threshold still get the fake-success sync, so the lie remains.

What we want:

1. Binary and text resources must not land in `defaultSyncResources()` as JSON syncables.
2. If a sync request still gets a non-JSON body, do not upsert a NULL row, do not stamp `sync_state`, do not exit 0.

The all-binary CLIs can keep using their hand-authored `harvest` / `catalog sync`. A stub is fine as a consequence of (1), not as the product.

tmchow:
Closing this so #4342 can go back into the queue. Thanks for the PR — the Aug 28 rewrite note is still the shape we want if you pick it up again.

PR #4365, opened 2026-08-25
fix(verifier): probe sync capabilities before bounding pages

madmoneymike5:
## Summary

Make the live data-pipeline verifier probe a CLI's `sync --help` output before appending `--max-pages 1`.

## Problem

The verifier was adding `--max-pages 1` to every live sync attempt. That is safe for paginated CLIs, but it makes a complete, non-paginated sync fail with an unknown-flag error. The verifier was changing the command it was supposed to inspect.

## Fix

- Run `sync --help` once for the live probe.
- Add the page bound only when the help output advertises `--max-pages`.
- Leave mock mode and unpaginated live sync unchanged.
- Add a regression test using a harmless live-sync probe that rejects `--max-pages`.

## Verification

- Focused sync-capability tests — passed
- `GOFLAGS=-buildvcs=false go test ./internal/pipeline` — passed
- `GOFLAGS=-buildvcs=false go vet ./...` — passed
- `GOFLAGS=-buildvcs=false go build ./...` — passed
- `git diff --check` — passed

The unmodified upstream full suite also contains an existing generator expectation failure (`TestGeneratorSkipsReservedNovelFeatureRootCommands`); the verifier package and this patch's focused tests are green. The `GOFLAGS` setting only disables VCS stamping for generated temporary test binaries.

### DIFF
--- a/internal/pipeline/runtime.go
+++ b/internal/pipeline/runtime.go
@@ -655,11 +655,13 @@ func runDataPipelineTest(binary, cliDir, mode string, envFn func() []string, exp
 	dbPath := filepath.Join(tmpDir, "test.db")
 	env = append(env, "HOME="+tmpDir) // so sync uses temp location
 
-	// Test sync (if it exists)
-	// Discover whether live sync supports pagination before adding the
-	// bounding flag; some CLIs expose a complete, non-paginated sync.
+	// Test sync (if it exists). Discover whether live sync supports pagination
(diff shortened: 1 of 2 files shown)


tmchow:
Closing this. Generated syncs already declare `--max-pages`. Forcing `--max-pages 1` on every verifier probe is the wrong default: it hides a missing flag instead of catching it.

Issue #3547, opened 2026-07-10
[retro] .mcpb manifest: non-spec top-level keys + generic binary path + under-declared multi-cred user_config (Claude Desktop install fails)


DamienStevens:
## Context
Surfaced by Servosity/msp-skills#185 (connectwise-manage v0.1.3 `.mcpb`). The
generated `.mcpb` manifest fails to install via Claude Desktop → Settings →
Extensions. Fixed at the artifact level in Servosity/msp-skills#188, but the root
cause is the generator — every binary connector's `.mcpb` inherits the same
defects. Three generator findings:

## 1. Manifest emits non-spec top-level keys → validator hard-rejects
The generator writes `printing_press_version` and `press_provenance` at the
**top level** of `manifest.json`. These are not MCPB manifest spec keys, so
Claude Desktop rejects the entire manifest:

> Failed to preview extension: Invalid manifest: Unrecognized key(s) in object:
> 'printing_press_version', 'press_provenance'

**Fix:** emit provenance under the spec's reverse-DNS metadata slot,
`_meta["io.github.mvanhorn.cli-printing-press"]`, instead of top-level. (Verified
against anthropics/mcpb MANIFEST.md — `_meta` is a recognized key; top-level extras
are not.) Any downstream reader (e.g. catalog builders) should read from `_meta`.

## 2. `command`/`entry_point` point at a binary that is never shipped
The manifest sets `command: ${__dirname}/bin/<slug>-mcp`, but releases ship only
**arch-suffixed** binaries (`-darwin-arm64`, `-darwin-amd64`, `-windows-amd64.exe`,
`-linux-amd64`, `-linux-arm64`). No file matches the generic name on any platform →
ENOENT everywhere, not just Windows.

**Fix:** emit `platform_overrides` selecting the correct binary per OS. Note MCPB
`platform_overrides` keys on OS only (`darwin`/`linux`/`win32`) with **no CPU-arch
axis**, so per-OS overrides still can't distinguish Intel vs Apple-Silicon macOS or
amd64 vs arm64 Linux. Full coverage needs either a tiny arch-detecting launcher
shim as the `command`, or universal/fat binaries. Recommend the shim.

## 3. Multi-credential auth under-declares `user_config`
For connectwise-manage the CLI requires 5 env credentials (CW_COMPANY_ID,
CW_PUBLIC_KEY, CW_PRIVATE_KEY, CW_CLIENT_ID, CW_SITE) but the generator wired only
[...]

PR #4366, opened 2026-08-26
feat(pipeline): multi-platform .mcpb bundles that actually launch (#3547 finding 2)

DamienStevens:
## Intent

Issue: Refs #3547 (finding 2)

#3547 landed finding 1 (provenance moved out of the non-spec top-level keys). Finding 2 did not, and it is the one that makes an extension fail for a real user:

> The manifest sets `command: ${__dirname}/bin/<slug>-mcp`, but releases ship only **arch-suffixed** binaries. No file matches the generic name on any platform.

The press's own `BuildMCPBBundle` is single-platform and internally consistent - it writes the binary *to* the path the manifest declares, so `command` always resolves. The defect appears in a consumer that packs every released binary into one bundle, which is what you want, because a `.mcpb` is downloaded once and installed on whatever machine the operator has. The press had no way to express that: `MCPBLaunchSpec` had no `platform_overrides`, so there was no correct way to build a multi-platform bundle and consumers built an incorrect one.

Downstream receipt: Servosity/msp-skills currently publishes 64 such bundles. Verified by downloading two:

```
$ unzip -l hudu-mcp.mcpb | grep bin/
  bin/hudu-mcp-darwin-amd64   bin/hudu-mcp-darwin-arm64
  bin/hudu-mcp-linux-amd64    bin/hudu-mcp-linux-arm64
  bin/hudu-mcp-windows-amd64.exe
$ jq -r .server.mcp_config.command manifest.json
${__dirname}/bin/hudu-mcp          # in no bundle, on any platform
```

Claude Desktop validates the manifest but does not test-run the binary, so the extension installs, prompts for credentials, and only fails at the first tool call.

## Approach

Three pieces.

**`platform_overrides` in the manifest type.** `MCPBLaunchSpec` gains an omitempty `PlatformOverrides map[string]MCPBLaunchOverride`, so a multi-platform bundle is expressible at all.

**A universal binary for darwin.** MCPB keys `platform_overrides` on OS only - `darwin`, `linux`, `win32` - and defines no `${arch}` template variable, so per-OS overrides still cannot separate Intel from Apple Silicon. #3547 called for either a launcher shim or a universal binary and recommended the shim; this uses the universal binary instead, because a shim depends on the archive's executable bit surviving unpacking and on a shell being present, and neither is guaranteed. `writeMachoUniversal` builds the fat Mach-O directly through `debug/macho` - no `lipo`, no Xcode dependency, so a Linux runner can produce a macOS-universal bundle. Linux and Windows have no fat-binary format, so those overrides name one architecture (amd64 first: Windows on ARM runs x64 under emulation, the reverse is not true).

**A validator that makes the defect unshippable.** `ValidateMCPBBundle` opens the finished archive and asserts every command path the manifest declares - base `command`, each `platform_overrides.<os>.command`, and `entry_point` - names a member the archive actually contains. It runs at the end of **both** bundle paths, so the single-platform builder is covered too. That is what turns this from a first-tool-call mystery into a build error.

## Scope

Primary area: `internal/pipeline` (MCPB manifest + bundle)

Why this belongs in this repo: the manifest is generated here, and the missing `platform_overrides` capability is why no consumer could build a correct multi-platform bundle. Fixing it in one consumer would leave every other one to rediscover the same thing.

## Risk

Additive. `platform_overrides` is `omitempty`, so single-platform manifests are byte-identical to before. `BuildMCPBMultiPlatformBundle` is a new entry point; existing callers (`promote_artifacts.go`, `internal/cli/bundle.go`) are untouched.

The one behaviour change to an existing path: `BuildMCPBBundle` now returns `ValidateMCPBBundle`'s error instead of `nil`. For the single-platform path that is a tautology today - it writes the binary to the declared `entry_point` - so it should never fire; it is there so a future change to either half cannot silently reintroduce the defect. The full `internal/pipeline` suite passes unchanged, which is the evidence that it does not fire on current behaviour.

A universal-binary merge failure returns an error rather than degrading to a single-arch bundle, deliberately: a silent degrade breaks one darwin population and looks fine.

## Output Contract

Changes generated `manifest.json` only when a caller opts into the multi-platform builder. Covered by emitted-output assertions rather than by prose:

- `TestValidateMCPBBundleCatchesTheShippedDefect` constructs the exact archive shape shipped today (generic `command`, arch-suffixed members) and asserts it does not validate.
- `TestValidateMCPBBundleAcceptsAResolvableBundle` and `...CatchesABrokenOverride` prove the check discriminates, and that it is not satisfied by the base command alone.
- `TestBuildMCPBMultiPlatformBundle` cross-compiles real darwin/amd64 and darwin/arm64 binaries, builds the bundle, then reopens the merged file with `macho.OpenFat` and asserts both `CpuAmd64` and `CpuArm64` slices are present - so the universal binary is verified as a real fat Mach-O, not just as bytes on disk.
- `TestBuildMCPBMultiPlatformBundleSingleDarwinArch` covers the no-merge case.

## Verification

```
go build ./...                                    ok
go vet ./...                                      ok
go test ./internal/pipeline/                      ok  (81s, full suite, unchanged)
go test ./internal/pipeline/ -run 'MCPBMultiPlatform|ValidateMCPBBundle' -v
                                                  ok  (all cases above)
```

## AI / Automation Disclosure

### DIFF
--- a/internal/pipeline/mcpb_bundle.go
+++ b/internal/pipeline/mcpb_bundle.go
@@ -117,7 +117,13 @@ func BuildMCPBBundle(params BundleParams) error {
 	if err := zw.Close(); err != nil {
 		return fmt.Errorf("finalizing bundle archive: %w", err)
 	}
-	return nil
+	// Assert the finished archive actually contains every path its manifest
+	// declares. The single-platform path writes the binary to the declared
+	// entry_point, so this is normally a tautology - but it is the cheap guard
+	// that stops a future change to either half from shipping a bundle that
+	// installs, prompts for credentials, and fails at the first tool call
+	// (mvanhorn/cli-printing-press#3547 finding 2).
+	return ValidateMCPBBundle(params.OutputPath)
 }
 
 func rewriteMCPBManifestVersion(data []byte, version string) ([]byte, error) {
--- a/internal/pipeline/mcpb_manifest.go
+++ b/internal/pipeline/mcpb_manifest.go
@@ -105,6 +105,20 @@ type MCPBLaunchSpec struct {
 	Command string            `json:"command"`
 	Args    []string          `json:"args"`
 	Env     map[string]string `json:"env,omitempty"`
+	// PlatformOverrides selects a different command per OS. The MCPB spec keys
+	// this on OS only - "darwin", "linux", "win32" - with no CPU-architecture
+	// axis and no ${arch} template variable, so a bundle carrying one binary per
+	// (os, arch) cannot be addressed by Command alone. See BuildMCPBBundle for
+	// how a multi-platform bundle keeps every declared path resolvable.
+	PlatformOverrides map[string]MCPBLaunchOverride `json:"platform_overrides,omitempty"`
+}
+
+// MCPBLaunchOverride is the per-OS half of platform_overrides. Every field is
+// optional; the host merges what is present over the base MCPConfig.
+type MCPBLaunchOverride struct {
+	Command string            `json:"command,omitempty"`
+	Args    []string          `json:"args,omitempty"`
+	Env     map[string]string `json:"env,omitempty"`
 }
 
 // MCPBVar is one entry in user_config — a value the host collects from the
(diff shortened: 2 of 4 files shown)


tmchow:
Closing this. Main already writes a target-specific `.mcpb`. A multi-platform zip is a new installer product, not a press operator bug.

PR #4440, opened 2026-08-30
fix(cli): set multipart file-part Content-Type from extension

### LINKED ISSUE #3667: retro(v4.29.0): binary file param routed to Fields not FileFields; file-part Content-Type; hidden resource groups prune novel children from MCP mirror
Retro from a real print (private FastAPI accounting SaaS, curated 15-op OpenAPI spec, `cli-printing-press 4.29.0`, Go 1.26.5, Linux). Three generator-class findings, each verified against the live API and fixed in our fork; sharing so the next print gets them for free.

## 1. Multipart binary file param emitted into `Fields` instead of `FileFields`

The spec declares a standard FastAPI `multipart/form-data` upload (`file` as `type: string, format: binary`). The generated command correctly calls `c.PostMultipartWithParams(...)` (so this is distinct from closed #886/#1545), but routes the flag value into the **string fields map**:

```go
if bodyFile != "" {
    fields["file"] = bodyFile   // path sent as a string form field
}
```

`fileFields` stays empty, the API receives the literal path and rejects with `422 "Value error, Expected UploadFile, received: <class 'str'>"` — including when running the exact example printed by the command's own `--help`. Both fresh-agent evaluations in our A/B pilot hit this and had to fall back to `curl -F`.

**Fix (one line, verified live):** route `format: binary` params into `fileFields["file"] = bodyFile`.

## 2. File parts hardcode `Content-Type: application/octet-stream`

After fix 1, `encodeMultipartBody` uses `writer.CreateFormFile`, which hardcodes `application/octet-stream` on the part. APIs that validate the part's media type reject it: `415 "Content-Type 'application/octet-stream' is not supported. Supported types: application/pdf, image/jpeg, image/png."` Possibly related to open #3451, but this is the multipart *part* header rather than raw-body content-type honoring.

**Fix (verified live):** build the part with `writer.CreatePart` and set `Content-Type` from `mime.TypeByExtension(filepath.Ext(path))`, falling back to octet-stream.

## 3. Hidden promoted resource groups prune their visible novel children from the MCP mirror

The generator emits promoted resource groups (e.g. `batch`, `invoice`) with `Hidden: true` to keep root `--help` lean, and wires approved novel commands as their children (`batch upload`, `batch wait`, `invoice wait` in our run). But `cobratree.walk` skips hidden commands entirely:

```go
if child.Hidden || isMCPHidden(child) {
    continue
}
```

so every novel command nested under a resource group silently disappears from the MCP tool mirror (top-level novels like `preflight` register fine). Result: MCP `tools/list` exposed 41 tools with 3 of our 4 approved novel features absent; agents on the MCP surface can never reach the differentiating commands the absorb gate approved.

**Fix (verified live, 51 tools after):** descend into `Hidden` groups without registering the hidden command itself, keeping `mcp:hidden` as the explicit whole-subtree opt-out:

```go
if isMCPHidden(child) {
    continue
}
childPath := append(append([]string{}, path...), child.Name())
if child.Hidden {
    walk(child, childPath, visit)
    continue
}
visit(child, childPath)
```

## Impact

Findings 1+2 make every `format: binary` multipart endpoint unusable out of the box (upload being the core use case of document-processing APIs). Finding 3 silently breaks the "novel commands become MCP tools automatically" contract for any spec with promoted resource groups. Happy to share more context from the run artifacts if useful.

tmchow:
## Intent

Generated multipart uploads still send every file part as `application/octet-stream` because `encodeMultipartBody` uses `CreateFormFile`. After the FileFields routing fix, APIs that whitelist part types (`application/pdf`, `image/jpeg`, `image/png`) reject the generated command with 415, including the example in the command's own `--help`.

Closes #3667

## Approach

Confirm on current main that the other two #3667 findings are already fixed, then change only the live encode path:

- `format:binary` body params already route to `fileFields` via `multipartBodyMaps` / `isBinaryParam`.
- cobratree walk already visits children of `Hidden` groups while skipping `mcp:hidden`.
- Multipart `--dry-run` preview already skips FileFields IO.

Replace `writer.CreateFormFile` with `writer.CreatePart` and set `Content-Type` from `mime.TypeByExtension(filepath.Ext(path))`, falling back to `application/octet-stream` when the extension is unknown. Do not reopen the dry-run preview path.

## Scope

Primary area: `internal/generator/templates/client.go.tmpl` (`encodeMultipartBody`) plus generated-output tests and the `generate-golden-api` client golden.

Why this belongs in this repo: this is a generator-owned client emission defect. Every printed CLI with a `multipart/form-data` binary file field inherits the hardcoded octet-stream part type until the template changes.

## Risk

- Live multipart uploads of known extensions now send a matching part Content-Type. APIs that previously accepted octet-stream only should still accept the more specific type.
- Unknown extensions keep the previous octet-stream fallback.
- Dry-run preview, FileFields routing, and Hidden-group MCP walk are unchanged.
- Generated `internal/client/client.go` changes for CLIs with multipart requests (golden fixture updated).

## Output Contract

- Emitted `encodeMultipartBody` now calls `CreatePart` and `mime.TypeByExtension`; it no longer calls `CreateFormFile`.
- Evidence: `TestGenerateMultipartRequestBodyUsesMultipartClient` asserts the emitted helpers; `TestMultipartFilePartContentTypeFromExtension` compiles a generated CLI and encodes `.pdf` / `.png` / `.jpeg` / unknown parts.
- `scripts/golden.sh update` refreshed `testdata/golden/expected/generate-golden-api/printing-press-golden/internal/client/client.go` only.

## Verification

Commands:

- Inspected #3667 against current `main` and confirmed FileFields routing + Hidden-group walk already land.
- `gofmt -w internal/generator/multipart_test.go`
- `scripts/golden.sh update` (kept only the intended `generate-golden-api` client.go fixture; discarded unrelated dogfood-verdict-matrix drift from a local fixture-build failure)
- `go test ./internal/generator -run 'TestGenerateMultipartRequestBodyUsesMultipartClient|TestMultipartDryRunDoesNotOpenFileFields|TestMultipartFilePartContentTypeFromExtension|TestGenerateOpenAPIMultipartRequestBodyUsesMultipartClient'` — pass
- `go test -timeout 15m ./...` — all packages except `internal/generator` pass; that package exceeds a 15m unsharded budget on this runner (CI shards it)
- `go test -timeout 45m ./internal/generator` — pass (870s)
- `bash scripts/verify-generator-output.sh generate-golden-api` — pass
- `python3 .github/scripts/pr-review-state/greptile_feedback.py 4440` — READY

## AI / Automation Disclosure

<div><a href="https://cursor.com/agents/bc-3777fe0e-2e56-4193-95c2-40bb5d7f0922?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-3777fe0e-2e56-4193-95c2-40bb5d7f0922&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/internal/generator/multipart_test.go
+++ b/internal/generator/multipart_test.go
@@ -354,9 +354,14 @@ func TestMultipartFilePartContentTypeFromExtension(t *testing.T) {
 	outputDir := filepath.Join(t.TempDir(), naming.CLI(apiSpec.Name))
 	require.NoError(t, New(apiSpec, outputDir).Generate())
 
-	clientSrc := readGeneratedFile(t, outputDir, "internal", "cli", "assets_upload.go")
-	assert.Contains(t, clientSrc, `fileFields["assetData"] = bodyAssetData`)
-	assert.NotContains(t, clientSrc, `fields["assetData"]`)
+	clientSrc := readGeneratedFile(t, outputDir, "internal", "client", "client.go")
+	assert.Contains(t, clientSrc, `mime.TypeByExtension(filepath.Ext(path))`)
+	assert.Contains(t, clientSrc, `writer.CreatePart(h)`)
+	assert.NotContains(t, clientSrc, `writer.CreateFormFile(`)
+
+	endpointSrc := readGeneratedFile(t, outputDir, "internal", "cli", "promoted_assets.go")
+	assert.Contains(t, endpointSrc, `fileFields["assetData"] = bodyAssetData`)
+	assert.NotContains(t, endpointSrc, `fields["assetData"]`)
 
 	const runtimeTest = `package client
 
--- a/internal/generator/multipart_test.go
+++ b/internal/generator/multipart_test.go
@@ -61,7 +61,9 @@ func TestGenerateMultipartRequestBodyUsesMultipartClient(t *testing.T) {
 
 	clientSrc := readGeneratedFile(t, outputDir, "internal", "client", "client.go")
 	assert.Contains(t, clientSrc, `func (c *Client) PostMultipart(ctx context.Context, path string, fields map[string]string, fileFields map[string]string) (json.RawMessage, int, error)`)
-	assert.Contains(t, clientSrc, `writer.CreateFormFile(fieldName, filepath.Base(filePath))`)
+	assert.Contains(t, clientSrc, `mime.TypeByExtension(filepath.Ext(path))`)
+	assert.Contains(t, clientSrc, `writer.CreatePart(h)`)
+	assert.NotContains(t, clientSrc, `writer.CreateFormFile(`)
 	assert.Contains(t, clientSrc, `req.Header.Set("Content-Type", contentType)`)
 
 	endpointSrc := readGeneratedFile(t, outputDir, "internal", "cli", "assets_upload.go")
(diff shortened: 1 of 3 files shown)


tmchow:
@mergifyio queue

PR #4506, opened 2026-09-02
fix(cli): report lock stale only when the lock is actually stale

### LINKED ISSUE #4468: lock status reports stale: true for every held lock, at any age
**Component:** cli (lock)
**Priority:** P2
**Found on:** v4.31.3, Windows 11, local offset UTC+03:00

## Summary

`lock status --json` reports `"stale": true` for **every held lock**, regardless of
age or heartbeat. A lock acquired 60 milliseconds ago reports as stale. `stale`
only reads `false` when no lock is held at all, where it is vacuous.

This makes the flag unusable for its purpose: an operator or skill cannot use it
to distinguish an abandoned lock from a live one, and the documented stale-lock
recovery path has no working signal to key on.

## Reproduction

```
$ cli-printing-press lock acquire --cli pp-hb --scope my-workspace
$ cli-printing-press lock status --cli pp-hb --json
  "stale": true,
  "age_seconds": 0.0614246,

$ cli-printing-press lock update --cli pp-hb --phase build
$ cli-printing-press lock status --cli pp-hb --json
  "stale": true,
  "phase": "build",
  "age_seconds": 0.0656387,

$ cli-printing-press lock release --cli pp-hb
$ cli-printing-press lock status --cli pp-hb --json
  "held": false,
  "stale": false,
```

A fresh heartbeat via `lock update` does not clear it.

## Note on the timestamps

`age_seconds` is computed correctly. The lock's own fields are consistent and fresh:

```
"acquired_at": "2026-08-31T10:44:45.6346618+03:00",
"updated_at":  "2026-08-31T10:44:45.6346618+03:00"
```

So `stale` is not disagreeing with `updated_at` vs `acquired_at` — those are
identical here. Whatever computes `stale` is not using the same basis as
`age_seconds`.

**Unverified hypothesis, offered only as a place to look:** this machine runs at
UTC+03:00, and the serialized timestamps carry a `+03:00` offset. If the staleness
comparison parses those into a zero-offset/naive time and compares against a
local `now`, every lock would appear ~3 hours old and trip any threshold under
that. I could not test this — Go on Windows ignores the `TZ` environment variable,
so I had no way to vary the offset without changing the OS setting. If the
staleness threshold is under 3h, a UTC-machine reproduction may show `stale:
false`, which would confirm it and would also explain why this has gone
unreported.

## Suggested fix

Compute `stale` from the same normalized instant `age_seconds` already uses, and
add a regression test that asserts `stale == false` immediately after `acquire`
under at least one non-UTC local offset.

tmchow:
## Intent

`lock status --json` reported `"stale": true` for every held lock, including one acquired milliseconds earlier. A heartbeat via `lock update` did not clear it. Operators and skills could not tell an abandoned lock from a live one.

Closes #4468

## Approach

`IsStale` treated a dead owner PID as stale. `lock acquire` and `lock update` are discrete CLI processes that exit immediately, so every held lock looked abandoned even when `age_seconds` and `updated_at` were fresh.

Staleness now follows the documented 30-minute heartbeat threshold only. The same check is used for status reporting and auto-reclaim. PID liveness helpers are removed because they cannot describe this lock's process model.

## Scope

Primary area: `internal/pipeline` lock status / acquire reclaim.

Why this belongs in this repo: this is the Printing Press lock command, not a printed-CLI-only fix.

## Risk

A crashed build that used to look immediately stale now stays held until the heartbeat is older than 30 minutes, matching the documented policy. Fresh locks can no longer be auto-reclaimed just because the acquire process exited.

Generated output, MCP, auth, publish, verifier, and scorer are unchanged.

## Output Contract

N/A — lock status JSON field semantics change at runtime; no generator or golden fixtures are involved.

## Verification

- Reproduced on current `main` with the real binary: a ~10ms-old lock reported `stale: true`; after this change the same acquire/update path reports `stale: false`.
- `TestLockStatus_StaleFollowsHeartbeatAge` covers fresh acquire (dead PID, UTC+03:00 timestamps), heartbeat, and a lock past `StaleLockThreshold`.
- `go test ./...` passed.

- [x] Generator/template change: verified generated output, including emitted-code assertions or compiled generated CLI output — N/A, not a generator change
- [x] Generator/template change: covered the affected fallback or variant shape, not only happy-path fixtures — N/A
- [x] Generator/template change: checked emitted definitions and call sites for matching gates — N/A

## AI / Automation Disclosure

<div><a href="https://cursor.com/agents/bc-db1192e4-a965-4ee8-b7e4-133bbfa09423?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-db1192e4-a965-4ee8-b7e4-133bbfa09423&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


tmchow:
@mergifyio requeue

PR #4519, opened 2026-09-02
fix(cli): persist rotating refresh tokens across live dogfood

### LINKED ISSUE #4462: scorer: dogfood --live's env-var credential is single-use against rotating refresh tokens, so the acceptance marker is unreachable (and the operator's token gets revoked)
## Summary

`dogfood --live`'s only credential channel is an environment variable (`--auth-env`). Against an IdP
that **rotates refresh tokens on use**, an environment variable is single-use: the first sandboxed
subprocess refreshes and rotates the token, and every subsequent subprocess fails because it holds the
now-revoked original.

Observed on a printed CLI with `auth.type: oauth2_refresh` (Casdoor-backed API):

| Run | Result |
|---|---|
| No credentials | 136 checks `unverified` — `{"message":"No token provided"}` |
| `--auth-env <VAR>` with a valid refresh token | 0 auth-missing errors, matrix grew 186 → **322** executed tests, **284 checks failed** with `invalid_grant` |

Representative failure, repeated across ~70 commands:

```
Error: refreshing access token: HTTP 400:
{"error":"invalid_grant","error_description":"refresh token is invalid, expired or revoked"}
```

Acceptance marker: `{"status":"fail","matrix_size":322,"tests_passed":175,"tests_failed":147}`
with `failure_summary.http_4xx = 146`.

**To be clear: `--auth-env` itself works.** It delivered the credential — the "No token provided" 401s
disappeared and ~136 previously-unrunnable checks began executing. The defect is that a single static
env value cannot survive rotation across N independent subprocesses.

## Isolated proof

Three consecutive invocations of one command, same env-var refresh token, fresh `HOME` each time (so
nothing can persist a rotated token):

```
attempt 1: exit=4  invalid_grant
attempt 2: exit=4  invalid_grant
attempt 3: exit=4  invalid_grant
--- control: same binary, file-persisted credential ---
normal use exit=0
```

The control passes because the CLI writes the rotated token back to its credentials file. The runner's
sandbox has no such file, so every subprocess replays a dead token.

## Collateral damage worth noting

The matrix run **revoked the operator's stored refresh token**. That did not surface for ~47 minutes,
because the cached access token was still valid and no refresh was needed; `doctor` reported
`Credentials: valid` the whole time. A verification run silently invalidating the operator's
credential, with a delayed failure, is worse than failing loudly.

## Why this generalizes

Refresh-token rotation is Casdoor's default and is standard-or-optional in Auth0, Okta, and Keycloak.
Any printed CLI generated with `auth.type: oauth2_refresh` against one of those providers hits this,
regardless of API shape or spec source. The consequence is that `phase5-acceptance.json` can never
reach `status: pass` for that whole class of CLI, which is the gate the main SKILL's Phase 5.6
requires before promoting.

## Possible fixes (uncertain — pick after inspecting the runner)

I could not read the runner's source (binary-only install, `IN_REPO=false`), so I am not prescribing
one implementation:

1. **Give the sandbox a writable credential store** shared across subprocesses, so the first rotation
   persists and subsequent subprocesses use the current token. Closest to how the CLI behaves normally.
2. **Refresh once in the runner** and hand subprocesses the resulting short-lived *access* token rather
   than the refresh token, so no subprocess performs a rotation.
3. **Detect the cascade and stop.** If `invalid_grant` appears after the first success, abort with a
   clear diagnosis instead of reporting ~300 unrelated command failures — and warn that the operator's
   stored token may now be revoked.

(3) is worth doing regardless of whether (1) or (2) lands, because the current output blames the CLI's
commands for an auth-plumbing failure.

## Acceptance criteria

- **positive:** on an `oauth2_refresh` CLI against a rotating IdP, a full matrix run completes with no
  `invalid_grant` failures, and the operator's stored refresh token still works afterward.
- **negative:** a non-rotating API (static API key, bearer token) behaves exactly as today; the
  subprocess sandbox is not otherwise loosened.
- **positive:** where rotation cannot be supported, the runner stops on the `invalid_grant` cascade
  with a message naming the cause, rather than emitting ~300 per-command failures.

## Environment

- cli-printing-press v4.31.3 (Linux/WSL, Go 1.26.7); the same behavior was reached from v4.31.2 on Windows.
- Printed CLI: `auth.type: oauth2_refresh`, Casdoor IdP, second required header supplied via
  `required_headers`.

## Related issues

- #4286 — `verify --env-var` never reaches live mode (`related-area`: same "scorer cannot authenticate"
  family, different subcommand and different mechanism; that one is about mode selection, this one is
  about credential lifetime).
- #4346 — `lock promote` installs a CLI whose phase5 marker says `status: fail` (`related-area`:
  the downstream consequence of an unreachable `pass` marker).
- #4259 — live-dogfood coverage rules undocumented (`extends`).

---

*Filed from `/printing-press-retro`. Artifacts intentionally not attached: the finding is
provider-shaped rather than API-shaped, and the run's manuscripts describe a private vendor's
internal API surface.*

paride-cmd:
## Intent

Fixes #4462.

`dogfood --live --auth-env` inherited a static refresh token in every sandboxed subprocess. Against rotating IdPs (`oauth2_refresh` / Casdoor), the first child rotated and revoked that token; later commands got `invalid_grant` and the operator's stored refresh token could be revoked. The acceptance marker was unreachable.

## Approach

Prefer a writable credential store shared across live-dogfood subprocesses (issue option 1), with an `invalid_grant` cascade abort as a safety net (option 3).

- Mirror and sync-back `credentials.toml` for `oauth2_refresh` (the generated CLI persists tokens there, not in `config.toml`).
- Seed the sandbox file from `--auth-env` / refresh-token-shaped env when the file has no `refresh_token`.
- Strip rotating refresh-token env vars from subprocesses so the shared file wins after the first rotation.
- After a live happy/json pass, abort remaining commands if a later fail contains `invalid_grant`, with a diagnosis that the operator token may now be revoked.
- API-key `--auth-env` is unchanged (no strip, no seed).

## Scope

Primary area: live dogfood runner (`internal/pipeline`), plus `--auth-env` help and the Phase 5 skill note.

Why this belongs in this repo: the runner, not a printed CLI, was replaying a static env credential into isolated subprocesses. Every `oauth2_refresh` CLI hits this on `--live`.

## Risk

oauth2_refresh live dogfood now prefers file creds over a static refresh-token env var. Non-rotating API-key `--auth-env` still inherits. Sync-back writes the rotated token into the operator `credentials.toml` (compare-and-swap; refuses if the operator file changed during the run). If rotation still leaks, the matrix fails fast instead of recording hundreds of command failures.

## Output Contract

N/A for generator/templates/manifests/MCP/scorecard. `--auth-env` help text changed; there is no golden fixture for `dogfood --help`.

## Verification

- [x] `gofmt` on changed Go files
- [x] `go test ./internal/pipeline -count=1` including rotating-refresh, API-key unchanged, cascade abort, credentials.toml mirror, and env-strip tests
- [x] `GOTOOLCHAIN=go1.26.6 go test ./internal/spec ./internal/cli ./internal/pipeline ./internal/reachability ./cmd/cli-printing-press -count=1`

Local `go test ./...` on Go 1.27 fails to compile `github.com/enetx/http2` (`DisableClientPriority`); rerun with `GOTOOLCHAIN=go1.26.6`. Generator package needs `-timeout` longer than the default 10m; it was not changed in this PR. Golden not run: no generated-output contract change.

## AI / Automation Disclosure

Made with [Cursor](https://cursor.com)

### DIFF
--- a/internal/cli/dogfood.go
+++ b/internal/cli/dogfood.go
@@ -108,7 +108,7 @@ func newDogfoodCmd() *cobra.Command {
 	cmd.Flags().StringVar(&level, "level", "full", "Live dogfood depth: quick or full")
 	cmd.Flags().DurationVar(&timeout, "timeout", 30*time.Second, "Timeout for each live dogfood test")
 	cmd.Flags().StringVar(&writeAcceptance, "write-acceptance", "", "Write phase5-acceptance.json to this path on every outcome (status:pass on success, status:fail with a failure_summary block on failure)")
-	cmd.Flags().StringVar(&authEnv, "auth-env", "", "Environment variable that proves an API credential was available for the acceptance marker")
+	cmd.Flags().StringVar(&authEnv, "auth-env", "", "Environment variable holding an API credential for the acceptance marker. For oauth2_refresh, the value is copied into a shared sandbox credential file so refresh-token rotation persists across subprocesses")
 	cmd.Flags().StringVar(&authTier, "auth-tier", "", "Credential tier for live dogfood; falls back to PP_AUTH_TIER and skips commands annotated pp:requires-tier on mismatch")
 	cmd.Flags().BoolVar(&allowDestructive, "allow-destructive", false, "Re-enable testing of endpoints classified as destructive-at-auth. Default skips them to prevent runner-credential rotation.")
 	cmd.Flags().BoolVar(&overwriteCommandMirror, "overwrite-command-mirror", false, "Overwrite a differing command_mirror_capabilities block in internal/mcp/tools.go from research.json")
--- a/internal/pipeline/live_dogfood.go
+++ b/internal/pipeline/live_dogfood.go
@@ -62,6 +62,13 @@ const reasonUnverifiedNeedsAccess = "unverified-needs-access"
 // marker is accepted by `lock promote`.
 const reasonCookieAuthNoHarnessSession = "cookie-auth-no-harness-session"
 
+// reasonRefreshTokenRotationCascade is the Fail reason when live dogfood
+// sees invalid_grant after an earlier live pass. Rotating IdPs revoke the
+// previous refresh token on use; a static --auth-env value then poisons
+// every later subprocess. Prefer the shared credential file; this abort is
+// the safety net when rotation still leaked.
+const reasonRefreshTokenRotationCascade = "invalid_grant cascade: a refresh token was rotated by an earlier command and later commands reused the revoked value. oauth2_refresh live dogfood persists rotation in a shared sandbox credential file and strips the rotating env var; the operator's stored refresh token may now be revoked. Re-authenticate before retrying."
+
 // liveDogfoodVerdictCookieAuthNoSession is the report Verdict for a clean
 // cookie-auth skip outcome. Distinct from PASS/FAIL so the CLI exits 0 (the
 // 401s are not a defect) and writeLiveDogfoodAcceptance emits a skip marker
@@ -177,7 +184,7 @@ func RunLiveDogfood(opts LiveDogfoodOptions) (*LiveDogfoodReport, error) {
 			}},
 		}, nil
 	}
-	homeScope, err := scopeLiveDogfoodSubprocessHome(opts.CLIDir, opts.BinaryName)
+	homeScope, err := scopeLiveDogfoodSubprocessHome(opts.CLIDir, opts.BinaryName, opts.AuthEnv)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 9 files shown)


tmchow:
Real bug, and the shared `credentials.toml` approach is the right shape. Two holes before this can land:

1. Don't treat every `--auth-env` value as a refresh token. In `liveDogfoodRotatingRefreshEnvVars` / `liveDogfoodRotatingRefreshToken`, only seed and strip when the env var is refresh-token-shaped (`IsOAuthRefreshTokenEnvVar` / `OAuth2RefreshTokenEnvVar`). If `--auth-env` is the client ID (common on the oauth2_refresh publish path), it must not be written into `credentials.toml` as `refresh_token` and must not be stripped from children. Please add a regression test for that case.
2. Clear the go-lint modernize findings in `live_dogfood_refresh.go` (`slices.Contains`, `strings.SplitSeq`).

Optional: in the skill note, say "rotating identity providers" instead of naming Casdoor/Auth0/Okta.

tmchow:
Thanks for the start. Taking this from here so we can land the credentials.toml path with the refresh-shaped --auth-env guard and the go-lint cleanups. Will open a maintainer PR shortly.

PR #4520, opened 2026-09-03
fix(cli): prefer list response_path for sync resource unwrap

### LINKED ISSUE #4454: generator: responsePathForResource is keyed per resource, so a resource whose sampled endpoint is not its list endpoint loses the list path
## What happened

With `response_path: data.list` authored on `invoices.list`, the emitted
`responsePathForResource` returned `data` for the whole `invoices` resource:

```go
case "assignments": return []string{"data.assignments"}   // list is the sampled endpoint
case "comms":       return []string{"data.list"}          // ditto
case "invoices":    return []string{"data"}               // create was sampled instead
```

Resources whose sampled endpoint happened to be the list one received the correct path;
resources with more endpoints did not.

## Root cause

The unwrap path is keyed on **resource** identity while `response_path` is an
**endpoint-level** field, so one endpoint's value is taken for the whole resource.

## Why this generalizes

Any spec where a resource's list endpoint wraps its payload differently from its
create/update/delete endpoints — the normal REST shape. Setting the same `response_path` on
every endpoint to compensate breaks the non-list endpoints' output.

## Acceptance criteria

- positive: a resource whose `list` wraps at `data.list` and whose `create` wraps at `data`
  unwraps both correctly
- negative: a single-endpoint resource, or one with a uniform path, is unchanged

## Scope boundary

Keep the resource-level value as the fallback.

---
Component: `internal/generator/` · reproduced on v4.31.2

paride-cmd:
## Intent

Issue: Fixes #4454

`responsePathForResource` is keyed per resource. When create and list declare different `response_path` values, alphabetical tie-breaking could pick create's envelope (`data`) for sync instead of list's (`data.list`).

## Approach

Keep the resource-level switch (sync still unwraps by resource identity, not live URL), but rank endpoints so Syncable and list-shaped names beat create/update/delete. Endpoint-mirror commands continue baking each endpoint's own `ResponsePath`.

## Scope

Primary area: `internal/generator` (`responsePathCases` + sync template comment)

Why this belongs in this repo: every printed CLI inherits `responsePathForResource` from the sync template.

## Risk

Specs where a non-list Syncable endpoint intentionally owned the resource fallback could change which envelope sync uses; list-named Syncable endpoints still win first. Endpoint CLI/MCP unwrap paths are unchanged.

## Output Contract

- Emitted `responsePathForResource` still switches on resource name; list/syncable envelope preferred when endpoints disagree.
- Golden sync.go comment updated for generate-golden-api / query-endpoint / sync-walker / tier-routing.
- Evidence: `TestResponsePathCasesPrefersListOverCreate`, `TestGeneratedSyncPrefersListResponsePathWhenCreateDiffers` (compile + runtime unwrap), existing syncable/dedup cases.

## Verification

Commands run (scoped; no `go test ./...`, no pressauth):
- `GOTOOLCHAIN=go1.26.6 go test ./internal/generator/ -run 'TestResponsePathCases|TestGenerateSyncResponsePathDedup|TestGeneratedSync' -count=1`
- `GOTOOLCHAIN=go1.26.6 bash scripts/verify-generator-output.sh generate-query-endpoint-api generate-golden-api`
- `GOTOOLCHAIN=go1.26.6 bash scripts/golden.sh verify` (decoy HOME to avoid local library attribution drift)

## AI / Automation Disclosure

Made with [Cursor](https://cursor.com)

### DIFF
--- a/internal/generator/generator.go
+++ b/internal/generator/generator.go
@@ -6575,10 +6575,12 @@ func globalScopeEnvName(apiName string, p spec.Param) string {
 }
 
 // responsePathCase is one deduplicated entry for the sync template's
-// responsePathForResource switch: the switch key is the resource name (the
-// syncable endpoint wins when several share a resource) and the response
-// envelope path to return for it. The live request path is not part of the
-// key, so absolute or rewritten URLs still unwrap.
+// responsePathForResource switch: the switch key is the resource name and the
+// response envelope path to return for it. When a resource declares different
+// response_path values per endpoint, the list/syncable endpoint wins so sync
+// unwraps the collection envelope; endpoint-mirror commands still bake each
+// endpoint's own ResponsePath. The live request path is not part of the key,
+// so absolute or rewritten URLs still unwrap.
 type responsePathCase struct {
 	Key          string
 	ResponsePath string
@@ -6730,8 +6732,10 @@ func syncParamDefaultCases(syncable []profiler.SyncableResource, dependents []pr
 }
 
 // responsePathCases returns deterministic switch cases deduplicated by resource
-// identity. Syncable endpoints take precedence because the generated lookup is
-// used by sync; endpoint name breaks ties to keep output stable. The request
+// identity. Sync uses this lookup, so when endpoints disagree on response_path
+// the list/syncable envelope wins and mutating endpoints (create/update/…) do
+// not steal it. Endpoint-mirror commands keep their own ResponsePath. Rank:
+// Syncable, then list-shaped names/responses, then endpoint name. The request
 // path is not part of the key so unwrap still fires when the live URL differs
 // from the spec path.
 func responsePathCases(resources map[string]spec.Resource) []responsePathCase {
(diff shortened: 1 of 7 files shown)


tmchow:
Real bug — list's `response_path` must win over create when the resource-level sync fallback has to pick one.

Please tighten the ranking so it can't disagree with the endpoint sync actually calls. A non-Syncable `list*` name (e.g. `listArchive` with an object envelope) must not outrank the collection endpoint sync uses (e.g. Syncable `search` / `list`) when those declare different `response_path` values. Prefer the sync/collection endpoint's path for the resource fallback; keep demoting create/update/delete so the invoices list-vs-create case still returns `data.list`.

Add a regression test for that `listArchive` vs collection shape. Endpoint-mirror commands should keep each endpoint's own `ResponsePath`.

tmchow:
Thanks for the start. Taking this from here so we can land the sync/collection response_path ranking (and the listArchive regression). Will open a maintainer PR shortly.

PR #4541, opened 2026-09-03
fix(skills): reference in-tree secret-protection copy to avoid cross-skill escape

kesslerio:
## Intent

Fix a cross-skill resource link in the `printing-press-publish` skill that escaped its own skill directory. The Codex skill registry hard-blocks any `..` reference that resolves outside the skill dir, so the publish skill could not load `references/secret-protection.md` from the sibling `printing-press` skill.

Issue: N/A

## Approach

Duplicate `skills/printing-press/references/secret-protection.md` into `skills/printing-press-publish/references/secret-protection.md` and change the PII-scanning link in `printing-press-publish/SKILL.md` from `../printing-press/references/secret-protection.md` to the in-tree `references/secret-protection.md`. A drift-notice header on the copy points future editors back to the base file and the `--printing-press-only` re-sync flow.

## Scope

Primary area: `printing-press-publish` skill definitions and references.

Why this belongs in this repo: the `printing-press-publish` skill ships in this repo (the machine), and the startup-context skill registry validates skill resource links. Publishing CLIs to the public library is unaffected; this restores the publish skill's self-contained PII-scan reference.

## Risk

N/A — pure reference/duplication move, no generator, CLI, MCP, publish-flow, verifier, or scorer behavior changes.

### DIFF
--- a/skills/printing-press-publish/SKILL.md
+++ b/skills/printing-press-publish/SKILL.md
@@ -1831,7 +1831,7 @@ to proceed. The user makes the final call on those non-mandatory findings.
 ### PII pattern scanning (mandatory)
 
 Beyond the secret scans above, run the **PII pattern scanning** step from
-[../printing-press/references/secret-protection.md](../printing-press/references/secret-protection.md#pii-pattern-scanning)
+[references/secret-protection.md](references/secret-protection.md#pii-pattern-scanning)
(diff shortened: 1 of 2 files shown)


kesslerio:
Closing per operator request — keeping the fix in the kesslerio fork only; no upstream PR intended.

PR #4577, opened 2026-09-06
fix(verify): enable explicit environment live mode

### LINKED ISSUE #4286: scorer: verify --env-var can never reach live mode, so live_api_verification is permanently unscoreable
## Summary

`verify --env-var NAME` cannot reach live mode. The mode branch consults only
`cfg.APIKey`; `cfg.EnvVar` is read nineteen lines later, after mode is already
fixed to `"mock"`. A mock server is then started and every check runs against
it. `live_api_verification` is therefore never scoreable through the
`--env-var` path, and `shipcheck --env-var` reports
`HOLD (unverified: 1/7 legs)` on every auth-requiring CLI regardless of how
much real live testing was done.

## Root cause

`internal/pipeline/runtime.go` lines 132-137:

```go
// 2. Determine mode
if cfg.APIKey != "" {
    report.Mode = "live"
} else {
    report.Mode = "mock"
}
```

`cfg.EnvVar` is first read at line 151. Line 157 then starts a mock server
because `report.Mode == "mock"`.

The rest of the chain is correct and was verified by reading source:

| Link | Location | Status |
|---|---|---|
| shipcheck forwards `--env-var` to verify | `internal/cli/shipcheck.go:91` | correct |
| shipcheck passes `--write-manifest` to both legs | `internal/cli/shipcheck.go:79,154` | correct |
| verify persists its report | `internal/cli/verify.go:117` | correct |
| scorecard loads the report back | `internal/cli/scorecard.go:56` | correct |
| scorecard refuses non-live reports | `internal/pipeline/scorecard.go:1297` | correct by design |

Only the mode decision is wrong. The scorecard is right to withhold the
dimension — it has no live evidence to score.

## Why this is worth fixing

`--env-var` exists so the credential stays off the command line, out of shell
history, and out of process listings. `--api-key` works, but recommending it
instead is a security regression, and `shipcheck --env-var` is documented as
the way to give verify a credential.

The failure is silent. Verify prints `Mode: mock`, then two lines later
`Pass Rate: 100% (78/78 passed, 0 critical)` and `Verdict: PASS`. The
scorecard's `HOLD: unverified dimensions: live_api_verification` never states
the cause. In the session that produced this retro, the agent driving the tool
read that output and reported to the user that the CLI had passed 78/78 checks
*against the live API*; the claim had to be retracted after reading the source.
If the agent misread it with the source one command away, an operator reading
only the terminal will misread it too.

This branch runs before any spec is consulted and before any request is made,
so it is API-independent by construction. Every auth-requiring CLI verified via
`--env-var` is affected identically.

## Proposed fix

Resolve the env var into `cfg.APIKey` before the mode determination:

```go
if cfg.APIKey == "" && cfg.EnvVar != "" {
    if v := os.Getenv(cfg.EnvVar); v != "" {
        cfg.APIKey = v
    }
}
```

Explicit `--api-key` keeps precedence; an unset or empty variable still falls
back to mock.

Independently: when `--env-var` is given but the named variable is unset,
verify should say so rather than silently degrading. One line —
`--env-var FOO given but FOO is unset; running in mock mode` — turns a silent
misread into an obvious one.

## Acceptance criteria

- **positive:** with `X` exported to a valid key, `verify --env-var X` reports
  `Mode: live`, starts no mock server, and a following
  `scorecard --write-manifest <same manifest>` scores `live_api_verification`
  instead of listing it under `unverified_dimensions`.
- **negative:** with `X` unset, `verify --env-var X` still reports `Mode: mock`,
  still starts the mock server, and the dimension stays unscored. Explicit
  `--api-key` still wins over `--env-var`.
- **diagnostic:** `--env-var X` with `X` unset emits a line naming the variable
  and stating that mock mode was used.

## Scope boundary

Does not change how the scorecard scores the dimension, does not change
verify's check set, does not alter `--api-key` handling beyond preserving
precedence.

## Reproduction

```
cli-printing-press verify --dir <cli> --env-var FEC_API_KEY \
  --write-manifest <cli>/.printing-press.json
```

with `FEC_API_KEY` exported to a valid key. Output begins `Mode: mock`.

## Related issues

- #3963 — `scorer: scorecard total and grade are not reproducible across runs`.
  Adjacent: same component, also concerns scorecard output trustworthiness.

---

Complexity: small. Component: scorer.
Surfaced while printing a CLI for the OpenFEC API (92/100, Grade A).

jeffbander:
## Intent

Enable explicitly selected environment credentials to opt runtime verification into live mode, without placing secrets in shell arguments.

Issue: Fixes #4286.

## Approach

Choose live mode when `--api-key` is nonempty or the explicitly named `--env-var` resolves to a nonempty value. Leave an environment-sourced credential in its original environment variable: copying it into the shared APIKey field would broadcast it across multiple credential slots. Ambient credentials without explicit opt-in do not change the default mock mode.

An unset or empty explicitly named variable retains mock fallback and reports why, without reporting its value.

## Scope

Primary area: runtime verifier and CLI help.

Why this belongs in this repo: the mode selection affects all printed CLIs and the canonical shipcheck, rather than an API-specific endpoint implementation.

## Risk

An explicitly named, nonempty variable now triggers read-only live verification instead of mock execution. Explicit API keys retain precedence. Credential transport and multi-key separation are unchanged.

## Output Contract

Optional `verify.mode_detail` JSON field and a matching human-readable diagnostic for empty/unset explicit environment variables. Existing default output has no new field. No generator/template change.

## Verification

- PASS: `GOTOOLCHAIN=go1.26.6 go test ./internal/pipeline -run '^TestVerifyMode$' -count=1` (default mock, explicit key, explicit environment, empty environment, key precedence, ambient credential isolation, secret-free diagnostics).
- PASS: `GOTOOLCHAIN=go1.26.6 go vet ./internal/pipeline ./internal/cli`.
- PASS: native Printing Press binary build.
- PASS: golden verification, all 39 cases including verify-runtime-matrix.
- PASS: canonical shipcheck with explicit environment authentication on a read-only printed CLI, all seven legs; full live dogfood 92/92.
- `GOTOOLCHAIN=go1.26.6 go test ./...`: other packages passed; internal/generator reached Go's default ten-minute package timeout. A rerun with `-timeout 30m` is in progress; this is not reported as a pass.
- Generator/template output checks: not applicable; no generated output changed.

## AI / Automation Disclosure


tmchow:
Thanks for the start. Taking this from here so we can rebase cleanly after the progressive-phases skill landing and keep the fix off the skill router. Will open a maintainer PR for #4286 shortly.

Issue #4574, opened 2026-09-05
Documented ghs_ scrubbing pattern cannot match a GitHub App installation token in its new JWT form


smoochy:
## What

The GitHub server-to-server token rules in the secret-scrubbing references use an alphanumeric-only character class. That class cannot match a GitHub App installation token in the format GitHub is currently rolling out, so the rule silently matches nothing rather than matching partially.

## Where

- `skills/printing-press/references/secret-protection.md` — `'bearer-github-server|Bearer ghs_[A-Za-z0-9]{36,}|Bearer <REDACTED:github-server-token>'`
- `skills/printing-press-retro/references/secret-scrubbing.md` — `'github-server|ghs_[A-Za-z0-9]{36,}'`, and the same shape inside the combined `CRED_REGEX` further down the file.

## Why the class fails

Per GitHub's changelog of 2026-05-15 (https://github.blog/changelog/2026-05-15-github-app-installation-tokens-per-request-override-header/), installation tokens — `ghs_`, which includes the Actions `GITHUB_TOKEN` — are moving from a short opaque string to a stateless JWT: `ghs_` followed by roughly 520 base64url characters carrying two dots. Base64url contains `.`, `-` and `_`.

A JWT's header segment ends at its first dot after well under 36 characters, so `[A-Za-z0-9]{36,}` never reaches its minimum repetition and the alternative fails outright. The failure mode is a complete miss, not a truncated redaction, which is the worse of the two: a 520-character credential passes through a scrubbing pass that reports success.

GitHub's own recommended validation shape is `ghs_[A-Za-z0-9\.\-_]{36,}`.

## Suggested fix

Widen the class to `ghs_[A-Za-z0-9._-]{36,}` at all three occurrences. The dash goes last so it is not parsed as a range.

`ghp_`, `gho_` and `github_pat_` are a different token family, unaffected by this change, and should keep their current patterns.

## Note on scope

This is a documentation-and-pattern change only; no generator or golden output is involved. Happy to send it as a PR if you would prefer that over fixing it directly — I opened this as an issue because the same defect exists in more than one downstream copy of these references and the upstream fix is the one that actually propagates.

tmchow:
Closing as out of triage scope for the printed-CLI defect queue.

[...]

PR #4578, opened 2026-09-06
fix(skills): match GitHub installation tokens in their JWT form

smoochy:
Follow-up to #4574, per the closing note offering to take a PR that widens `ghs_` to the JWT-safe class.

## Problem

GitHub App installation tokens (`ghs_`, which includes the Actions `GITHUB_TOKEN`) are moving from a short opaque string to a stateless JWT: `ghs_` followed by roughly 520 base64url characters carrying two dots. Base64url contains `.`, `-` and `_`.

The documented class `[A-Za-z0-9]{36,}` cannot reach its minimum repetition, because a JWT's header segment ends at its first dot well under 36 characters, so the alternative fails outright. The failure mode is a complete miss rather than a truncated redaction: a 520-character credential passes a scrubbing pass that reports success.

## Change

Widen the class to `ghs_[A-Za-z0-9._-]{36,}` at all three occurrences (dash last so it is not parsed as a range), matching GitHub's own recommended validation shape:

- `skills/printing-press/references/secret-protection.md` - `bearer-github-server` entry
- `skills/printing-press-retro/references/secret-scrubbing.md` - `github-server` vendor pattern
- `skills/printing-press-retro/references/secret-scrubbing.md` - the same shape inside `CRED_REGEX`

`ghp_`, `gho_` and `github_pat_` are a different token family, unaffected, and keep their current patterns.

## Verification

Against a synthetic 379-character `ghs_<header>.<payload>.<signature>` token:

- old `ghs_[A-Za-z0-9]{36,}` - 0 matches
- new `ghs_[A-Za-z0-9._-]{36,}` - matches the full token

Docs and pattern only; no generator or golden output is involved.

### DIFF
--- a/skills/printing-press-retro/references/secret-scrubbing.md
+++ b/skills/printing-press-retro/references/secret-scrubbing.md
@@ -65,7 +65,7 @@ scrub_body() {
     'stripe-test-key|sk_test_[A-Za-z0-9]{20,}'
     'github-pat|ghp_[A-Za-z0-9]{36,}'
     'github-oauth|gho_[A-Za-z0-9]{36,}'
-    'github-server|ghs_[A-Za-z0-9]{36,}'
+    'github-server|ghs_[A-Za-z0-9._-]{36,}'
     'slack-bot-token|xoxb-[A-Za-z0-9-]{20,}'
     'slack-user-token|xoxp-[A-Za-z0-9-]{20,}'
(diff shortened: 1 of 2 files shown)


tmchow:
Thanks for the start. Taking this from here so we can widen the Go publish/archive scanners along with the skill patterns (JWT-form ghs_). Will open a maintainer PR shortly.

PR #4676, opened 2026-09-11
fix(cli): skip typed non-zero dogfood exits

### LINKED ISSUE #4641: [P1] scorer: live dogfood records a non-zero-exit happy path as pass, not skip (hid a shipped data-loss bug for 3 gate runs)
> Filed by `/printing-press-retro` from the **PBS** run (`library/other/pbs`, press 4.31.7,
> scorecard 84/100 grade A, verify 100%, live dogfood 92/92). Artifacts were not uploaded —
> no consent was given for a public upload this session, and the evidence below is complete
> without them. The full retro document, including the skip and drop tables, is local at
> `manuscripts/pbs/<run>/proofs/20260908-173034-retro-pbs-pp-cli.md` and can be attached on request.

## What happened

Live dogfood recorded `verdict: PASS, matrix 92, 92 passed, 0 failed` while **six of those
92 "passed" entries carried a non-zero exit code**:

| command | kind | exit |
|---|---|---|
| `verify` | happy_path | 5 |
| `verify` | json_fidelity | 5 |
| `basket` | happy_path | 3 |
| `basket` | json_fidelity | 3 |
| `weights` | happy_path | 3 |
| `weights` | json_fidelity | 3 |

The commands were behaving correctly. They were reporting "I have no data" against the
clean-room panel a gate run does not populate. The defect is that the harness labelled that
`pass` instead of `skip` — and `skip` already exists in the same vocabulary and is used 65
times in that very run for unmet preconditions.

## Why this matters more than a cosmetic label

The mislabelling **concealed a real, shipped data-loss bug for three consecutive full live
dogfood runs**, and it was an external code reviewer who caught it, not the machine.

In the PBS CLI a release is two files: an annexure (the city × item price grid) and a report
(the item weight vector, weight totals, quintile indices). Resume keyed on the annexure row
alone, so reports were never fetched for 99 of 148 weekly releases, which were recorded as
complete while holding no weights, no totals and no indices. `verify` exited 5 for exactly
that reason — its weight-total, section-count and pct-column invariants had no data to check.

A gate that reports that as `pass` cannot be used as evidence, which is its only purpose.
The proof of the causal link is that after fixing the CLI bug, those same six legs now exit 0
and the only remaining pass-with-non-zero entries are the six `error_path` legs, where a
non-zero exit IS the expected outcome.

## Cross-API evidence

Measured directly from committed `publish-live-gate.json` artifacts in the public library:

| CLI | tests in gate | happy_path/json_fidelity marked `pass` with non-zero exit |
|---|---|---|
| `amazon-jobs` | 126 | **8** — `teach`, `teach-pattern`, `teach-playbook`, `playbook amend`, all exit 2 |
| `cdc-pakistan` | 189 | **4** — `coverage map`, `stats history`, both exit 4 |
| `nccpl` | 326 | **2** — `contract-check`, exit 4 |
| `pbs` | 157 | **6** — as above (0 after the CLI fix) |
| `psx` | 333 | 0 — control |
| `peekaboo` | 166 | 0 — control |

Three named APIs besides PBS, with two clean controls showing it is not universal: it fires
on any CLI whose commands need local state a clean-room run has not built.

## Root cause

`scorer` — the live-dogfood happy-path evaluator treats a documented typed exit code as
success. PBS declares `pp:typed-exit-codes: "0,5"` on `sync`, and the affected commands exit
3 or 5 when the local panel is empty. The exit code is already captured in the artifact, so
no new information is needed to fix this.

## Proposed fix

1. In the happy-path and json-fidelity evaluators, a non-zero exit becomes `skip` with a
   reason naming the code (`"exited 3: precondition not met"`), never `pass`.
2. Leave `error_path` semantics untouched — there a non-zero exit is the expected outcome,
   and six such legs in this same PBS run are correctly `pass`.
3. Have `--write-acceptance` record `tests_passed` net of those skips, so the matrix count
   means what it says. This matters because that marker is the artifact gating
   `publish package`.

### Acceptance criteria

- **positive:** a happy-path leg exiting 3 against an empty store is recorded `skip` with a
  reason, and is excluded from `tests_passed`
- **negative:** an `error_path` leg exiting 3 stays `pass`; a happy-path leg exiting 0 stays `pass`

## Related issues

- #2488 — asks for a live *stateful* pass because per-command smoke misses populated-store
  bugs. Complementary rather than duplicate: that issue adds a capability, this one fixes the
  labelling in the existing pass. They reinforce each other — had these legs been marked
  `skip`, the operator would have seen plainly that those commands were never really
  exercised, which is precisely #2488's complaint.
- #4083 — finding 1 there is "dogfood reports a pass shaped result when it validated zero
  paths", for whole runs where no spec resolves. Same family ("reports pass for something it
  did not verify"), different mechanism: this one is per-leg and exit-code-based.

kunallanjewar:
## Intent

Live dogfood currently counts declared non-zero exits as passes for `happy_path` and `json_fidelity`, even though those runs produced no successful output to validate. This inflates the pass and matrix counts and can hide unexercised commands.

Fixes mvanhorn/cli-printing-press#4641

## Approach

Treat exit `0` as pass, but record a declared non-zero exit as skip with the exit code in the reason. Keep undeclared non-zero exits as failures and leave `error_path` semantics unchanged. The existing report finalizer then excludes the new skips from `Passed` and `MatrixSize`.

## Scope

Primary area: live-dogfood scoring in `internal/pipeline`

Why this belongs in this repo: this is systemic Printing Press scoring behavior shared by every printed CLI, not an API-specific CLI fix.

## Risk

Runs whose only apparent happy-path or JSON successes were declared non-zero exits will no longer have a successful live signal. Quick runs may also fall below their acceptance threshold when those entries leave the matrix. Both changes are intentional: a run that produced no successful output should not pass the live gate.

## Output Contract

For `happy_path` and `json_fidelity`, a declared non-zero exit now produces `status: skip` with reason `declared non-zero exit <code>`. Acceptance `tests_passed` and matrix size exclude those skips. Exit-zero results and `error_path` behavior are unchanged.

No templates or generated files changed. Repository-wide tests and golden/generated-output verification are left to the pull-request CI gates.

## Verification

Reviewed SHA: `54ff4a3c376eb821c2419cb95dfac82fc8dabf19`

The focused regression failed on the test-only commit `81bf4ffd5f97e073988e8d82407484b8c2ce9a96` with expected `skip`, actual `pass`, then passed on the reviewed SHA.

- `go build ./...`
- `go vet ./...`
- `go test -short -timeout 8m ./internal/pipeline/... -run TestRunLiveDogfoodSkipsDeclaredNonzeroSuccessExits`
- `gofmt -l internal/pipeline/live_dogfood.go internal/pipeline/live_dogfood_test.go` (no output)
- `go test ./...` not run locally; the full suite runs in upstream pull-request CI
- Golden and generated-output checks not run locally; upstream pull-request CI owns those gates, and this patch does not touch templates

- [ ] Generator/template change: verified generated output, including emitted-code assertions or compiled generated CLI output — N/A; no generator or template change
- [ ] Generator/template change: covered the affected fallback or variant shape, not only happy-path fixtures — N/A; no generator or template change
- [ ] Generator/template change: checked emitted definitions and call sites for matching gates — N/A; no generator or template change

## AI / Automation Disclosure


tmchow:
@mergifyio requeue

Speculative draft #4702 timed out full-test (generator-0) at 15m. PR head itself is green and this change only touches live-dogfood scoring + skill docs, so I'm putting it back in the queue.

tmchow:
@mergifyio requeue

Latest speculative draft #4703 was dequeued by Mergify checks_timeout (60m) while CI was still queued/starting; those checks later went green. Prior draft #4702 hit the same generator-0 15m timeout flake. PR head CI is green — requeueing.

PR #4688, opened 2026-09-13
fix(cli): isolate generated config permission tests from real credentials

### LINKED ISSUE #4288: config_perms_test.go isolates only env credentials, so it measures the developer's real credentials file
## What happens

The generated `config_perms_test.go` fails three cases on any machine that has a credentials file:

```
--- FAIL: TestLoad_RefusesOverPermissiveConfigOnRead
    config_perms_test.go:127: precondition: owner-locked config should load the token, got ""
--- FAIL: TestLoad_RefusesSymlinkToLoosePermsTarget
    config_perms_test.go:204: symlink to loose-perms target must be refused (canonicalize-first)
--- FAIL: TestLoad_DanglingSymlinkIsMiss
    config_perms_test.go:227: dangling symlink must be a miss
```

Seen on a Slack CLI at press 4.29.0, run id `20260801-150754-271fdf29`, but nothing about it is Slack-specific.

## Why

`clearCredEnv` — the helper every case in the file calls first — clears the two credential env vars and stops there:

```go
func clearCredEnv(t *testing.T) {
	t.Helper()
	t.Setenv("SLACK_BOT_TOKEN", "")
	t.Setenv("SLACK_USER_TOKEN", "")
}
```

But the generated `Load` it exercises has a second credential source. `config.go` consults `cliutil.LoadCredentials()` for `<data-dir>/credentials.toml` and, when that file has values, clears the config-file credential fields and applies its own:

```go
creds, ok, err := cliutil.LoadCredentials()
...
if ok && creds.HasValues() {
	cfg.clearCredentialFields()
	cfg.applyCredentials(creds)
```

So the real file wins over the fixture the test just wrote, and the over-permissive case fails on its own precondition: it asserts `cfg.RefreshToken` holds the fixture secret when it holds nothing.

## Why it is worth fixing in the template

The file's whole job is asserting that a group/world-readable token config gets refused. It passed for months on this machine and started failing the day `auth set-token` first created a credentials file there — nothing in the test or the code under test changed. A green suite one `auth set-token` away from red.

To be fair to it: it goes loudly red rather than silently vacuous, so no guarantee was quietly lost. The same gap on an assertion that happened to still hold would pass while measuring the wrong file, which is the version worth pre-empting.

And the fix cannot live downstream. The file carries the `DO NOT EDIT` header, so a reprint drops any local repair — we are carrying ours as a patch record for exactly that reason.

## Suggested fix

Have the template isolate every credential source it generates a reader for, not just the env vars. `<PREFIX>_HOME` is already emitted by the path resolver and redirects config, data, state and cache under one root in a single variable, so it needs one line:

```go
func isolateCredentials(t *testing.T) {
	t.Helper()
	t.Setenv("SLACK_BOT_TOKEN", "")
	t.Setenv("SLACK_USER_TOKEN", "")
	t.Setenv("SLACK_HOME", t.TempDir())
}
```

Worth renaming the helper too. `clearCredEnv` reads as covering only the environment, which is the mistake it invites.

Verified as load-bearing rather than a mask: with that line the package is green, commenting it out brings back the same three failures, and restoring it returns to green.

Happy to open a PR against the template if useful.

https://claude.ai/code/session_01TbWCDV2TGAWnnhimPPnBKe

tmchow:
Generated `config_perms_test.go` isolated credential env vars and the data dir, but `config.Load` still falls through to `LoadCredentials()` at the operator's default credentials path. On any machine with a real credentials file, `TestLoad_RefusesOverPermissiveConfigOnRead`, `TestLoad_RefusesSymlinkToLoosePermsTarget`, and `TestLoad_DanglingSymlinkIsMiss` fail or measure the wrong file.

The emitted helper now resets `--home`, sandboxes every user-directory lookup, and sets `<PREFIX>_HOME` to that sandbox so permission tests use only temp fixtures.

A generated-tree test plants a decoy credentials file outside the fixture and requires the suite to stay green without rewriting it. A companion case strips the isolation and requires the decoy to turn the suite red.

Closes #4288

<div><a href="https://cursor.com/agents/bc-06c84f24-3a7e-4290-83b0-02c366d5fa28?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-06c84f24-3a7e-4290-83b0-02c366d5fa28&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


tmchow:
@mergifyio requeue

PR #4689, opened 2026-09-13
fix(cli): infer browser-sniff requests from traffic, not responses

### LINKED ISSUE #4255: Browser-sniff capture redaction misses credentials stored in header values
> Filed by `/printing-press-retro` from a run of **Clay** (scorecard 80/100, 100% verify pass).

## Summary

The browser-sniff capture redaction described in `references/browser-sniff-capture.md` matches
credentials by JSON **field name** (`"apiKey"`-shaped keys). It does not match credentials that
appear as HTTP **header values** nested inside a captured request body. During a capture of a
product that lets users store their own outbound credentials, a live third-party credential
(an `Authorization` header with a base64 user:pass pair, stored by the user inside an
HTTP-request column) passed through redaction untouched and was written to the on-disk capture
artifact and into agent context. It was caught only by incidental manual inspection.

This is a redaction-completeness defect in the capture playbook, not agent error: the playbook
is what the agent is told to trust for this step.

## Where to look

- **Component:** `comp:skill`
- **Issue type:** `bug`
- **Likely area:** `skills/printing-press/references/browser-sniff-capture.md` (redaction guidance),
  and any shared redaction helper it delegates to.
- **Triggered when:** browser-sniffing any product whose users configure outbound credentials
  inside the product itself — integration platforms, workflow builders, iPaaS, scraper/HTTP-request
  column features. Independent of the captured API's own auth mode.

## What we observed

The captured request body for a user-configured HTTP-request column contained a headers map of
roughly this shape (credential elided; not reproduced here, per the redaction rule below):

```json
{"name":"headers","formulaMap":{
  "Content-Type":"\"application/json\"",
  "Authorization":"\"Basic <REDACTED:vendor-basic-credential>\""}}
```

The redaction pass in use targeted `"(apiKey|token|secret|password|...)"\s*:\s*"..."` field names.
The credential above is a *value* under a `formulaMap` key, so no pattern matched and the artifact
was written with the credential intact.

A follow-up corrective pass then over-corrected: a blind sweep over long base64-shaped tokens was
applied across the whole artifact, which matched URL path segments and corrupted the `url` fields
of ~200 captured requests, destroying endpoint evidence that had to be re-captured.

> Note: the actual leaked value is deliberately not quoted here. Quoting it to prove the leak
> exists would re-leak it.

## Suspected root cause

High confidence on the mechanism: the redaction pattern set is keyed on field names rather than
value shapes, so any credential that is a value (rather than a conventionally-named key) is
invisible to it. Lower confidence on where the durable fix belongs — the guidance lives in the
skill reference, but if a shared Go helper performs this scrub for sniffed artifacts, the fix
belongs there instead. Verify by checking whether capture-artifact scrubbing is implemented in
the skill prose only or in a shared helper.

## Suggested direction

Match on **value shape**, not field name, and bound the scrub's scope:

1. Redact any `Authorization` header value regardless of scheme.
2. Redact bare `Basic` / `Bearer` / `Token` blobs anywhere in a body.
3. Redact `x-api-key`-style header names by pattern.
4. Redact base64 blobs that decode to a `user:pass` pair.
5. **Scope guard (important):** scrub request/response **bodies only** — never URL, host, or path
   fields. The over-correction above is what happens without this rule, and it is a worse failure
   than the original miss because it silently destroys evidence.

## Acceptance criteria

- positive: a capture whose body contains an `Authorization: Basic <blob>` inside a nested map is
  redacted in the written artifact.
- negative: a capture whose URLs contain long base64-shaped path segments retains those URLs
  byte-for-byte.

## Related issues

- #3768 — credential handling hardening in skills (chat paste, exact-value scans). Different
  mechanism, same goal; this issue adds the capture-artifact path.
- #3780 — raw browser-sniff discovery captures carrying PII into archived/published manuscripts.
  Adjacent surface: that issue is about not shipping the artifact, this one is about the artifact
  being unredacted in the first place.

### LINKED ISSUE #4662: browser-sniff: preserve request/route/origin semantics in real multi-host HAR captures
﻿## Summary

A real browser HAR can currently produce a syntactically valid printed CLI whose request surface differs materially from the traffic that succeeded in the browser. A 4books.com capture reproduces four generic inference failures on current `main` (`2dde45af`):

1. Response fields become request params: `GET /api/ms-1/skills` emits a required `--skills`, and a map-shaped XP response emits `--1`, `--2`, etc.
2. Literal snake_case route names can be mistaken for prefixed IDs: `/api/v1/users/35854/read_progresses` becomes `/api/v1/users/{user_id}/{user_id_2}`.
3. Raw valid JSON bodies advertised as `application/x-www-form-urlencoded` are parsed as one form-field name instead of JSON fields.
4. Endpoint-origin preservation is coupled to secondary-host inclusion: default mode drops a third-party auth host but loses the correct origin for a first-party HTML route; `--preserve-hosts` restores origin but also re-includes the third-party host and lets its `?key=` contaminate global auth.

The common problem is conflating evidence channels: response schema with request input, route literals with identifiers, body syntax with declared media type, and endpoint origin with host-selection policy.

## Suggested direction

- Never synthesize request params from response fields.
- Require opaque-ID evidence in prefixed-ID tails instead of accepting ordinary lowercase snake_case route words.
- Infer body syntax/encoding from the successfully parsed payload independently of the advertised media type; preserve replay semantics when they disagree.
- Separate endpoint-origin preservation from secondary-host inclusion/filtering.

## Acceptance criteria

- GET with no observed request params emits zero request params even when its response has fields.
- `/users/35854/read_progresses` becomes `/users/{user_id}/read_progresses`; opaque prefixed IDs still parameterize.
- Raw valid JSON is modeled as JSON fields even under a misleading form Content-Type, without changing successful replay semantics.
- A surviving first-party route on a secondary origin retains its own `base_url` without requiring unrelated third-party hosts to be included.
- Focused regression tests plus one multi-host HAR-style fixture cover the behavior.

Found while re-running a May 2026 4books dogfood capture against current main; the old v4.9.0 output showed the same core semantic failures.

tmchow:
## Intent

Issue: #4662 and #4255

A real multi-host HAR could emit a syntactically valid printed CLI whose request surface did not match the traffic that succeeded in the browser, and capture redaction missed credentials stored as nested HTTP header values.

Closes #4662
Closes #4255

## Approach

Treat request evidence, route literals, body syntax, and endpoint origin as separate channels:

- Never copy response fields onto request params. A GET with no observed query/path/body inputs emits zero request params even when the response has fields.
- Parameterize prefixed path segments only when the tail looks opaque. `/users/35854/read_progresses` stays `/users/{user_id}/read_progresses`; `t_0t3vswhpKogASf2XZpW`-shaped IDs still become path params.
- Infer body fields from a successfully parsed JSON payload even when Content-Type says form-urlencoded, and replay that payload as JSON so the CLI does not treat the raw document as one form-field name.
- Always keep a surviving route's `base_url` when its origin differs from the spec base URL. `--preserve-hosts` only controls whether secondary API hosts are included.

Redaction scrubs nested header maps, `Basic`/`Bearer`/`Token` values, and query/userinfo credentials on URL-like scalars. Opaque URL path segments stay intact.

## Scope

Primary area: `internal/browsersniff` inference and sample redaction, plus the capture-playbook strip rules in `skills/printing-press/references/browser-sniff-capture.md`.

Why this belongs in this repo: these are generator/browsersniff failures that ship a wrong CLI and leak credentials from sniffed artifacts. The durable fix is in the Printing Press, not one printed CLI.

## Risk

Sniffed specs that previously inherited response fields as required flags will lose those flags (the correct request surface). `--preserve-hosts` still includes secondary hosts and can still infer auth from them; first-party origin no longer depends on that flag. Body redaction is broader for nested Authorization-shaped keys, auth-scheme blobs, and URL query/userinfo; URL path segments remain excluded from whole-string sweeps.

## Output Contract

No generator templates changed. `browser-sniff-sample` and all `generate-*` goldens are unchanged. Focused package tests cover the 4books-style inference failures and header-value / URL-credential redaction.

## Verification

- [x] `go test ./internal/browsersniff/`
- [ ] `go test -timeout 40m ./...` (in progress after Greptile follow-up)
- [x] `scripts/golden.sh verify` — `browser-sniff-sample` and generate fixtures passed on the first revision; no golden update
- [x] Covered URL-like scalar query/userinfo redaction without destroying path segments

## AI / Automation Disclosure

<div><a href="https://cursor.com/agents/bc-fcca7216-b46d-4429-939b-aa0bd85a670b?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-fcca7216-b46d-4429-939b-aa0bd85a670b&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


tmchow:
@mergifyio requeue

tmchow:
@mergifyio requeue

Speculative drafts #4694 and #4706 both failed full-test (generator-0) with the shared 15m package timeout (orphan `go`/`compile` during requireGeneratedCompiles). PR head CI is green, Greptile/threads clean — treating as speculative flake.

PR #4707, opened 2026-09-14
fix(cli): classify typed rate limits

### LINKED ISSUE #4700: generator: classifyAPIError misses the generator's own *platform.RateLimitedError, so documented exit 7 ships as exit 5
_Intended labels (could not be applied — filed by a non-collaborator without `triage` permission): `source:retro, bug, priority:P2, comp:generator`_

> Filed by `/printing-press-retro` from a run of **Artificial Analysis** (scorecard 90/100, 100% verify pass, live dogfood 223/223).

## Summary

A printed CLI documents exit 7 for "rate limited" in its README, its SKILL, and its `pp:typed-exit-codes` annotations. When the generated client's own budget tracker gives up, the CLI exits 5 instead. The classifier looks for rate limiting by searching the error message for the substring `HTTP 429`, and the generator's own `*platform.RateLimitedError` does not contain it. A caller then cannot tell "retry tomorrow" from "the API is broken". On a metered API, an agent that reads 5 as transient spends the user's remaining quota retrying.

## Where to look

- **Component:** generator
- **Issue type:** `bug`
- **Likely area:** the template behind `internal/cli/helpers.go`, specifically `classifyAPIError` and `classifyAPIErrorOnly`
- **Triggered when:** any CLI whose generated client has a rate-limit budget, which is the default

## What we observed

`internal/platform/ratelimit.go` builds the message:

```go
func (e *RateLimitedError) Error() string {
	message := fmt.Sprintf("rate limited: endpoint class %s exhausted after %d attempts", e.EndpointClass, e.Attempts)
	...
}
```

`internal/cli/helpers.go` classifies it:

```go
var typed *cliError
if errors.As(err, &typed) {
	return err
}
msg := err.Error()
switch {
...
case strings.Contains(msg, "HTTP 429"):
	return rateLimitErr(err)
default:
	return apiErr(err)
}
```

The pass-through guard covers `*cliError`, the type a CLI author writes. It does not cover the types the generator emits from its own `internal/platform` package, so `*platform.RateLimitedError` reaches the string switch, matches nothing, and lands on `default`.

Checked across the local library: all 16 printed CLIs carry the `HTTP 429` string match and 11 also carry `platform.RateLimitedError`. Six were read line by line (treni, partitaiva24, scrape-creators, instagram-dms, drafts, apple-notes). None of them matches the typed error before reaching the string comparisons.

## Suspected root cause

Certain. The classifier recognises an error by its message text, and the generator's own error types carry none of the text it looks for. #1549 added the right mechanism to this function, as a guard for `*cliError`, and covered only the types a CLI author writes by hand.

## Suggested direction

Extend the `errors.As` guard in the same function to the generator's own typed errors before any string comparison, mapping `*platform.RateLimitedError` to `rateLimitErr`. Keep every string case as the fallback for messages that come off the wire.

## Acceptance criteria

- positive: a fake client returning `*platform.RateLimitedError` makes the command exit 7
- negative: a wire response carrying `HTTP 429` still exits 7; `HTTP 500` still exits 5; a CLI-authored `*cliError` still passes through untouched

## Frequency

every CLI with a client-side rate-limit budget

## Complexity

small

## Scope boundary

Only the classifier's matching order. No change to the budget tracker, the retry policy, or the documented exit-code table.

## Dependencies

None.

## Related issues

- #1549 (`extends`): added the `*cliError` pass-through to this same function. Nobody added the generator's own error types to it.

---

Tethys0:
## Intent

Fixes #4700.

Generated CLIs can exhaust their own client-side rate-limit budget and return `*platform.RateLimitedError`. The classifier only recognized wire errors containing `HTTP 429`, so the documented rate-limit exit code was not emitted for the generator's own typed error.

## Approach

Check for the generator-owned typed rate-limit error immediately after the existing `*cliError` pass-through in both classifier helpers. Reuse the existing `rateLimitErr` mapping so the existing exit-7, JSON envelope, and message behavior stay consistent with HTTP 429 responses.

## Scope

Primary area: generator error classification in the emitted CLI helper.

Why this belongs in this repo: the defect is in the reusable generator template and affects every future printed CLI with the client-side rate-limit budget; it is not API-specific.

## Risk

Low. This only adds a typed-error match before the existing message-based fallback. Existing typed CLI errors and HTTP status-string fallbacks are unchanged.

## Output Contract

This changes emitted `internal/cli/helpers.go` behavior for generator-owned rate-limit errors. The regression test pins the typed classification block in both emitted classifier paths, and the generator compile suite covers generated CLI output. No CLI flags, MCP schemas, or documented exit-code values change.

## Verification

- [x] `go test ./internal/generator -run 'TestClassifyAPIErrorRecognizesGeneratedRateLimitError|TestGenerateProjectsCompile|TestPaginatedGet' -count=1 -v`
- [x] `go fmt ./...`
- [x] `git diff --check`
- [x] `go test ./...` reached 6,622 passing tests across 44 packages before the scheduled runner's 420-second command timeout; no test failure record was emitted
- [x] Independent spec and quality/security review completed
- [ ] `scripts/golden.sh verify` - not completed because the runner lacks system `jq`; a project-local `gojq` substitute was incompatible with this script's `jq -S` invocation and was removed
- [ ] `scripts/verify-generator-output.sh` - not run after the same `jq` prerequisite blocker

## AI / Automation Disclosure

- [x] AI-reviewed only: AI or automation reviewed the work, but no human reviewed it before submission


tmchow:
Thanks for the typed rate-limit classification work — the approach matches what we want.

Closing this draft so we can land the same fix on a maintainer branch (community draft triage skips drafts, which was parking #4700). Feel free to open a follow-up anytime if you spot another printed-CLI exit-code miss.

PR #4749, opened 2026-09-18
fix(cli): score data-pipeline dimensions on collection-spec CLIs with no store instead of N/A

### LINKED ISSUE #4748: scorecard: a spec with GET collections but no emitted store is N/A'd, hiding generator under-detection
**Title:** scorecard: a spec with GET collections but no emitted store is N/A'd, hiding generator under-detection

## The symptom

When the generator's profiler fails to detect a syncable resource, the printed CLI comes out with no `internal/store/store.go`. That CLI is broken by omission: it has no `sync`, no local data, no `sql`/`search` surface. But every quality signal we have reports it as healthy:

- `cli-printing-press verify` self-skips the data-pipeline check (`SKIP (CLI has no sync command)` / `SKIP (CLI has no local store)` inside `runDataPipelineTest`) and stays green.
- `cli-printing-press score` marks `data_pipeline_integrity` and `sync_correctness` as N/A (the `!hasScorecardLocalStore` branch in `scoreDomainDimensions`, from #3319) and stays green.

So a generator that silently dropped a syncable resource scores as healthy on every quality signal. The blanket no-store N/A from #3319 is right for legitimately stateless CLIs; the gap is that it never looks at the spec to tell those apart from under-detection. (No shipped CLI hits this today; the repro below is a fixture.)

## The tell

A spec that declares a readable REST collection (a `GET /items` list plus a `GET /items/{id}` detail sibling) is exactly the shape the profiler turns into a store. Outside the store-less shapes listed under Suggested fix, a spec with that pair and a tree with no store means the profiler missed it. RPC-style APIs (only `/v1/load`, `/v1/sql`-style actions) have no such pair and stay legitimately stateless, so the N/A remains correct for them.

## Repro

The repro is a fixture, not a live CLI, and that is the honest state of it: I swept every spec under `testdata/` (48 files, 38 of which contain GET collections, all 38 of which emit stores today) and 28 CLIs in my local library, and today's profiler emits a store for every collection-shaped spec it is handed. Zero live under-detections, zero false positives. The guard is latent: it fires the moment a future profiler regression drops a resource, which is the only time it should fire.

The test in the attached PR is the repro: a collection-shaped spec plus a store-less tree must keep `data_pipeline_integrity` and `sync_correctness` scored (at 0, with a gap line naming the generator), not N/A them.

## Suggested fix

- Add a GET-gated collection predicate (`GET P` + `GET P/{param}` pairing) over the spec's paths.
- In `scoreDomainDimensions`'s no-store branch: when the spec has collections and the tree has no store, score both dimensions 0 and append one specific gap line pointing at the generator, so self-improve agents fix the profiler instead of hand-adding a store to the printed CLI.
- Leave verify's SKIP behavior alone. A hard verify FAIL for a defect the sweep could not produce is not worth the false-positive risk on shipcheck.
- Exclusions for shapes the generator deliberately emits store-less: device CLIs, local-datastore manifests, GraphQL, synthetic specs, and the HTML page-mode sync stub.

The gap line keeps TypeFidelity on the current N/A model; this touches only the two pipeline dimensions.

Known limitation: no spec means no claim. The guard needs the spec's method data; without `--spec` (or an archived spec) it stays silent, which is the current behavior for every check that reads the spec.

Cameron-Fulton:
## Intent

Issue: #4748

Closes #4748

A spec that declares a readable REST collection (`GET /items` + `GET /items/{id}`) but prints a CLI with no `internal/store/store.go` means the generator's profiler dropped a syncable resource. Today the scorecard N/As `data_pipeline_integrity` and `sync_correctness` for any store-less tree (#3319), so that bug scores as healthy. This PR scores those two dimensions 0 and appends a gap line that names the generator as the fix target.

## Approach

One GET-gated predicate, `storeUnderDetected(dir, specGETPaths, specKind, specIsGraphQL)`, consulted only inside the existing `!hasScorecardLocalStore` branch of `scoreDomainDimensions`. It composes the collection pairing (`GET P` + `GET P/{param}`) with exclusions for every shape the generator deliberately emits store-less: device CLIs, local-datastore manifests, GraphQL (spec flag and dir marker), synthetic specs, and the HTML page-mode sync stub. GET-method data is threaded through the scorecard's raw paths loop and the verify/dogfood loader (`GETPaths` next to `Paths`), with `strings.EqualFold` method comparison throughout, so a write-only `POST /orders` + `DELETE /orders/{id}` pair never reads as a collection.

The gap string is appended by `appendStoreUnderDetectionGap` after both GapReport rebuilds (`finalizeScorecard` and `ApplyLiveCheckToScorecard`), mirroring how `appendNovelFeatureDepthGaps` is wired, so the live-check path cannot drop it. Verify's SKIP behavior is untouched. `type_fidelity` stays on the current N/A model; only the two pipeline dimensions change.

## Scope

Primary area: `internal/pipeline` (scorecard + spec loaders).

Why this belongs in this repo: this is the machine's scoring model, not a printed-CLI defect. The under-detection it guards against is a generator profiler regression; the scorecard is where that regression is currently invisible.

## Risk

Score impact only: a CLI whose spec has GET collections and whose tree has no store now scores 0 on two dimensions that were N/A before. A sweep over every `testdata/` spec (48 files, 38 of which contain GET collections, all 38 of which emit stores today) and 28 local-library CLIs found zero such trees today, so no real CLI's score changes; the guard is latent until a profiler regression drops a resource. No automated gate consumes these two dimension scores, and the `GETPaths` field is read only by the new predicate. When the scorecard has no spec (no `--spec`, no archived spec), the guard stays silent and both dimensions stay N/A, which is the current behavior. False positives are contained by the GET gate plus the exclusion set, which covers every deliberate store-less shape we know of, and a cross-loader test pins that the scorecard and verify loaders agree on the answer for the same spec bytes.

## Output Contract

Scorecard output only: one additional gap-report line in the under-detected case, and two dimensions moving from `unscored_dimensions` to scored-0. No templates, generated files, manifests, MCP schemas, or pipeline artifacts change; `scripts/golden.sh verify` shows 0 diff.

## Verification

(Not generator/template changes; the boxes above do not apply. What was run:)

- `go test ./...` (full suite, green)
- `scripts/golden.sh verify` (0 diff)
- `scripts/verify-generator-output.sh` (green)
- `golangci-lint run ./...` (clean)
- New tests: `TestSpecHasCollectionShapedResource` (pairing table), `TestStoreUnderDetected` (every exclusion flips the answer), `TestStoreUnderDetectedCrossLoader` (scorecard and verify loaders agree; write-only pair stays negative), and an additive subtest in `TestRunScorecard_UnscoredSpecDimensions` (collection spec + store-less tree keeps both dimensions scored at 0 with the specific gap line; `type_fidelity` stays unscored; the existing no-spec subtests are unchanged)

## AI / Automation Disclosure


Cameron-Fulton:
Addressed both findings in 2b7343e7:

- **Path Pairing Misclassifies Stores (P1)**: exported `profiler.IsScalarItemArray` and reused it (instead of re-deriving the decision) to exclude scalar-item-array GET legs from all three GETPaths loaders before the collection-pairing check. A `GET /items` returning bare scalar IDs no longer counts as the list leg of a pairing, matching the profiler's own exclusion.
- **Comments Restate Their Functions (P2)**: reworded the four flagged comments to lead with the non-obvious rationale instead of paraphrasing the function signature.

Also fixed CodeRabbit's related finding on the fork mirror PR (scorecard's raw-JSON loader didn't resolve OpenAPI 3.1 `$ref`-based Path Item Objects) since it's the same code path.

New tests: `TestLoadOpenAPISpecDataGETPathsExcludeScalarArrayResponses`, `TestLoadOpenAPISpecDataGETPathsResolvesPathItemRefs`, `TestCollectGETPathLoadersExcludeScalarArrayResponses`.

Cameron-Fulton:
Good catch — fixed in 09a4fe33. The scalar-array check only rejected an inline scalar item type; a `$ref` to a named scalar type (`items: {$ref: ".../ItemID"}`) wasn't dereferenced.

Rather than add a third one-off ref resolver, I found this repo already has a general-purpose one (`resolveRawSchemaRef`/`resolveRawPointer` in `dogfood_mock.go`, used by the nested-envelope detection) and switched both the schema-ref and the path-item-ref resolution in this guard to use it — removes the two narrower resolvers I'd added in the last commit.

New test: `TestLoadOpenAPISpecDataGETPathsExcludeReferencedScalarArrayResponses`.

tmchow:
Thanks for the careful scorecard work — Greptile landed at 5/5 and the scalar/$ref exclusions look solid.

Closing though: this changes how we score store-less trees (N/A → 0 on pipeline dimensions), not a printed-CLI defect an operator hits. Same bucket as other scorer/verify fairness work we don't land via community PRs. Leaving #4748 open for issue triage.

